PR #11, opened 2015-07-03
Update traceur to version 0.0.90

rchavarria:
I first didn't know how to do the first ECMAScript exercise with `exercism`. But then I realized that I must have `traceur` installed (as well as `jasmine-node`).

So, I installed them both, and tried to execute some code to pass the first test in `palindrom-products_test.spec.js`. I used the commands in `Makefile` file (the `make` command works, so those commands should be the right ones). But I got this error:

    Exception loading: /[...]/ecmascript/palindrome-products/palindrome-products_test.spec.js
    [ReferenceError: $traceurRuntime is not defined]

The error was due to traceur inserted this line of code before the code requires `traceur-runtime`:

    'use strict';
    $traceurRuntime.options.symbols = true; // use of this var before requiring runtime
    exports.generate = generate;
    require('./traceur-runtime');
    function generate($__16) {
        //...
    }

Whe I installed `traceur`, its version was `0.0.90`. But `traceur` version in [xecmascript](https://github.com/exercism/xecmascript) project was `0.0.42`. I think it failed because the newer version implements more features and there must be some differences between those versions.

I managed to fix that problem by disabling the `symbols` feature:

    traceur --experimental --symbols=false --modules=commonjs --script example.js --out palindrome-products.js

This pull request updates `traceur` and fixes a little problem with the new version.


kytrinyx:
Thank you!

PR #13, opened 2015-07-07
Make it easier for a user to run assignment's tests 

rchavarria:
This pull request includes Node.js dependencies file (`package.json`) and a Gulp file (`gulpfile.js`) that allows the user to run assignment tests with just one command (once all dependencies are installed):

```bash
$ gulp test
```

These files are included in the only existing assignment directory (palindrome-products). Each assignment should have the same files so that there is a common way to run tests for all assignments.

The pull request also includes content for the `SETUP.md` file explaining how to install dependencies and how to run tests. Finally, it modifies `.gitignore` to ignore transpiled JavaScript files (from ES6 to ES5) and tools installed for each assignment.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -1,0 +1,38 @@
+## Setup
+
+Go through the setup instructions for ECMAScript to
+install the necessary dependencies:
+
+http://help.exercism.io/getting-started-with-ecmascript.html
+
+## Requirements
+
+They are already described in the link above, but just as a
+quick reference:
+
+Install globally a tool to run [Gulp](http://gulpjs.com) if
+it is not installed yet:
+
+```bash
+$ npm install -g gulp-cli
+```
+
+Install assignment dependencies:
+
+```bash
+$ npm install
+```
+
+## Making the test suite pass
+
+Execute the tests with:
+
+```bash
+$ gulp test
+```
+
+In many test suites all but the first test have been skipped.
+
+Once you get a test passing, you can unskip the next one by
+changing `xit` to `it`.
+
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ npm-debug.log
 tmp
 bin/configlet
 bin/configlet.exe
+
+*/node_modules
+*/traceur-output
--- a/palindrome-products/gulpfile.js
+++ b/palindrome-products/gulpfile.js
@@ -1,0 +1,46 @@
+var TRACEUR_OUTPUT_DIR = 'traceur-output',
+  gulp = require('gulp'),
+  jasmine = require('gulp-jasmine'),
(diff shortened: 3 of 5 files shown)


SETUP.md
```diff
@@ -0,0 +1,41 @@
+## Setup
+
+Go through the setup instructions for JavaScript to
```
kytrinyx: Should this say ECMAScript rather than JavaScript?

kytrinyx:
This looks good, aside from the one tiny comment.

Would you mind squashing the commits?

SETUP.md
rchavarria: Oooops. Yep, you're right.

rchavarria:
Ok, I'll try to squash commits, but I'm not very good at git.

I'll also remove some lines in `SETUP.md`, because some content is already in the [Getting started with ECMAScript](https://github.com/exercism/docs/blob/master/app/pages/languages/getting-started-with-ecmascript.md) page.

rchavarria:
Squash done! The pull request only contains one commit.

Just to know: are pull requests with one commit preferred to ones with multiple commits? I have no problem, but I like multiple small commits than one large commit. Maybe, it's easier to merge one single commit than multiple ones. I'm just asking.

kytrinyx:
I don't necessarily have a rule/preference around one commit per PR, but one change per commit. In general, if there were a bunch of little commits that kind of mean "oops, I forgot this one bit" or "darn, that didn't work the way I expected" (things that I do all the time!) then I squash my commits into a single one.

If, while doing a PR I come across a kind-of related thing that I think needs fixing, then I generally create a separate branch, separate commit, and separate PR. This makes it much easier to review the pull requests, and to reason about the changes.

Often when making a change, it will really be two changes: first, refactor the code to make it easier to make the change, then make the change. Then I will have two commits in the same PR. Or, sometimes the refactoring really is several distinct refactorings that are related, but not the same. Then I will put each one into its own commit.

Does this make sense?

rchavarria:
Yeah! Absolutely. And I agree. But I think I splitted one change (make it easier to run tests) into much smaller tasks (add Node.js dependencies file, add Gulp script,...). And those tasks where too narrow to be in the PR scope.

Thanks to share your thoughts.

kytrinyx:
"add node js dependencies" isn't really an atomic change in terms of the project. I think it could make sense as "add dependencies to make it possible to run the tests" as one thing, and "change the problem to use new dependencies" as another.

rchavarria:
Yep. Those are much better names for the tasks. I see now that I was too focused on the technical side of the task.

Very useful feedback, thanks.

kytrinyx:
My pleasure. Thanks again for the contribution.

PR #14, opened 2015-07-08
Configure Travis CI to run all tests

rchavarria:
Configure Travis CI to run the `make test` command that will run
all tests. Jasmine skipped tests are enabled by replacing `xit`
methods by `it` ones.

And the `README.md` file shows correctly the icon about the
Travis build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # assignments
 ASSIGNMENT ?= ""
 IGNOREDIRS := "^(\.git|bin|node_modules)$$"
-ASSIGNMENTS = $(shell find . -maxdepth 1 -mindepth 1 -type d -exec basename -a {} + | sort | grep -Ev $(IGNOREDIRS))
+ASSIGNMENTS = $(shell find . -maxdepth 1 -mindepth 1 -type d | cut -d'/' -f2 | sort | grep -Ev $(IGNOREDIRS))
 
 # output directories
 TMPDIR ?= "/tmp"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 # assignments
 ASSIGNMENT ?= ""
 IGNOREDIRS := "^(\.git|bin|node_modules)$$"
-ASSIGNMENTS = $(shell find . -maxdepth 1 -mindepth 1 -type d -exec basename -a {} + | sort | grep -Ev $(IGNOREDIRS))
+ASSIGNMENTS = $(shell find . -maxdepth 1 -mindepth 1 -type d | cut -d'/' -f2 | sort | grep -Ev $(IGNOREDIRS))
```
rchavarria: The `-exec basename -a` part was introduce by @wilmoore but it failed on Travis CI. So I decided to copy the line from the [xjavascript Makefile, line 4](https://github.com/exercism/xjavascript/blob/master/Makefile). So far, it works, but there is only one assignment. I hope it will continue working with more assignments.

If there is a better option, I don't mind changing it.

kytrinyx: I'm not against changing it per se, but I would be curious to understand why basename failed on CI.

kytrinyx:
Ok, here are earlier PRs/discussions:

- https://github.com/exercism/xjavascript/pull/60
- https://github.com/exercism/xjavascript/pull/24
- https://github.com/exercism/xjavascript/pull/59

rchavarria:
The failing part in the `find` command was `-exec basename -a {} +`. And this was the error output as it can be seen at [build 8 on CI](https://travis-ci.org/rchavarria/xecmascript/builds/70107934):

```bash
$ make test
basename: invalid option -- 'a'
Try `basename --help' for more information.
```

It seems the `basename` command is available but option `-a` is not. I forced Travis to run `basename --help` to see what happens, see [build 14 on CI](https://travis-ci.org/rchavarria/xecmascript/builds/70190186):

```bash
$ basename --help
Usage: basename NAME [SUFFIX]
  or:  basename OPTION
Print NAME with any leading directory components removed.
If specified, also remove a trailing SUFFIX.
      --help     display this help and exit
      --version  output version information and exit
Examples:
  basename /usr/bin/sort       Output "sort".
  basename include/stdio.h .h  Output "stdio".
```

There is no `-a` option available. According to the [GNU basename documentation](https://www.gnu.org/software/coreutils/manual/html_node/basename-invocation.html), that option means `basename` accepts more than one argument. But it doesn't seem to be the case for the command available in Travis CI.

Why `-a` option is not available? I don't know. Build's logs say:

> ...job is running on a container-based infrastructure

But the documentation talks about `sudo` and other commands but it says nothing about `basename`.

kytrinyx:
Yeah, that makes sense. Thanks!

PR #15, opened 2015-07-17
Add new assignment: hello-world

rchavarria:
This PR adds a new assignment for ECMAScript: hello-world. It's a very simple exercise, copied from exercism.io JavaScript track. This exercise is configured before the existing one because it's simpler.

Additionally, it changes the way to compile ECMAScript and to run tests during development time, in Travis CI. Now, it uses Gulp tasks, same as individual exercises do, to compile and run tests.

There was a problem with `traceur` and `jasmine-node`. Compiling code with a `traceur` command and running tests with `jasmine-node` caused this error:

```bash
$ make test
\e[4mRunning tests for hello-world assignment\e[0m
Exception loading: /tmp/hello-world.jBtjpsXi1A/hello-world_test.spec.js
[ReferenceError: $traceurRuntime is not defined]
ReferenceError: $traceurRuntime is not defined
    at Object.<anonymous> (/tmp/hello-world.jBtjpsXi1A/hello-world.js:1:80)
    ...
```

`traceur` included a reference to `$traceurRuntime` even before the code could require the `traceur-runtime.js` file (where that variable is initialized). Google, Stack Overflow or Traceur documentation say nothing about this issue.

```javascript
"use strict";
$traceurRuntime.options.symbols = true;
require('./traceur-runtime');
var HelloWorld = $traceurRuntime.initTailRecursiveFunction(function() {
  //...
```

There is another tool (`npm` module) to run tests, [`jasmine`](https://www.npmjs.com/package/jasmine-node). It needs a JSON configuration file, where `traceur-runtime.js` can be configured to be loaded before running the tests. And it works for individual exercises, but it doesn't for all exercises testing them with the `make` command. `make` command compile code and move some files to temp folders. The only way to access them is through absolute paths. And `jasmine` tool doesn't work with absolute paths. So, `jasmine` can't be used.

Everything works with Gulp tasks for individual exercises. This PR provides some Gulp tasks, that can be run from the `make` command. These tasks reads temp folders to use from the command line and they are able to use those temp folders to get the code, compile it and run the tests.


kytrinyx:
Thanks! I think having the gulp tasks is an improvement.

The more I think about it, the more I am skeptical of actually vendoring the node modules. What do you think? How hard would it be to git ignore all the node modules and then make people run a command to install dependencies?

In the future, would you please submit new assignments as separate PRs? It's really hard to review pull requests when they contain different types of changes.

rchavarria:
Thanks for merging the pull request. I wasn't sure this solution was the best, because it's quite different from the tools used with JavaScript in exercism.io, but in ECMAScript, code must be compiled and this extra step make the process a bit more complicated.

**To ignore or not to ignore node modules**

That's the question. There are arguments that support both, with pros and cons. An [intermediate point of view](https://www.quora.com/Should-I-put-node_modules-in-gitignore?share=1) seems to emerge: check in into the repository those modules you need to deploy your application, ignore those you need only for development. They recommend to check in production modules not to depend on third parties (a external module repository). 

Personally, I prefer ignoring modules. Module repositories exist for a reason, to download modules, so use them!. But now that I've read other opinions, I like very much the intermediate approach: check in only production modules. 

How hard would it be? It should be quite easy. Just a line in `.gitignore` and delete `node_modules` directory from the repository. Then, anybody will be able to install all dependencies with just one command `npm install`.

**Single PR for each new assignment**

My original idea was to do exactly that: one single PR for each new assignment. But, as I faced the problem with `traceur` and `jasmine-node` I thought it would be a good idea to make a PR with both the new assignment and the solution. If I'd made a PR only with the new assignment, tests would have failed. If I'd made a PR only with the solution, there wouldn't have been a problem to solve that justified the solution.

So, my apologies to mix several concerns in the same PR. I could have tried harder. I could have tried to explain the problem first, then PR a solution, and finally PR the new assignment. So, again, I have to thank you for accepting this PR. I can do better, I promise :smiley:

kytrinyx:
I don't know what the right answer is, but I lean towards having people do an `npm install`. It seems kind of silly to have thousands of files stored in this git repo, when the whole point is that one should be able to install dependencies with one command.

PR #18, opened 2015-07-25
Improve `gulpfile.js`

rchavarria:
This PR includes the following improvements:

1. Replace deprecated dev dependency, `gulp-clean` with `del`
2. Decide whether input and output dirs come from command line arguments or harcoded values. Hardcoded values will be used in individual assignments.
3. `make` command creates copies of `gulpfile.js` in each assignment. This way, it is as if there were no `gulpfile.js` duplication. Any change in ./gulpfile.js will be copied to each `<assignment>/gulpfile.js` so that all `gulpfile.js`s will be synchronized.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ node_modules: package.json
 
 test-assignment: node_modules
 	@printf "\e[4mRunning tests for $(ASSIGNMENT) assignment\e[0m\n"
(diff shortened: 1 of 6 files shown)


kytrinyx:
Thanks, @rchavarria!

PR #19, opened 2015-07-29
Add new assignment: Word Count

rchavarria:
Not too much to comment here, just a new assignment.

Import/export modules ECMAScript 6 feature is used and it works without extra work, so other assignments will use it too.


kytrinyx:
:+1:

PR #20, opened 2015-07-29
Use import/export modules ES6 feature

rchavarria:
Import/export modules were used successfully in PR #19. This PR use that feature for existing assignments, making existing code more ES6-ish.


kytrinyx:
Nice!

PR #22, opened 2015-08-08
Add new exercise: Anagram

rchavarria:
Just a new exercise.

### DIFF
--- a/anagram/anagram_test.spec.js
+++ b/anagram/anagram_test.spec.js
@@ -1,0 +1,82 @@
+import Anagram from './anagram.js';
+
+describe('Anagram', () => {
+
+  it("no matches", () => {
+    let subject = new Anagram("diaper");
+    let matches = subject.matches([ "hello", "world", "zombies", "pants"]);
+
+    expect(matches).toEqual([]);
+  });
+
+  xit("detects simple anagram", () => {
+    let subject = new Anagram("ant");
+    let matches = subject.matches(['tan', 'stand', 'at']);
+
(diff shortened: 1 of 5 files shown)


kytrinyx:
Thanks!

anagram/anagram_test.spec.js
```diff
@@ -0,0 +1,82 @@
...
+  xit("detects anagrams case-insensitively", () => {
+    let subject = new Anagram("Orchestra");
+    let matches = subject.matches(['cashregister', 'Carthorse', 'radishes']);
+
+    expect(matches).toEqual(['Carthorse']);
+  });
+
+  xit("does not detect a word as its own anagram", () => {
+    let subject = new Anagram("banana");
+    let matches = subject.matches(['Banana']);
```
iamvery: Curious, what is the intended behavior if the cases match. For e.g. `subject.matches(['banana'])`. Would you intend `expect(matches).toEqual([])` here as well? I'm assuming so based on the test description, but I don't think that case is covered. What do you think?

iamvery: Ah forgive my ignorance, I just realized that the combination of this test w/ the one before it _does_ cover that case. Apologies!

kytrinyx: @iamvery 👍

We have a canonical data set that tracks are welcome to (but not obliged to) follow: https://github.com/exercism/problem-specifications/blob/master/exercises/anagram/canonical-data.json#L80-L100

PR #26, opened 2015-09-07
Add new exercise: Beer song

rchavarria:
**Beer song** exercise based on the same exercise from the [JavaScript](https://github.com/exercism/xjavascript) track.

### DIFF
--- a/beer-song/beer-song_test.spec.js
+++ b/beer-song/beer-song_test.spec.js
@@ -1,0 +1,56 @@
+import Beer from './beer-song';
+
+describe('Beer', () => {
+
+  it('prints an arbitrary verse', () => {
+    var expected = `8 bottles of beer on the wall, 8 bottles of beer.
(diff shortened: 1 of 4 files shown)


kytrinyx:
Beer song and food chain are both text-heavy algorithmic songs--we should probably not have them right next to each other.

rchavarria:
Yep. It surprised me when I solved them in the JavaScript track, as a exercism.io user. That's why I put them together.

Now, reviewing the code of xjavascript, there are new exercises in the first ten positions. Let me implement rna-transcription, gigasecond and phone-number. That will do the first xecmascript exercises the same as xjavascript ones. Then, I'll modify `config.json` to reflect the same order as xjavascript.

Do you agree? If so, and if you don't mind, let's merge this PR and I'll fix `config.json` once the first ten exercises are implemented on xecmascript.

kytrinyx:
Yepp, this sounds like a good plan.

PR #28, opened 2015-09-13
Add new exercise: Gigasecond

rchavarria:
Usually, I don't change the spec's descriptions, but this time I've thought they could be more descriptive.

If you find them appropriate, I could update this exercise on the xjavascript track.

### DIFF
--- a/gigasecond/gigasecond_test.spec.js
+++ b/gigasecond/gigasecond_test.spec.js
@@ -1,0 +1,23 @@
+import Gigasecond from './gigasecond';
+
+describe('Gigasecond', () => {
+
+  it('tells a gigasecond anniversary since midnight', () => {
+    let gs = new Gigasecond(new Date(Date.UTC(2015, 8, 14)));
+    let expectedDate = new Date(Date.UTC(2047, 4, 23, 1, 46, 40));
+    expect(gs.date()).toEqual(expectedDate);
+  });
+  
+  it('tells the anniversary is next day when you are born at night', () => {
+    let gs = new Gigasecond(new Date(Date.UTC(2015, 8, 14, 23, 59, 59)));
+    let expectedDate = new Date(Date.UTC(2047, 4, 24, 1, 46, 39));
+    expect(gs.date()).toEqual(expectedDate);
+  });
+
+  xit('even works before 1970 (beginning of Unix epoch)', () => {
+    let gs = new Gigasecond(new Date(Date.UTC(1959, 6, 19, 5, 13, 45)));
+    let expectedDate = new Date(Date.UTC(1991, 2, 27, 7, 0, 25));
+    expect(gs.date()).toEqual(expectedDate);
+  });
+
+});
--- a/config.json
+++ b/config.json
@@ -7,6 +7,7 @@
     "hello-world",
     "leap",
     "bob",
+    "gigasecond",
     "rna-transcription",
     "word-count",
     "hamming",
--- a/gigasecond/example.js
+++ b/gigasecond/example.js
@@ -1,0 +1,14 @@
+const GIGASECOND_IN_MILIS = 1e9 * 1e3;
+
+export default class Gigasecond {
+
+  constructor(dateOfBirth) {
+    this.dateOfBirth = dateOfBirth;
+  }
+
+  date() {
+    let birthTime = this.dateOfBirth.getTime();
+    return new Date(birthTime + GIGASECOND_IN_MILIS);
+  }
+
+}
--- a/gigasecond/gulpfile.js
+++ b/gigasecond/gulpfile.js
@@ -1,0 +1,63 @@
+function getInputDirectory(argv) {
+  if (argv.input) {
+    return argv.input;
+  }
+  return '.';
+}
+
+function getOutputDirectory(argv) {
+  if (argv.output) {
+    return argv.output;
+  }
+  return 'traceur-output';
+}
+
+var gulp = require('gulp'),
+  jasmine = require('gulp-jasmine'),
+  traceur = require('gulp-traceur'),
+  del = require('del'),
+  argv  = require('yargs').argv,
+  inputDir = getInputDirectory(argv),
+  outputDir = getOutputDirectory(argv);
+
+// Gulp tasks definition
+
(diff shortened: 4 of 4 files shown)


gigasecond/gigasecond_test.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    expect(gs.date()).toEqual(expectedDate);
+  });
+
+  xit('tells gigasecond birth of a future person', () => {
+    let gs = new Gigasecond(new Date(2020, 3, 25));
+    let expectedDate = new Date(2052, 0, 2);
+    expect(gs.date()).toEqual(expectedDate);
+  });
+
+  xit('even works before 1970 (begginning of Unix epoch)', () => {
```
kytrinyx: I like this description

gigasecond/gigasecond_test.spec.js
```diff
@@ -0,0 +1,23 @@
+import Gigasecond from './gigasecond';
+
+describe('Gigasecond', () => {
+
+  it('tells the gigasecond birth of a living person', () => {
```
kytrinyx: I'm skeptical of this one. Being born in 1977 doesn't guarantee that they're alive now.

gigasecond/gigasecond_test.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    let gs = new Gigasecond(new Date(1977, 5, 13));
+    let expectedDate = new Date(2009, 1, 19);
+    expect(gs.date()).toEqual(expectedDate);
+  });
+
+  xit('tells gigasecond birth of a future person', () => {
```
kytrinyx: It's not really important that it's prior or future. Also, we should use timestamps instead of dates, and have at least one test case that tests for a timestamp early in the day and one late in the day so we catch if people just calculate the number of days. We have one like this for ruby: https://github.com/exercism/xruby/blob/master/gigasecond/gigasecond_test.rb

rchavarria: If the only thing that matters is the day, in the ruby example, these both tests should assert the same returned value:

1. `test_full_time_specified`
2. `test_full_time_with_day_roll_over`

Because both constructor parameter `Time.utc(2015, 1, 24, 22, 0, 0)` and `Time.utc(2015, 1, 24, 23, 59, 59)` should be converted to a value equal to `Time.utc(2015, 1, 24, 0, 0, 0)` (up to the day matters) and `Gigasecond.from()` should return the same value.

rchavarria:
Based on comments on the code, I can think of the following tests:

1. Test that given an arbitrary date, it returns a day 1 gigasecond in the future.
2. Test that given a date with hours, minutes and seconds set to 0, it returns a date 1 gigasecond in the future (this test try to prevent the use of hardcoded values in the implementation)
3. Test that given the same date as before, but setting hours, minutes and seconds to the end of the day, it returns the same date as the previous test case (this test forces the implementation to ignore hours, minutes and seconds)
4. Test that given a date before 1970, it still works.

I'm not sure if those descriptions are understable, I'll try with dates. I didn't check if returned values are correct, they are used just as examples, let's assume they're correct:

1. Given `new Date(1977, 5, 13, 0, 0, 0)` willl return `new Date(2009, 1, 19, 0, 0, 0)`.
2. Given `new Date(2015, 1, 24, 0, 0, 0)` will return `new Date(2046, 10, 2, 0, 0, 0)`.
3. Given `new Date(2015, 1, 24, 23, 59, 59)` will return `new Date(2046, 10, 2, 0, 0, 0)`.
4. Given `new Date(1959, 6, 19, 0, 0, 0)` will return `new Date(1991, 2, 27, 0, 0, 0)`.

gigasecond/gigasecond_test.spec.js
kytrinyx: Agreed, both should be timestamps. The only thing that matters is _not_ the day. It's the second. But a date assumes midnight.

kytrinyx:
I don't want to deal in days, I want to deal in gigaseconds.

rchavarria:
I pushed new changes to both tests and implementation. I hope they are meaningful tests:

1. Return a gigasecond annyversary since midnight
2. Return a gigasecond annyversary since the same day, but at the end of the day
3. Do the computation before 1970

I'm not sure tests are clear enough because of the use of `Date.UTC()`. It makes them verbose. The problem is that `Date.UTC()` does not return a `Date` object, but a timestamp you can use to create the `Date` object.

`Date.UTC()` use is mandatory. If not, the test with dates before 1970 fails when is run in different time zones. Running them in GMT+1 returns a different result that running them on UTC (Travis CI server).

gigasecond/gigasecond_test.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    expect(gs.date()).toEqual(expectedDate);
+  });
+  
+  it('tells the annyversary is next day when you are born at night', () => {
+    let gs = new Gigasecond(new Date(Date.UTC(2015, 8, 14, 23, 59, 59)));
+    let expectedDate = new Date(Date.UTC(2047, 4, 24, 1, 46, 39));
+    expect(gs.date()).toEqual(expectedDate);
+  });
+
+  xit('even works before 1970 (begginning of Unix epoch)', () => {
```
kytrinyx: s/begginning/beginning/

gigasecond/gigasecond_test.spec.js
```diff
@@ -0,0 +1,23 @@
+import Gigasecond from './gigasecond';
+
+describe('Gigasecond', () => {
+
+  it('tells a gigasecond annyversary since midnight', () => {
```
kytrinyx: s/annyversary/anniversary/

kytrinyx:
This looks good. I think that UTC is the right choice no matter what--timezones make this overly complicated. There are a couple of small typos, but I think that when those are fixed, this is ready.

rchavarria:
Sorry for my poor English, and thanks for pointing the typos out. What a shame for me!

kytrinyx:
No, there is no shame :) Thanks for taking the time to work on this!

rchavarria:
:smile: You're welcome.

PR #29, opened 2015-09-15
Add new exercise: Phone number

rchavarria:
Here is another exercise for the track.

I've also reorganized `config.json` problems property to mirror it from the javascript track.


kytrinyx:
Sweet!

PR #32, opened 2015-10-09
Fix an error that prevents users to run tests

rchavarria:
There were two node modules missing on each assignment's `package.json` file: `del` and `yargs`. They are mandatory to be able to run tests through `gulp test`.

Even some exercises lack of a `package.json` file. It has been added where missing.


kytrinyx:
Good catch!

rchavarria:
I have a question, just to know a bit more about exercism.io: now that the language track is active, how much time does it take for changes in a PR to be in production? Are they put in production automatically and *instantly*?

kytrinyx:
It's not automatic, but I try to deploy about once a day.

Eventually figuring out a way to make this automatic would be a superb fix.

The current process is:

* cd to the x-api directory
* run `bin/update`
* if the tests fail, fix
* commit, push, and deploy

The x-api lives on heroku.

rchavarria:
Agree. It would be great! 

Unfortunately, x-api is written almost completely in ruby (I don't know ruby) and I just deployed a couple of basic NodeJS applications (simple ones) on heroku. I should learn several things to be helpful on this task. I'm not your person right now :cry:

kytrinyx:
haha, no worries. I don't mind doing this manually for now, since I check in with exercism issues daily anyhow.

PR #35, opened 2015-10-14
first test omitted many sets of factors

matthewmorgan:
This first test for this exercise did not agree with the one in the JS track, omitting most of the factors for the first palindrome.  

There was also a small typo in the description for the second one.

### DIFF
--- a/palindrome-products/palindrome-products_test.spec.js
+++ b/palindrome-products/palindrome-products_test.spec.js
@@ -7,10 +7,10 @@ describe("Palindrome", function() {
     var largest = palindromes.largest;
 
     expect(largest.value).toEqual(9);
-    expect([1, 9]).toEqual(largest.factors);
+    expect([[3, 3], [1, 9]]).toContain(largest.factors);
   });
 
-  xit("largets palindrome from double digit factors", function() {
+  xit("largest palindrome from double digit factors", function() {
     var palindromes = generate({ maxFactor: 99, minFactor: 10 });
     var largest = palindromes.largest;
 
--- a/palindrome-products/example.js
+++ b/palindrome-products/example.js
@@ -1,96 +1,36 @@
-/**
- * Generate palindrom-products table and calculate largest/smallest.
- *
- * @param {Number} [minFactor]
- * Minimum Factor
- *
- * @param {Number} [maxFactor]
- * Maximum Factor
- *
- * @return {{largest: number, smallest: number}}
- */
+let isPalindrome = (num) => num.toString().split('').reverse().join('') === num.toString();
 
-function generate({ minFactor = 1, maxFactor = Number.MAX_VALUE }) {
-  let factors = range(minFactor, maxFactor);
-  let palindromes = [for (x of factors) for (y of factors) if (isPalindrome(x * y)) palindrome(x, y)];
+let Palindromes = (params) => {
+  let maxFactor, minFactor, maxProduct, minProduct, data;
+  maxFactor = params.maxFactor;
+  minFactor = params.minFactor || 1;
+  maxProduct = 1;
+  minProduct = Infinity;
+  data = [];
 
-  return  {
-    largest: palindromes.sort(comparator).pop(),
-    smallest: palindromes.shift()
+  for (let ii = minFactor; ii < maxFactor; ii++) {
+    for (let jj = ii; jj <= maxFactor; jj++) {
+      let product = ii * jj;
+      if (isPalindrome(product)) {
+        data[product] = [ii, jj];
+        maxProduct = Math.max(maxProduct, product);
+        minProduct = Math.min(minProduct, product);
+      }
+    }
   }
-}
 
-/**
- * Create a product/factor record.
- *
- * @param {Number} multiplicand
- * Multiplicand factor.
- *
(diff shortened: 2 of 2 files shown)


palindrome-products/palindrome-products_test.spec.js
```diff
@@ -7,7 +7,7 @@ describe("Palindrome", function() {
     var largest = palindromes.largest;
 
     expect(largest.value).toEqual(9);
-    expect([1, 9]).toEqual(largest.factors);
+    expect([[3, 3], [1, 9]], [[1, 9], [3, 3]]).toContain(largest.factors);
```
rchavarria: In fact, in the JS track, the expected value has an additional array level. Instead of:

```
[ [3, 3], [1, 9] ], [ [1, 9], [3, 3] ]
```

that value in JS is:

```
[ [ [3, 3], [1, 9] ], [ [1, 9], [3, 3] ] ]
```

Running the JS exercise, `largest.factors` is `[ [ 1, 9 ], [ 3, 3 ] ]` while in this track `largest.factors` is `[1, 9]`. It seems the implemented solution has some errors.

I've checked JS and Ruby tracks, and they look the same. This track looks a little different. A good test could be to copy the JS code and paste it in this track, just to see if all tests pass.

palindrome-products/palindrome-products_test.spec.js
```diff
@@ -7,7 +7,7 @@ describe("Palindrome", function() {
...
   });
 
   xit("largets palindrome from double digit factors", function() {
```
rchavarria: Don't be shy. Fix the typo :stuck_out_tongue_closed_eyes:

matthewmorgan: @rchavarria -- weird, I did fix the typo.  I must have forgotten to push the commit.  

Are you saying the test I committed is incorrect?  Sorry to be dense...

rchavarria: No, no, no. The test passes. But it's still different from the JS one. That's why I think the ECMAScript solution is not 100% correct.

The ECMAScript solution (link to the [example.js](https://github.com/exercism/xecmascript/blob/master/palindrome-products/example.js) file) must be different to the JS one (link to the [example.js](https://github.com/exercism/xjavascript/blob/master/palindrome-products/example.js)) because `largest.factors` are different. JS tests look very similar to Ruby tests, so I assume the JS tests and solution are correct. Then, the ECMAScript solution might not be 100% correct. It's not your fault, the ECMAScript solution code was there before you catch the mistake in the tests.

In order to get the solution right, I was proposing to copy & paste the JS code and then refactor it to make to more ECMAScript 2015 style.

matthewmorgan:
I see what you're saying: the broken test worked because the ```example.js``` was broken.

On closer inspection, I don't think the JS track is set up sensibly.  I think what you actually want, for both tracks, is a test like this:

```expect([[1, 9], [3, 3]]).toContain(largest.factors);```

[1,9] and [3,3] are both legitimate answers to the question 'what are the factors of 9?'

So, are you proposing to change the ```example.js``` yourself?  I can change the test and resubmit.  Let me know how you want to proceed.

rchavarria:
Feel free to change `example.js`. That will make the PR more balanced. And I think you have the context of the code, so you're in a better position than me.

The code is all yours.

matthewmorgan:
OK, I think it's set.  Please check it out and let me know if it looks good.

rchavarria:
It looks nice! Great job!

Would you mind to rebase the commits so that there is only one commit to merge?

matthewmorgan:
OK, I think I did that correctly.  It's been a while since I've used rebase.

rchavarria:
The idea is to have only one commit in the pull request. But now, there are four :cry: I struggle a lot with `rebase`. Sometimes I prefer to discard my changes and start from scratch.

I think the problem was that you worked on master, so merged changes on exercism/xecmascript master branch affected your work. And now, you have more commits in this PR.

Don't worry, I will merge this PR. It's not a big deal.

Next time, to avoid these problems, I suggest to follow this workflow:

1. Be sure your `master` branch is synchronized with exercism/xecmascript. I have a git remote called `exercism` referencing that repository. Locally, I am in `master` branch. I run `git pull exercism master`.
2. Create a new branch: `git checkout -b <new-branch>`. This prevents my work to be affected by changes in `master` branch. It can cause your work to be out of date, so don't work for too long in a branch.
3. Work on that branch. Commit as many times as you want.
4. Rebase and squash all commits on that branch into a single commit with all changes. `git rebase -i <commit-just-before-first-commit-on-branch>`
5. Push that branch to your GitHub repo: `git push origin <new-branch>`
6. Go to GitHub and create a pull request
7. Once the pull request has been merged. Go to your master branch: `git checkout master`. And update with exercism repo: `git pull exercism master`.
8. Repeat :wink:

matthewmorgan:
Thanks for the detailed instructions.  I did in fact forget to branch after forking.

BTW I have not seen the new exercises you have committed appearing.  Should they be?

rchavarria:
You should see them directly in the [repository](https://github.com/exercism/xecmascript). Then, when there is a deploy, new exercises should be available to practice.

PR #45, opened 2015-10-19
Change underscore filenames

AbdullahAlger:
In addition to the changes made to xJavaScript, I thought that the same changes would be appropriate for xECMAScript.


rchavarria:
Ooops. I submited a PR addressing the same problem. I didn't know there was someone working on it. It wasn't my intention to duplicate the work.

Both PRs are not exactly the same. I rename files as `<assignment>.spec.js`. I thought that was what @kytrinyx  and you agreed on [exercism/exercism.io#2581](https://github.com/exercism/exercism.io/issues/2581).

@AbdullahAlger created first the PR, so I don't mind if mine is rejected.

AbdullahAlger:
@rchavarria 

I simply kept test in the filename. I like `.spec.js` better, but I wasn't sure if `test` was still wanted in the filename. I don't mind deleting `test` from all the filenames because using `.spec.js` indicates that it is a test file. However, if we are going to do that, then I'd also delete the word `test` from all the JS filenames. I'd also do that for the Ruby test files as well where `_test` could be changed to `_spec`.  Just a thought.

rchavarria:
Well, each language use its own conventions for naming files. For JavaScript (and ECMAScript) I would name files as `<assignment>.spec.js`. 

I can't talk for Ruby. I don't know whether `test` or `spec` is more accepted in that language.

rchavarria:
@AbdullahAlger Do you mind to remove `test` from the names so that they become `<assignment>.spec.js`?

I agree to do that also in the JavaScript track.

For the Ruby track, sorry, I can't help you. I think you could ask opening a issue in exercism/xruby repo.

AbdullahAlger:
@rchavarria Yep, I can do that.

kytrinyx:
This looks great, thank you!

PR #50, opened 2015-10-25
Use 'const' instead of 'var'

rchavarria:
Not a big deal. 

ECMAScript introduces `let` and `const` as an alternative to `var`. Some great developers (like [John Resig on twitter](https://twitter.com/jeresig/status/657262368150396928?s=09)) recommend using `const` whenever possible. If not, use `let`.


kytrinyx:
Agreed. It's always nice to follow best practices where possible.

PR #52, opened 2015-10-25
Replace Traceur with Babel

### LINKED ISSUE #51: Replace Traceur with Babel to transpile code
*cf. Issue #37*

It seems [Babel](https://babeljs.io/) is a better choice for transpiling ES2015 code to ES5.

rchavarria:
Stop using Traceur to transpile ECMAScript 2015 code into ES5
and start using Babel. Babel has implemented more ES2015
features than any other project, even browsers.

`gulpfile.js` and `package.json` files of each exercise has changed because when users fetch an exercise in exercism.io they need the Babel dependencies and they need new Gulp tasks to use Babel instead of Traceur to transpile code.

Fixes #51

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ node_modules
 .idea
 
 */node_modules
-*/traceur-output
+*/babel-output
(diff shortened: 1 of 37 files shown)


kytrinyx:
You're on a roll!

rchavarria:
Yeah! I was enjoying my Sunday afternoon!

PR #54, opened 2015-10-27
Use ES2015 feature Set() to store unique robot names

rchavarria:
Use the new collection added in ES2015 `Set` to check that all generated names are different.


kytrinyx:
This looks good, thanks!

PR #59, opened 2015-11-03
Remove bigint

matthewmorgan:
BigInt does not need to be imported to grains.spec.js


rchavarria:
Thanks for taking care of this.

matthewmorgan:
No worries!

PR #60, opened 2015-11-04
Update gulpfile

matthewmorgan:
gulpfile has been changed to reflect switch from traceur to babel.  Makefile IGNOREDIRS adds .idea for simpletons like me who use IDEs ;)


rchavarria:
Thank you! It's much better to have the right name for the ouput dir.

PR #65, opened 2015-11-09
Provide a pattern to identify a file as a test file based on its name

rchavarria:
A test pattern can be defined in `config.json` to help exercism.io to identify files that contain tests.

That would avoid having small bugs as described on [exercism/exercism.io#2610](https://github.com/exercism/exercism.io/issues/2610).

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,6 +3,7 @@
   "language": "ECMAScript",
   "repository": "https://github.com/exercism/xecmascript",
   "active": true,
+  "test_pattern": ".*[.]spec[.]js$",
   "problems": [
     "hello-world",
     "leap",


rchavarria:
@kytrinyx I've used a pattern I think it would work, escaping dots to match exactly test file names. Let me know if you think a simpler pattern would do the same job:

- `.spec.js`
- `.*spec.js`
- Whatever

kytrinyx:
I think it's a good idea to match the period exactly, but this looks like it's double-escaping, so the pattern is a literal backslash and a wildcard dot.

rchavarria:
If the pattern is `.*\.spec\.js` (just one backslash to scape the dot) it fails on Travis CI:

```
0.22s$ bin/fetch-configlet

The command "bin/fetch-configlet" exited with 0.
0.00s$ bin/configlet .

Evaluating .
-> config.json is invalid. Try jsonlint.com
The command "bin/configlet ." exited with 1.

48.43s$ make test
...
```

Using jsonlint.com to check the JSON code:

```
Parse error on line 6:
...    "test_pattern": ".*\.spec\.js",    
-----------------------^
Expecting 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '['
```

JSON is valid  with two backslashes: `.*\\.spec\\.js`. But it seems it does not match the right file names.

rchavarria:
In [Escape dot in a regex range](http://stackoverflow.com/questions/10397968/escape-dot-in-a-regex-range) question on Stack Overflow, they say dots have not to be scaped inside a character class: `[.]`.

I tested `.*[.]spec[.]js` on http://regexpal.com and seems to work. I'm going to update the PR.

matthewmorgan:
@rchavarria is the goal to match something like below:

some-possibly-hyphenated-problem-name.spec.js

If so, maybe a closer match on the regex would be:

\w+(-\w+)+[.]spec[.]js

Which would require some alpha string of at least one letter before the
initial . and 0 or more hyphenated additional words.

What do you think?


On Thu, Nov 12, 2015 at 9:23 AM, Rubén Chavarría <notifications@github.com>
wrote:

> In Escape dot in a regex range
> <http://stackoverflow.com/questions/10397968/escape-dot-in-a-regex-range>
> question on Stack Overflow, they say dots have not to be scaped inside a
> character class: [.].
>
> I tested .*[.]spec[.]js on http://regexpal.com and seems to work. I'm
> going to update the PR.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/exercism/xecmascript/pull/65#issuecomment-156114725>.
>



-- 
Matt Morgan
Director of Technology
Dulcet Software LLC
(734) 926-8142

rchavarria:
The idea was just to match any file name ending literally as `.spec.js`. I don't mind what was before `.spec.js`.

This regexp is not to validate test file names, but to select them so that they can be shown in exercism.io UI when users are taking a look to solutions (see [exercism/exercism.io#2610](https://github.com/exercism/exercism.io/issues/2610)).

The regexp can be as complex as we want, but I think it should be the simplest that works. Regexps are always generating new kind of problems, you know :wink: 

Another problem with `\w+(-\w+)+[.]spec[.]js` is the use of `\` characters. A JSON string containing those is not valid, and Travis CI will complain.

kytrinyx:
I'm ok with either `.*[.]spec[.]js$` or `.*.spec.js$` or even `spec.js` since it's unlikely that we'll have an exercise that is something-spec.

matthewmorgan:
@rchavarria Sounds good, and thanks for the explanation about Travis.

PR #70, opened 2015-11-14
gulpfile uses const instead of var

matthewmorgan:
one more change to the gulpfile to make it a little more ES6-y.  @rchavarria let me know if this looks OK.  Thanks!

### DIFF
--- a/anagram/gulpfile.js
+++ b/anagram/gulpfile.js
@@ -12,7 +12,7 @@ function getOutputDirectory(argv) {
   return 'babel-output';
 }
 
-var gulp = require('gulp'),
+const gulp = require('gulp'),
   jasmine = require('gulp-jasmine'),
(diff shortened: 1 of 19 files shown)


rchavarria:
I agree with the change.

There is a little thing. When you run the command `make`, to run all tests, the script copies the `gulpfile.js` file in the root directory to every exercise directory. Would you mind to run `make` and add those changes to the PR?

If you can't run `make` I can do it for you. There is no problem.

matthewmorgan:
@rchavarria should be all set.

rchavarria:
Thanks for all the work. Nice!

Again, it appears as *Closed*. I don't get it yet.

matthewmorgan:
I was having the same problem earlier...I couldn't figure it out.  The only way I could get the PR to show as 'merged' was by using the Big Green Button.

PR #73, opened 2015-11-17
new exercise circular-buffer

matthewmorgan:
@rchavarria here you go!  ~~I also updated the gulpfile a bit to use arrow functions.~~ The gulpfile does not like arrow functions in Travis.  Reverted.


matthewmorgan:
@rchavarria reverted the changes to the gulpfile, and unintentionally merged this.  Thanks for the feedback.

Issue #44, opened 2015-10-19
Configure ESLint to lint source code at development time


rchavarria:
As discussed in [exercism/xjavascript#101](https://github.com/exercism/xjavascript/issues/101) and [exercism/xjavascript#26](https://github.com/exercism/xjavascript/issues/26), ESLint seems to be a good tool to lint JavaScript and ECMAScript source code.

Add it and configure it to lint source code at development time.

rchavarria:
[exercism/docs#54](https://github.com/exercism/docs/issues/54) highly recommends the use of linters for all languages.

rchavarria:
ESLint tool has been added in #81. Now, it's time to decide what should be linted and in what way. ESLint has lots of [rules](http://eslint.org/docs/rules/) that can be checked, organized in several groups:

- [ ] [Possible errors](http://eslint.org/docs/rules/#possible-errors)
- [ ] [Best practices](http://eslint.org/docs/rules/#best-practices)
- [ ] [Strict mode](http://eslint.org/docs/rules/#strict-mode)
- [ ] [Variables](http://eslint.org/docs/rules/#variables)
- [ ] [Node.js and CommonJS](http://eslint.org/docs/rules/#nodejs-and-commonjs)
- [ ] [Stylistic issues](http://eslint.org/docs/rules/#stylistic-issues)
- [ ] [ECMAScript 2015](http://eslint.org/docs/rules/#ecmascript-6)

matthewmorgan:
@rchavarria At first consideration, without getting in to each of the available rules, it seems like we can omit two groupings:
1) Node.js and CommonJS: doesn't seem relevant.  Am I wrong?
2) Strict mode: this declaration is not needed in ES6, right?

Include some:
1) Possible errors (most if not all)
2) Best practices (most if not all)
3) Variables (most if not all)
4) ES 2015 (some of this seems like it belongs in 'stylistic'?)

@kytrinyx On the 'stylistic issues,' should exercism be expressing an opinion on this, or should that be left to discussion between users? Do other tracks that lint express opinions about style?

kytrinyx:
I think the stylistic questions should be left to the users in languages where there's traditionally more than one way to do things. In Python and Go, for example, it would make sense to encourage following the standard style guide. In Ruby (and probably JavaScript...) not so much!

matthewmorgan:
I think this is resolved.  Please let me know if not.  Thanks everyone!

rchavarria:
[...]

PR #81, opened 2015-11-19
Add eslint

matthewmorgan:
@rchavarria is this what you had in mind for #44 ?

### DIFF
--- a/anagram/gulpfile.js
+++ b/anagram/gulpfile.js
@@ -13,6 +13,7 @@ function getOutputDirectory(argv) {
 }
 
 const gulp = require('gulp'),
+  eslint = require('gulp-eslint'),
   jasmine = require('gulp-jasmine'),
   babel = require('gulp-babel'),
   polyfill = require('babel/polyfill'),
@@ -30,10 +31,24 @@ gulp.task('test', [ 'babel' ], function () {
     .pipe(jasmine());
 });
 
-gulp.task('babel', function () {
+gulp.task('babel', [ 'lint' ], function () {
   return gulp.src([ inputDir + '/*.js' ])
     .pipe(babel())
     .pipe(gulp.dest(outputDir));
+});
+
+gulp.task('lint', function () {
+  return gulp.src([ inputDir + '/*.js' ])
+    .pipe(eslint({
+      envs: [
+        'es6'  //turns on all es6 features except modules
+      ],
+      ecmaFeatures: {
+        'modules': true,  //this gives us modules :)
+      },
+    }))
+    .pipe(eslint.format())
+    .pipe(eslint.failAfterError());
 });
 
 gulp.task('clean', function (cb) {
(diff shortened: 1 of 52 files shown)


matthewmorgan:
@rchavarria It looks like perhaps the last merge resulted in some conflicts on this PR.  I will fix as soon as I can and ping you.

matthewmorgan:
@rchavarria should be set.  See what you think.  Thanks!

rchavarria:
About `config.json`. Right now, I can think of a workaround, and it's to edit different lines in different pull rquests. But I'm not sure it that would work. Later, in other PR we could reorganize all exercises to match the same order as xjavascript.

About ESLint. Thanks! That is exactly what I had in mind. I was planning to use a `.eslintrc` file to configure it, and to lint much more aspects of the JavaScript code. But this PR is the perfect starting point.

I would merge it, but it would cause a problem to exercism users trying to solve the exercises. The new `gulpfile.js` needs a new dependency, `gulp-eslint`. That dependency is only added to the `package.json` file in the root directory, and it should be added to every `package.json` of every individual exercise. Otherwise, users won't be able to run tests. The `gulp test` command will fail.

*Note: I'm not sure, but I think I tried to copy `package.json` as it's done with `gulpfile.js` but it didn't work. You probably need to edit each `package.json` file manually* :cry:

This made two problems arise:

1. Files `gulpfile.js`, `package.json` and `.eslintrc` (if we decide to use it in the future) are duplicated (or kind of) in every single exercise. @kytrinyx, is there a way to include those files when exercism users fetch an exercise to avoid having them copied on each exercise folder?
2. Running tests with `make` (at development time) and running tests as an exercism user are not the same process. It seems, all exercises share the same `npm` installation when running it with `make`. That's why the `make` command does not fail, but running tests as a user fails.

matthewmorgan:
@rchavarria Hmm... I tested some individual exercises using ```gulp test``` from within the exercise directories.  Must be it worked bc of the ```npm install``` for the xecmascript directory above it.

If I understand you correctly, changing all the ```package.json``` files in the individual exercises would fix things for now?  I could do that if @kytrinyx says it is complicated to fetch the global ```package.json``` file.

Also, I did intially set this up using ```.eslintrc``` in the global directory, but it seemed to be redundant with the option hash in the gulpfile.

Anyway, I'm happy to modify the individual ```package.json``` files for now-- just let me know.  Thanks!

rchavarria:
Yep, go for modifying individual `package.json` files please.

We can address other tasks in different issues/PRs: improving lint process, *duplicated* files and npm installations.

matthewmorgan:
@rchavarria there you go.  Thanks for catching that issue!

matthewmorgan:
@rchavarria I can merge this if you like. 😃  I think I made the changes as you requested.

rchavarria:
I looks totally fine now, thanks for the effort!!

Issue #84, opened 2015-11-23
Figure out a way to keep synchronized main and individual exercises `package.json` file


rchavarria:
Currently, there is no way to keep synchronized the main `package.json` file with the same file for individual exercises. It's mandatory to keep synchronized at least the `dependencies` section, because all exercises will use the same NodeJS packages to let users lint, transpile and run code tests.

Figure out a way to keep the synchronization.

matthewmorgan:
It would be possible to do this the same way as the gulpfile, by adding a
line to the makefile.

I don't know how the exercism app decides what content to deliver, that
seems like the best ultimate solution, but what about the makefile for now?


On Mon, Nov 23, 2015 at 4:55 PM, Rubén Chavarría <notifications@github.com>
wrote:

> Currently, there is no way to keep synchronized the main package.json
> file with the same file for individual exercises. It's mandatory to keep
> synchronized at least the dependencies section, because all exercises
> will use the same NodeJS packages to let users lint, transpile and run code
> tests.
>
> Figure out a way to keep the synchronization.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/exercism/xecmascript/issues/84>.
>



-- 
Matt Morgan
Director of Technology
Dulcet Software LLC
(734) 926-8142

rchavarria:
Yep, I think that is the easiest way for now. 

The problem is that only dependencies are strictly the content that needs to be synchronized. It would be nice to have customized values in `package.json` for fiels such as: `name`, `version`, `description`,...

That could be the final goal. Meanwhile, copying the whole file seems to be enough.

tejasbubane:
Refer https://github.com/exercism/x-common/issues/601. A lot more tracks are facing the same issue and as of this writing there is not better solution than what we have right now - Makefile to copy into each exercise directory.

rchavarria:
Should we close this issue then? Or maybe start a discussion on https://github.com/exercism/discussions/

tejasbubane:
I don't know, we either watch out the issue there and keep this open as a reminder or close this as it is not actionable (at the moment).

matthewmorgan:
#277 addresses this issue.

PR #87, opened 2015-11-25
Modify Makefile to copy package.json from root to all assignments

matthewmorgan:
Interim fix for issue #84.


rchavarria:
Small change but an important one, thanks!

PR #109, opened 2015-11-30
Add new exercise luhn

matthewmorgan:
Port exercise from JS track.  A very classy solution in honor of @rchavarria


rchavarria:
Thanks for this super classy one :smiley:

Issue #110, opened 2015-12-01
Add new exercise: Pig latin


rchavarria:
Take it from the JavaScript track

matthewmorgan:
Closed by #111.

PR #111, opened 2015-12-01
Add new exercise: Pig latin

rchavarria:
My small contribution, a new exercise, as described at #110


matthewmorgan:
@rchavarria thanks!  Awesome regex-fu in that `example.js`.  Was that you?

rchavarria:
Nope, it was in the JavaScript exercise. I'm not sure I could get to that regex.

Issue #112, opened 2015-12-02
Default params for Hello World. undefined vs empty?


matthewmorgan:
So, I proposed to a team member that he should try using default params in hello-world.  However, it doesn't work because the first test passes an empty string rather than an undefined value.  Here's his solution: http://exercism.io/submissions/35fced9768e840df98afa4b85fbfac56

Should the first test be changed to pass an undefined value to allow the use of default params?  EG
current:
```
  it('says hello world with no name', () => {
    expect(helloWorld.hello('')).toEqual('Hello, World!');
  });
```
maybe?:
```
  it('says hello world with no name', () => {
    expect(helloWorld.hello()).toEqual('Hello, World!');
  });
```
@kytrinyx @rchavarria what do you think? What do the other tracks do with empty but not undefined values?

kytrinyx:
Ruby doesn't test for an empty string, it only does no parameter (defaults to "World") and a specific parameter: https://github.com/exercism/xruby/blob/master/hello-world/hello_world_test.rb#L18-L31

I think the proposed change makes more sense.

matthewmorgan:
OK submitted as PR #113 Thanks!

rchavarria:
JavaScript track passes an empty string. But your proposal is better, because it allows the user to use a ES2015 functionality (the goal of this language track).

rchavarria:
Closed by #113

PR #113, opened 2015-12-02
Change first test to pass undefined name

matthewmorgan:
Per #112

### DIFF
--- a/hello-world/example.js
+++ b/hello-world/example.js
@@ -1,7 +1,6 @@
 
 class HelloWorld {
-    hello(name) {
-        name = name || 'World';
+    hello(name = 'World') {
         return `Hello, ${name}!`;
     }
 }
--- a/hello-world/hello-world.js
+++ b/hello-world/hello-world.js
@@ -4,7 +4,7 @@
 //
 
 class HelloWorld {
-  hello(name) {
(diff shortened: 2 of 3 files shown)


rchavarria:
Now that the test has changed, what if the implementation on `example.js` is changed and it uses *optional parameters*?

exercism.io users won't see that file, but the implementation would be better.

matthewmorgan:
Like this?  As it happens this was the implementation I was trying to suggest to my colleague that failed against the older version of the test!  And then I didn't write `example.js` that way...

kytrinyx:
I like this.

rchavarria:
Sorry @matthewmorgan, but `hello-world.js` file has been deleted. I think it's important to keep it in this very first exercise, to help users to get started as fast as possible. 

I know that the provided file almost solve the problem. Maybe, if the `name` parameter is removed from the `hello()` method... It would make the exercise a little harder to solve.

matthewmorgan:
@rchavarria I thought I had left that in by mistake, so I removed it!  Just an oversight on my part.

rchavarria:
No problem, that's the goal of code reviews and PR.

rchavarria:
Nice!

PR #114, opened 2015-12-02
Add new exercise pythagorean-triplet

matthewmorgan:
Ported from JS Track.  @rchavarria, this `example.js` has _two_ classes 😄

### DIFF
--- a/config.json
+++ b/config.json
@@ -39,6 +39,7 @@
     "trinary",
     "sieve",
     "simple-cipher",
+    "pythagorean-triplet",
     "luhn",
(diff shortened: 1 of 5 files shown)


rchavarria:
Extriclty speaking, two classes in a single file are too many classes. 

So far, all implementations are included in a single file, so the provided solution is ok, no problem with that. And it's not so bad, the `Triplets` class is not exported, so tests know nothing about that.

rchavarria:
+1 for that Pythagorean classes

Issue #43, opened 2015-10-19
Create About Language Section


rchavarria:
Add an `ABOUT.md` file in the path specified in [exercism/x-common#110](https://github.com/exercism/x-common/issues/110) to show a new section in the language track: About.

matthewmorgan:
Closed by #115

PR #115, opened 2015-12-02
Add ABOUT.md to docs

matthewmorgan:
As mentioned in #43 @rchavarria would something like this work for the ABOUT.md?

### DIFF
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -1,0 +1,5 @@
+ECMAScript 2015 is a new version of the standard that defines JavaScript, the language that allows web pages to be dynamic. It executes in modern web browsers, and can also be executed on servers where the NodeJS platform is installed. JavaScript/ECMAScript is an interpreted language, which means that it doesn't need to be compiled.
+
+_ECMAScript has grown to be one of the world’s most widely used general purpose programming languages. It is best known as the language embedded in web browsers but has also been widely adopted for server and embedded applications._ --[ECMA International Language Specification](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-ecmascript-overview)
+
+The current revision of ECMAScript is ECMAScript 2015, previously known as ES6 or Harmony.  This version brings many enhancements over ES5, including notably template strings, expressive arrow function syntax, and cleaner syntax for defining classes.  Support in current browsers is incomplete, and often requires transpilation with a tool like [Babel](https://babeljs.io/).   


kytrinyx:
Could we start out with a sentence that gives the highest-level context and then give details farther down?

Imagine someone never heard of ECMAScript and only has a vague idea of how JavaScript works (e.g. someone like me). What would you say to them about what ECMAScript is? Over drinks. Nobody from the standards committee is listening.

docs/ABOUT.md
```diff
@@ -0,0 +1,5 @@
+_The ECMAScript specification is a standardized specification of a scripting language developed by Brendan Eich of Netscape; initially it was named Mocha, later LiveScript, and finally JavaScript._ --[Wikipedia](https://en.wikipedia.org/wiki/ECMAScript)
```
rchavarria: There are two occurrences of *specification*. If the paragraph is identical to the source (Wikipedia) there's no problem, but I would start the paragraph this way: *ECMAScript is a standardized specification...*

docs/ABOUT.md
```diff
@@ -0,0 +1,5 @@
...
+
+_ECMAScript has grown to be one of the world’s most widely used general purpose programming languages. It is best known as the language embedded in web browsers but has also been widely adopted for server and embedded applications._ --[ECMA International Language Specification](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-ecmascript-overview)
+
+The current revision of ECMAScript is ECMAScript 2015, also known as ES6 or Harmony.  This version brings many enhancements over ES5, including notably template strings, expressive arrow function syntax, and cleaner syntax for defining classes.  Support in current browsers is incomplete, and often requires transpilation with a tool like [Babel](https://babeljs.io/).
```
rchavarria: Despite of *..., also known as ES6...* I would say *..., previously known as ES6...*

kytrinyx:
Can we add a paragraph at the very top that is conversational? When I read the wikipedia quote it doesn't tell me why I would want to learn ECMAScript, and I had to read it very, very closely to figure out that it had anything to do with JavaScript.

Something like:

_ECMAScript is JavaScript, except better, because [x, y, z]._ (what were the problems with javascript? How is it different? Do I still write it for the browsers? Do the browsers understand if I write it?)

rchavarria:
I'll try to write a first paragraph as @kytrinyx recommends:

> ECMAScript 2015 is a new version of the standard that defines JavaScript. It's the new version of the language that allows web pages to be dynamic. It executes in every modern browser, without exception. It can also be executed in servers where the NodeJS platform is installed because JavaScript is an interpreted language. That means that it doesn't need to be compiled. Browsers and NodeJS have the ability to to read, understand and execute JavaScript code.

I have to say that I also like the Wikipedia definition. It sound too formal, but it's ok.

kytrinyx:
I think we can have the wikipedia definition, but... after a conversational one so that people know what they're reading.

rchavarria:
I'll let @matthewmorgan to mix our suggestions with the text he provided :stuck_out_tongue_winking_eye: Please @matthewmorgan, don't get too angry.

matthewmorgan:
@rchavarria @kytrinyx I like the tone of @rchavarria's intro, except it might be confusing to say 'It executes in every modern browser, without exception.'  There are [holes in the support](https://kangax.github.io/compat-table/es6/) even in Chrome(65% support) and Firefox(74%.)  Safari is currently at a measly 54%.

I think I'd like to link at the end  to the compatibility table I referenced above.  Other than that, I think @rchavarria's intro should replace the one from Wikipedia, which doesn't really provide much useful info to someone new to the language.

What do you guys think?

matthewmorgan:
How about this:

>ECMAScript 2015 is a new version of the standard that defines JavaScript, the language that allows web pages to be dynamic. It executes in modern web browsers, and can also be executed on servers where the NodeJS platform is installed. JavaScript/ECMAScript is an interpreted language, which means that it doesn't need to be compiled.
>
>_ECMAScript has grown to be one of the world’s most widely used general purpose programming languages. It is best known as the language embedded in web browsers but has also been widely adopted for server and embedded applications._ --[ECMA International Language Specification](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-ecmascript-overview)
>
>The current revision of ECMAScript is ECMAScript 2015, previously known as ES6 or Harmony.  This version brings many enhancements over ES5, including notably template strings, expressive arrow function syntax, and cleaner syntax for defining classes.  Support in current browsers is incomplete, and often requires transpilation with a tool like [Babel](https://babeljs.io/).   
>For reference here is a [table outlining what will currently run where](https://kangax.github.io/compat-table/es6/).

kytrinyx:
I like this @matthewmorgan

matthewmorgan:
OK thanks @kytrinyx!  Between your feedback and @rchavarria's, I feel comfortable merging this one myself.

Issue #118, opened 2015-12-03
Add exercise difference-of-squares


matthewmorgan:
I will port this from the JS track.

rchavarria:
Closed by #119

PR #119, opened 2015-12-04
Add new exercise difference-of-squares

matthewmorgan:
As described in #118

### DIFF
--- a/difference-of-squares/example.js
+++ b/difference-of-squares/example.js
@@ -1,0 +1,25 @@
+export default class Squares {
+
+  constructor(max) {
+    this.max = max;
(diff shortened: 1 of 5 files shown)


difference-of-squares/example.js
```diff
@@ -0,0 +1,25 @@
+export default class Squares {
+
+  constructor(max) {
+    this.max = max;
+  }
+
+  get squareOfSums() {
```
rchavarria: Yeah! I like this and the rest of *getters*.

rchavarria:
Thank you very much for this, great!

Issue #122, opened 2015-12-04
Add new exercise linked-list


matthewmorgan:
I will port this from the JS track.

rchavarria:
Closed by #123

PR #123, opened 2015-12-04
Add new exercise linked-list

matthewmorgan:
As described in #122

### DIFF
--- a/linked-list/linked-list.spec.js
+++ b/linked-list/linked-list.spec.js
@@ -1,0 +1,79 @@
+import LinkedList from './linked-list';
+
+describe('LinkedList', () => {
+  it('add/extract elements to the end of the list with push/pop', () => {
+    const list = new LinkedList();
+    list.push(10);
+    list.push(20);
+    expect(list.pop()).toBe(20);
+    expect(list.pop()).toBe(10);
+  });
+  xit('extract elements from the beginning of the list with shift', () => {
+    const list = new LinkedList();
+    list.push(10);
+    list.push(20);
+    expect(list.shift()).toBe(10);
+    expect(list.shift()).toBe(20);
+  });
+  xit('add/extract elements from the beginning of the list with unshift/shift', () => {
+    const list = new LinkedList();
+    list.unshift(10);
+    list.unshift(20);
+    expect(list.shift()).toBe(20);
+    expect(list.shift()).toBe(10);
+  });
+  xit('unshift/pop', () => {
+    const list = new LinkedList();
+    list.unshift(10);
+    list.unshift(20);
+    expect(list.pop()).toBe(10);
+    expect(list.pop()).toBe(20);
+  });
+  xit('example', () => {
+    const list = new LinkedList();
+    list.push(10);
+    list.push(20);
+    expect(list.pop()).toBe(20);
+    list.push(30);
+    expect(list.shift()).toBe(10);
+    list.unshift(40);
+    list.push(50);
+    expect(list.shift()).toBe(40);
+    expect(list.pop()).toBe(50);
+    expect(list.shift()).toBe(30);
+  });
+  xit('pops undefined when there are no elements', () => {
+    const list = new LinkedList();
+    expect(list.pop()).toBe(undefined);
+  });
+  xit('can count its elements', () => {
+    const list = new LinkedList();
+    expect(list.count()).toBe(0);
+    list.push(10);
+    expect(list.count()).toBe(1);
+    list.push(20);
+    expect(list.count()).toBe(2);
+  });
+  xit('deletes the element with the specified value from the list', () => {
+    const list = new LinkedList();
+    list.push(10);
+    list.push(20);
+    list.push(30);
+    list.delete(20);
+    expect(list.count()).toBe(2);
+    expect(list.pop()).toBe(30);
+    expect(list.shift()).toBe(10);
+  });
+  xit('deletes the only element', () => {
+    const list = new LinkedList();
+    list.push(10);
+    list.delete(10);
+    expect(list.count()).toBe(0);
+  });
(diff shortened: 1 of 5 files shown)


linked-list/linked-list.spec.js
```diff
@@ -0,0 +1,72 @@
+import LinkedList from './linked-list';
+
+describe('LinkedList', () => {
+  it('push/pop', () => {
```
rchavarria: Could a better description be: `add/extract elements to the end of the list with push/pop`?

linked-list/linked-list.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    const list = new LinkedList();
+    list.push(10);
+    list.push(20);
+    expect(list.pop()).toBe(20);
+    expect(list.pop()).toBe(10);
+  });
+  xit('push/shift', () => {
```
rchavarria: Could a better description be: `extract elements from the beginning of the list with shift`?

linked-list/linked-list.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    const list = new LinkedList();
+    list.push(10);
+    list.push(20);
+    expect(list.shift()).toBe(10);
+    expect(list.shift()).toBe(20);
+  });
+  xit('unshift/shift', () => {
```
rchavarria: Could a better description be: `add/extract elements from the beginning of the list with unshift/shift`?

linked-list/linked-list.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    list.push(50);
+    expect(list.shift()).toBe(40);
+    expect(list.pop()).toBe(50);
+    expect(list.shift()).toBe(30);
+  });
+  xit('pops undefined when there are no elements', () => {
+    const list = new LinkedList();
+    expect(list.pop()).toBe(undefined);
+  });
+  xit('can count xits elements', () => {
```
rchavarria: Is `xits` a typo? Should it be `list`?

linked-list/linked-list.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    const list = new LinkedList();
+    expect(list.count()).toBe(0);
+    list.push(10);
+    expect(list.count()).toBe(1);
+    list.push(20);
+    expect(list.count()).toBe(2);
+  });
+  xit('deletes the last element from the list', () => {
```
rchavarria: The `delete` implementation deletes the element with the specified value, not the last element. The test description should reflect this.

linked-list/linked-list.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    list.delete(20);
+    expect(list.count()).toBe(2);
+    expect(list.pop()).toBe(30);
+    expect(list.shift()).toBe(10);
+  });
+  xit('deletes the only element', function() {
+    const list = new LinkedList();
+    list.push(10);
+    list.delete(10);
+  });
```
rchavarria: A expectation is missing here:

```
expect(list.count()).toBe(0);
```

linked-list/linked-list.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    list.push(30);
+    list.delete(20);
+    expect(list.count()).toBe(2);
+    expect(list.pop()).toBe(30);
+    expect(list.shift()).toBe(10);
+  });
+  xit('deletes the only element', function() {
+    const list = new LinkedList();
+    list.push(10);
+    list.delete(10);
```
rchavarria: A expectation is missing at the end of the test:

```
expect(list.count()).toBe(0);
```

linked-list/linked-list.spec.js
rchavarria: A new test could be added about `delete`:

```
xit('deletes does not modify the list if the element is not found', () => {
  const list = new LinkedList();
  list.push(10);
  list.delete(20);
  expect(list.count()).toBe(1);
});
```

linked-list/linked-list.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    const list = new LinkedList();
+    list.push(10);
+    list.push(20);
+    list.push(30);
+    list.delete(20);
+    expect(list.count()).toBe(2);
+    expect(list.pop()).toBe(30);
+    expect(list.shift()).toBe(10);
+  });
+  xit('deletes the only element', function() {
```
rchavarria: An arrow function could be used here.

rchavarria:
I commented on some lines. They could be suggestions for the xjavascript track. If so, I can take care of it. I'll wait for your opinion.

matthewmorgan:
I think we can change this-- does it make sense to change the test as a separate PR?  Either way is fine with me.

linked-list/linked-list.spec.js
kytrinyx: A bit of a tangent, but my guess is that it was `its` and a search/replace on `it` with `xit` turned it into `xits`.

matthewmorgan: I thought it was maybe Catalan...

rchavarria:
I think it would be better changed and incorporated in this PR. The better the code, the better the PR :wink:

linked-list/linked-list.spec.js
rchavarria: It looks so obvious to me now that it was `its`, that I don't understand why I didn't see it before.

Could this test description be changed to `can count its elements`?

matthewmorgan:
I will make the change.

rchavarria:
Thanks for take care of them.

While merging the PR, I deleted a `console.log` statement.

matthewmorgan:
Oops...thanks!

PR #124, opened 2015-12-05
Add new exercise wordy

matthewmorgan:
As described in #125

### DIFF
--- a/wordy/example.js
+++ b/wordy/example.js
@@ -1,0 +1,45 @@
+const ArgumentError = () => ({
+  name:    "argument error",
+  message: "oops"
+});
+
+const re = new RegExp(/(plus|minus|divided by|multiplied by)+/g);
+
+class Wordy {
+
+  constructor(question) {
+    this.numbers = question.match(/[-]{0,1}\d+/g);
+    this.operands = question.match(re);
+  }
+
+  answer() {
+    if (!this.numbers || !this.operands) {
+      throw (new ArgumentError());
+    }
+    let ii = 1,
+        jj = 0,
+        result = +this.numbers[0];
+
+    while (ii < this.numbers.length + 1) {
+      const op = this.operands[jj++],
+            b = +this.numbers[ii++] || null;
+      switch (op) {
+        case 'plus' :
+          result += b;
+          break;
+        case 'minus' :
+          result -= b;
+          break;
+        case 'multiplied by' :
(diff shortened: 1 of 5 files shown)


wordy/wordy.spec.js
```diff
@@ -0,0 +1,89 @@
+import { WordProblem, ArgumentError } from './wordy';
+
+describe('Word Problem', () => {
+
+  it('add 1', () => {
+    const question = 'What is 1 plus 1?';
+    expect(2).toEqual(new WordProblem(question).answer());
```
rchavarria: In the rest of tests so far, the expect line is like:

```
expect(new WordProblem(question).answer()).toEqual(2);
```

In fact, somebody opened a issue in xjavascript about swapped *expected* and *actual* value. This can be ported to xjavascript track, for this exercise.

wordy/example.js
```diff
@@ -0,0 +1,53 @@
...
+      throw (new ArgumentError());
+    }
+    let ii = 1,
+      jj = 0,
+      result = +this.numbers[0];
+
+    while (ii < this.numbers.length + 1) {
+      const op = this.operands[jj++],
+        b = +this.numbers[ii++] || null;
+      switch (op) {
```
rchavarria: Is it needed to wrap each `case` in curly braces? Wouldn't it be simpler:

```
case 'plus':
  result += b;
  break;
case 'minus':
//...
```

matthewmorgan:
I would ask the same question about this as for the linked-list: does it make sense to do this as part of porting the exercise, or as its own PR?  What do you think?

rchavarria:
These kind of PR are *special* ones, because we're taking code from another track and adapting it to this track. 

During the process, if I see some code that I can improve, I improve it. I do this because the code in this track can be considered as independent of the code in xjavascript. These improvements can be a detail that ES2015 makes easier to do, better variable names,... 

From my point of view, if there are small improvements that don't change dramatically the code of the tests or the solution while adding the new exercise, I would include them in the first PR. If there is something I want to discuss, I'll first add the exercise, and then, I'll open a issue on the subject.

matthewmorgan:
I'll change this one and resubmit.

kytrinyx:
I would consider each track to be independent and fix anything about an exercise in the first PR.

rchavarria:
Fantastic! A new exercise!

PR #134, opened 2015-12-07
Normalize YAML in Travis CI config

kytrinyx:
The YAML was not quite valid. Travis can read it
and wasn't complaining, but this ensures that the YAML
is clean.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,9 @@
+---
 language: node_js
-
 node_js:
-  - "4.1"
-
+- '4.1'
 sudo: false
-
 script:
-  - bin/fetch-configlet
-  - bin/configlet .
-  - make test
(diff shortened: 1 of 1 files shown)


rchavarria:
As Travis CI wasn't complaining I thought it was a valid one.

I didn't know all yaml file should start with three dashes. And, I thought lists (as the three elements in `script`) needed to be indented.

Thanks for fixing it.

kytrinyx:
Actually, someone else pointed out that it was valid, maybe those dashes are optional. Anyway, I ran it through a tool, so that I can more easily script changes later :)

rchavarria:
:+1: No problem. In fact, if there is a tool that analyses yaml files, I trust them more than myself :wink:

kytrinyx:
:)

Issue #126, opened 2015-12-05
Add new exercise hexadecimal


matthewmorgan:
I will port this from the JS track

rchavarria:
Closed by #135.

PR #135, opened 2015-12-07
Add new exercise hexadecimal

matthewmorgan:
As described in #126

### DIFF
--- a/config.json
+++ b/config.json
@@ -43,6 +43,7 @@
     "luhn",
     "kindergarten-garden",
     "pig-latin",
+    "hexadecimal",
     "pythagorean-triplet",
     "series",
     "wordy",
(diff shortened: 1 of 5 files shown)


rchavarria:
What do you think if the last three test descriptions are changed to something similar to this?

```
xit('HTML/CSS represents <color> with hexadecimal value <hex-value>', () => { ... }
```

matthewmorgan:
I was thinking the same thing, but had sort of enjoyed the 'puzzle' of figuring out the color reference when I originally did the exercise.  

Either way is fine with me.

rchavarria:
I see. Maybe a less explicit description:

```
xit('hexadecimal value for HTML/CSS <white/black/yellow> color', () => { ... }
```

matthewmorgan:
I think that's fine.  I'll make the change now.

rchavarria:
Nice!!

PR #137, opened 2015-12-07
Add new exercise: Binary search

rchavarria:
This PR adds a new exercise from the xjavascript track.

### DIFF
--- a/binary-search/example.js
+++ b/binary-search/example.js
@@ -1,0 +1,39 @@
+function isSortedArray(array) {
+  for (let i = 1; i < array.length; i++) {
+    if (array[i] < array[i - 1]) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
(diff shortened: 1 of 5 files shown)


binary-search/example.js
```diff
@@ -0,0 +1,39 @@
...
+    if (array[i] < array[i - 1]) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+function recursiveSearch(array, value, start, end) {
+  if (start == end) return -1;
```
matthewmorgan: Should this use `===` instead of `==`?

matthewmorgan:
@rchavarria I am noticing the test order actual/expected is reversed on most of the tests.

binary-search/example.js
rchavarria: Totally agree

rchavarria:
You're right. I didn't notice that.

Issue #140, opened 2015-12-07
Add test to robot-name to make sure internal state can't be modified


matthewmorgan:
The robot-name exercise on this track should not allow the internal state (eg robot's name) to be modified from the outside.

PR #141, opened 2015-12-07
Add new test to robot name to prevent modification of internal state

matthewmorgan:
As described in #140.  This also required changing the solution, which returned the actual variable containing the internal state.

### DIFF
--- a/robot-name/example.js
+++ b/robot-name/example.js
@@ -1,47 +1,27 @@
-const HOW_MANY_CHARS = 26;
-const HOW_MANY_NUMBERS = 1000;
-const CHAR_CODE_A = 'A'.charCodeAt(0);
+const ALPHA = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ',
+      BASE = 10,
+      usedNames ={};
 
-let generatedNames = {};
+const random = (max) => Math.floor(Math.random()*max);
 
-function generateName() {
-  let name = '';
-  name += generateRandomChar();
-  name += generateRandomChar();
-  name += pad(generateRandomNumber());
-
-  return getNotGeneratedName(name);
-}
-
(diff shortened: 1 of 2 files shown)


robot-name/example.js
```diff
@@ -1,47 +1,27 @@
...
+  }
+
+  get name() {
+    return this.robotName;
   }
 
   reset() {
-    this.name = generateName();
+    this.robotName = generateName();
+    return this;
```
rchavarria: Is `reset()` required to return `this`? If specs don't use the returned value, why not just simple return nothing?

robot-name/example.js
```diff
@@ -1,47 +1,27 @@
...
-}
-
-function getNotGeneratedName(name) {
-  if (generatedNames[name]) {
-    return generateName();
-  }
-
-  generatedNames[name] = true;
+const generateName = () => {
+  let name = alpha.charAt(random(26))
```
rchavarria: Should we extract the magic number `26` to a constant calculated from `alpha`'s length?

robot-name/example.js
```diff
@@ -1,47 +1,27 @@
...
+    +alpha.charAt(random(26))
```
rchavarria: Should we extract `10` to a constant?

robot-name/example.js
matthewmorgan: Yes.

robot-name/example.js
matthewmorgan: Good point.  I made that change too.

matthewmorgan:
Forgot to change '10'.

rchavarria:
Yep! I was reviewing it. I'll let you change it.

rchavarria:
Great!

Issue #148, opened 2015-12-07
Add new exercise: nth prime


rchavarria:
Take it from xjavascript track.

rchavarria:
Closed by #149

PR #149, opened 2015-12-07
Add new exercise: Nth prime

rchavarria:
As described in #148

### DIFF
--- a/config.json
+++ b/config.json
@@ -35,6 +35,7 @@
     "atbash-cipher",
     "accumulate",
     "crypto-square",
+    "nth-prime",
     "trinary",
     "sieve",
     "simple-cipher",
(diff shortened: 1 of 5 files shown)


matthewmorgan:
@rchavarria looks good to me.  Only thing I saw was the `possiblePrimes` could be `const` I believe.  Could be merged now, or make that one change if you like.

rchavarria:
I'll do the change.

rchavarria:
Looking at the code, I prefer to keep it as `let`. Technically, it can be `const` and there won't be an error. But as `possiblePrimes` is a variable that changes its value (new primes are added to the list), `let` seems more appropriate.

I remember I had a conversation about this with an exercism user, and I liked a lot his reasons: if it's a variable that changes its value (i.e.: a list where elements are added), `let` seems a better option.

matthewmorgan:
Yes, for objects the reference is the 'constant' bit rather than the value.  I don't have any issue with it myself!

rchavarria:
Ok, first PR merged. This is the easiest one.

Issue #152, opened 2015-12-10
Add new exercise say


matthewmorgan:
I will port this from the JS track

rchavarria:
Closed by #153

PR #153, opened 2015-12-10
Add new exercise say

matthewmorgan:
As described #152

### DIFF
--- a/config.json
+++ b/config.json
@@ -25,6 +25,7 @@
     "triangle",
     "clock",
     "scrabble-score",
+    "say",
     "roman-numerals",
     "circular-buffer",
     "binary",
--- a/say/example.js
+++ b/say/example.js
@@ -1,0 +1,79 @@
+const smallNumbers = {
+  0:  'zero',
+  1:  'one',
+  2:  'two',
+  3:  'three',
+  4:  'four',
+  5:  'five',
+  6:  'six',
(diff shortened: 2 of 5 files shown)


rchavarria:
Thanks for this one, it's fun.

Just asking an English question: shouldn't 142 be said as "one hundred **and** forty-two"?

And, I find it weird to export a `new Say()` object in `example.js` instead of just the `Say` class. What do you think?

matthewmorgan:
As far as the english, it depends.  When talking about money, usually 'and' separates dollars from cents.  EG 'one hundred and forty-two' = 100.42 rather than 142.  Otherwise you are correct.

Exporting the new `Say()` was a pain!  I would much rather have exported a class, the implementation would be cleaner.  I think this sort of inconsistency is what we were discussing on xjavascript with https://github.com/exercism/xjavascript/issues/128.

I would be happy to change it, but it would not match JS.  What do you think?

rchavarria:
I would just export the class. It doesn't matter too much if this track differs a little bit from xjavascript. 

Then, there is no a big change in the tests. Just add a new line saying `const say = new Say()` and all `it` blocks remain the same.

matthewmorgan:
OK @rchavarria I changed the export structure.

rchavarria:
Much nicer. Thank you so much!!

Issue #154, opened 2015-12-10
Add new exercise custom-set


matthewmorgan:
I will port this from the JS track

rchavarria:
Closed by #155

PR #155, opened 2015-12-10
Add new exercise custom-set

matthewmorgan:
As described in #154

### DIFF
--- a/custom-set/custom-set.spec.js
+++ b/custom-set/custom-set.spec.js
@@ -1,0 +1,117 @@
+import CustomSet from './custom-set';
+
+describe('CustomSet', () => {
+
+  it('can delete elements', () =>{
+    const expected = new CustomSet([1, 3]);
+    const actual = new CustomSet([3, 2, 1]).delete(2);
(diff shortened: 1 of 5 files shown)


custom-set/custom-set.spec.js
```diff
@@ -0,0 +1,117 @@
+import CustomSet from './custom-set';
+
+describe('CustomSet', () => {
+
+  xit('can delete elements', () =>{
```
rchavarria: Whoops, we should enable at least the first test, right?

rchavarria:
It looks like `example.js` contains only ES5 code, without any ES2015 feature. Are you ok with it? Wouldn't it be better to refactor it to some kind of ES2015 code?

matthewmorgan:
I'm sorry-- I must have committed the wrong file.  I'll resubmit, and also enable the first test.

matthewmorgan:
@rchavarria all set on the changes.  Please have a look.

rchavarria:
Awesome! Another new exercise :tada:

Issue #156, opened 2015-12-10
Add new exercise sum of multiples


matthewmorgan:
I will port this from the JS track

rchavarria:
Closed by #157

PR #157, opened 2015-12-10
Add new exercise sum-of-multiples

matthewmorgan:
As described in #156

### DIFF
--- a/sum-of-multiples/example.js
+++ b/sum-of-multiples/example.js
@@ -1,0 +1,18 @@
+const Sum = (factors = [3, 5]) => {
+  let self = this instanceof Sum ? this : Object.getPrototypeOf(Sum);
+  self.factors = factors;
+
+  self.to = limit => {
+    let multiples = {};
+    self.factors.forEach(factor => {
+      for (let ii = factor; ii < limit; ii += factor) {
+        multiples[ii] = ii;
+      }
+    });
+
+    return Object.keys(multiples).reduce((prev, curr) => prev += multiples[curr], 0);
+  };
+  return self;
+};
+
+export default Sum;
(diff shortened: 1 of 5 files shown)


sum-of-multiples/example.js
```diff
@@ -0,0 +1,18 @@
+const Sum = (factors = [3, 5]) => {
+  let self = this instanceof Sum ? this : Object.getPrototypeOf(Sum);
```
rchavarria: It's a strange statement. I haven't seem anything like this before. I'm ok with this, but just for the purpose of learning. I would have written the `example.js` in a different way. But I think it's ok to be merged.

matthewmorgan: I agree it's a little weird.  

Like with #153 I would rather export a class, but that's not what the test wants.  How would you have written `example.js`?

rchavarria: I think that this line is to use `Sum` as a constructor function (`new Sum()`) or just a function (`Sum()`).

I would write it as a class or just assuming it will be used as a constructor function:

```
class Sum {
  constructor(factors) {}
  to(limit) {}
}
```

or:

```
function Sum(factors) {
  this.factors = factors;
}
Sum.prototype.to = limit => {}
```

matthewmorgan:
Yes, I agree with that as the right approach.  With the current structure of the test, neither of those very sensible versions will work-- at least I couldn't get them to work!

If you have any interest in pursuing it, I would be curious to see if you can get either of those two approaches to work without changing the test structure.  I'm clearly missing something.

rchavarria:
There might be no way to change `example.js` without changing the tests. Do you see any problem in changing the tests? Instead of doing `SumOfMultiples().to(4)`, they should do `new SumOfMultiples().to(4)`. 

They're quite similar. I would prefer the second one, for consistency with other exercises. What do you think? Would it be interesting to have different approaches?

Issue #158, opened 2015-12-10
Add new exercise queen-attack


matthewmorgan:
I will port this from the JS track

rchavarria:
Closed by #159

PR #159, opened 2015-12-10
Add new exercise queen-attack

matthewmorgan:
As described in #158

### DIFF
--- a/queen-attack/example.js
+++ b/queen-attack/example.js
@@ -1,0 +1,47 @@
+const W = 8,
+  H = 8,
+  STARTING = { black : [7,3], white: [0,3] };
+
+const QueenAttack = ( params = STARTING )  => {
+  const self = this instanceof QueenAttack ? this : Object.getPrototypeOf(QueenAttack);
+  if (params && params.white === params.black){
+    throw new Error("Queens cannot share the same space");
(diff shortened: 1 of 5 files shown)


queen-attack/example.js
```diff
@@ -0,0 +1,47 @@
+const W = 8,
+  H = 8,
+  STARTING = { black : [7,3], white: [0,3] };
+
+const QueenAttack = ( params = STARTING )  => {
+  const self = this instanceof QueenAttack ? this : Object.getPrototypeOf(QueenAttack);
```
rchavarria: This is the same strange approach as #157, but ok.

queen-attack/example.js
```diff
@@ -0,0 +1,47 @@
+const W = 8,
+  H = 8,
+  STARTING = { black : [7,3], white: [0,3] };
+
+const QueenAttack = ( params = STARTING )  => {
```
rchavarria: There is a very good opportunity here to use *default parameters* and *deconstruction*:

```
const QueenAttack = ({ black = [7, 3], white = [0, 3] }) => {
// ...
```

Just in case you want to create another PR :stuck_out_tongue_winking_eye:

Issue #162, opened 2015-12-11
Add new exercise ocr-numbers


matthewmorgan:
I will port this from the JS track

rchavarria:
Closed by #163

PR #163, opened 2015-12-11
Add new exercise ocr-numbers

matthewmorgan:
As described in #162


rchavarria:
Thanks!! This exercise seems really fun!

Issue #164, opened 2015-12-12
Add new exercise meetup


matthewmorgan:
I will port this from the JS track

rchavarria:
Closed by #165

PR #165, opened 2015-12-12
Add new exercise meetup

matthewmorgan:
As described in #164

### DIFF
--- a/config.json
+++ b/config.json
@@ -57,6 +57,7 @@
     "sum-of-multiples",
     "queen-attack",
     "binary-search",
+    "meetup",
     "binary-search-tree",
     "matrix",
     "pascals-triangle",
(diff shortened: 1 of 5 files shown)


rchavarria:
Great!! Nice job.

Just in case you want, I would change the description of tests. Instead of use an underscore to separate words, I would write descriptions as regular sentences. They are strings, so there is no need to write them as Ruby methods.

Also, in `example.js`, I'm not a big fun of starting *private* methods with underscore. In ES2015, using modules, they are already private.

matthewmorgan:
I didn't even *notice* the underscores in the test descriptions!  

I completely agree about the underscores in the code, I am not a fan in general.  I think it makes the code hard to read, and the objects are not really private, they just 'look' private.  I think it is, as we say, putting lipstick on a pig.

I ported this exercise as is, with minimal changes, but it might be worth revisiting at some point.

rchavarria:
Yeah. You can open a issue if you want so that we won't forget.

Issue #166, opened 2015-12-12
Add new exercise bracket-push


matthewmorgan:
I will port this from the JS track

rchavarria:
Closed by #167

PR #167, opened 2015-12-12
Add new exercise bracket-push

matthewmorgan:
As described in #166


rchavarria:
Great! That's all! Well, there is still the last exercise :smiley: 

Thank you very much, I'll go for it.

Issue #172, opened 2015-12-15
Beer song: add a test to check the whole song


rchavarria:
Add a test that checks the whole song is returned:

```
99 bottles of beer on the wall, 99 bottles of beer.
Take one down and pass it around, 98 bottles of beer on the wall.
...
No more bottles of beer on the wall, no more bottles of beer.
Go to the store and buy some more, 99 bottles of beer on the wall.
```

matthewmorgan:
Is @StevenACoffman going to put in a PR for this, or did you intend to do it?

rchavarria:
I'd let @StevenACoffman decide if he wants to work on this or not. 

For the moment, I'm not planning to work on this. If @StevenACoffman isn't working on this, you can do it.

StevenACoffman:
I'm not going to be able to get to this for a few days, so by all means
feel free. Thanks.
-Steve

matthewmorgan:
Ok if find some time I will assign this to myself.  Otherwise it's all yours Steve.

matthewmorgan:
Closed by #173

PR #173, opened 2015-12-16
Add test for whole song

matthewmorgan:
As described in #172

### DIFF
--- a/beer-song/beer-song.spec.js
+++ b/beer-song/beer-song.spec.js
@@ -53,4 +53,28 @@ Go to the store and buy some more, 99 bottles of beer on the wall.
     expect(Beer.sing(3)).toEqual(expected);
   });
 
+  xit('sings all the verses', () => {
+    const finish = `2 bottles of beer on the wall, 2 bottles of beer.
+Take one down and pass it around, 1 bottle of beer on the wall.
+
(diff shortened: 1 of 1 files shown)


rchavarria:
Thanks!! In the end, you solved the exercise in just one test :wink: 

I had in mind something different, such as checking just if strings like these were present:

- `99 bottles of beer on the wall`
- `98 bottles of beer on the wall`
- ...
- `2 bottles of beer on the wall`

It's not 100% accurate, users could cheat that test, but the main goal of this new test was to test that the whole song starts in 99 bottles.

What do you think?

matthewmorgan:
Ha-- I guess the code for that test could be a legit solution on its own, eh?  A little brut-force, but still...

I'm not quite sure that I do understand what you mean: do you mean just to test the beginning of their output for the presence of `99 bottles of beer on the wall` ?

rchavarria:
Actually, I was thinking on having several `expect`s, or expect them inside a loop.:

```
const song = Berr.sing();
expect(song).toMatch('99 bottles of beer');
expect(song).toMatch('99 bottles of beer');
// ...
expect(song).toMatch('2 bottles of beer');
```

Search `toMatch` on [Jasmine docs](http://jasmine.github.io/2.0/introduction.html).

PR #174, opened 2015-12-18
Change test to match all expected verses

matthewmorgan:
@rchavarria perhaps this is what you were looking for in #172


rchavarria:
Well, not exactly, but it's ok. In the end, the test checks what we want to check, that the song starts at 99 bottles and counts down to no bottles.

matthewmorgan:
Well, I would be curious to see your implementation of what you had in mind.  If you have time, that is

rchavarria:
Yep. I'll give it a try. Be patient :wink:

PR #178, opened 2015-12-21
Make it simple how to test the whole song

rchavarria:
The test is not as complete as the existing one. It doesn't test the complete verses. But it relies on previous tests.

It just tests two things:

1. There are almost 100 verses starting with `XX bottles of beer on the wall`
2. The song consist of `3 * 100` lines, 3 lines each verse, 100 verses

@matthewmorgan Feel free to discuss this implementation. It's simpler than the existing one, but you may prefer a more complete version than mine. That's ok.

### DIFF
--- a/beer-song/beer-song.spec.js
+++ b/beer-song/beer-song.spec.js
@@ -54,22 +54,299 @@ Go to the store and buy some more, 99 bottles of beer on the wall.
   });
 
   xit('sings all the verses', () => {
-
     const song = Beer.sing();
+    expect(song).toEqual(`99 bottles of beer on the wall, 99 bottles of beer.
+Take one down and pass it around, 98 bottles of beer on the wall.
 
-    //are the first 97 verses present?
-    for (let verseNum = 3; verseNum < 99; verseNum++) {
+98 bottles of beer on the wall, 98 bottles of beer.
+Take one down and pass it around, 97 bottles of beer on the wall.
 
-      const verse = `${verseNum} bottles of beer on the wall, ${verseNum} bottles of beer.
-Take one down and pass it around, ${verseNum-1} bottles of beer on the wall.
+97 bottles of beer on the wall, 97 bottles of beer.
+Take one down and pass it around, 96 bottles of beer on the wall.
 
-`;
+96 bottles of beer on the wall, 96 bottles of beer.
+Take one down and pass it around, 95 bottles of beer on the wall.
 
-      expect(song).toMatch(verse);
-    }
+95 bottles of beer on the wall, 95 bottles of beer.
+Take one down and pass it around, 94 bottles of beer on the wall.
 
-    //how about the final 3?
-    const expected = `2 bottles of beer on the wall, 2 bottles of beer.
+94 bottles of beer on the wall, 94 bottles of beer.
+Take one down and pass it around, 93 bottles of beer on the wall.
+
+93 bottles of beer on the wall, 93 bottles of beer.
+Take one down and pass it around, 92 bottles of beer on the wall.
+
+92 bottles of beer on the wall, 92 bottles of beer.
+Take one down and pass it around, 91 bottles of beer on the wall.
+
+91 bottles of beer on the wall, 91 bottles of beer.
+Take one down and pass it around, 90 bottles of beer on the wall.
+
+90 bottles of beer on the wall, 90 bottles of beer.
+Take one down and pass it around, 89 bottles of beer on the wall.
+
+89 bottles of beer on the wall, 89 bottles of beer.
+Take one down and pass it around, 88 bottles of beer on the wall.
+
+88 bottles of beer on the wall, 88 bottles of beer.
+Take one down and pass it around, 87 bottles of beer on the wall.
+
+87 bottles of beer on the wall, 87 bottles of beer.
+Take one down and pass it around, 86 bottles of beer on the wall.
+
+86 bottles of beer on the wall, 86 bottles of beer.
+Take one down and pass it around, 85 bottles of beer on the wall.
+
+85 bottles of beer on the wall, 85 bottles of beer.
+Take one down and pass it around, 84 bottles of beer on the wall.
+
+84 bottles of beer on the wall, 84 bottles of beer.
+Take one down and pass it around, 83 bottles of beer on the wall.
+
+83 bottles of beer on the wall, 83 bottles of beer.
+Take one down and pass it around, 82 bottles of beer on the wall.
+
+82 bottles of beer on the wall, 82 bottles of beer.
+Take one down and pass it around, 81 bottles of beer on the wall.
+
+81 bottles of beer on the wall, 81 bottles of beer.
+Take one down and pass it around, 80 bottles of beer on the wall.
+
+80 bottles of beer on the wall, 80 bottles of beer.
+Take one down and pass it around, 79 bottles of beer on the wall.
+
+79 bottles of beer on the wall, 79 bottles of beer.
+Take one down and pass it around, 78 bottles of beer on the wall.
+
+78 bottles of beer on the wall, 78 bottles of beer.
+Take one down and pass it around, 77 bottles of beer on the wall.
+
+77 bottles of beer on the wall, 77 bottles of beer.
+Take one down and pass it around, 76 bottles of beer on the wall.
+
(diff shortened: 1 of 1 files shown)


matthewmorgan:
I see what you were thinking.  

I guess since the other tests cover the last verse, this would be OK.  But why replace a test that tests all the verses?

rchavarria:
My idea for issue #172 was to have a new test as simple as possible. Solution on #173 implements what we need in the issue, but I would like a simpler one.

It doesn't make sense to have #173 and this PR in production, and it's not a big problem for me. We can *pause* this PR and wait to know what users think about the test, or we could do an A/B test (we publish both tests and ask users what they think).

matthewmorgan:
I'm fine with whatever you want to do.  

I did have another thought: if we wanted to make sure the whole song output was an exact match, we could hash it, right?  It would mean hardcoding the md5 hash in the test and including a script to do the hashing.  The we could test the entire output in one function call.

rchavarria:
That's a good idea. A tiny cryptic one, but a good idea. The test will be very concise, what's nice.

I only have a problem, I don't know if a MD5 can be computed with plain JavaScript or an external library is needed.

matthewmorgan:
I think we should use a library.  We will need to add it to the makefile, like for the Grains exercise.  CryptoJS would be one library that we could use.

Do you want to implement this or shall I?

rchavarria:
I don't know about CryptoJS. I would be glad to see how to use it. Please, show me the code :wink:

matthewmorgan:
I thought of a different way.  We could modify the makefile to include 
```
sed -i '' 's/\"dependencies\": {}/\"dependencies\":{\"md5\": \"^2.0.0\"}/g' beer-song/package.json
```
to import the `md5` module locally to the `beer-song` exercise.  Then, it becomes simple to add a test using that module:

```
  it('checks the test of the whole song using an md5 hash', () => {
    const song = Beer.sing();
    expect(md5(song)).toEqual('42571321165c54931597cdf8fdd6c4c3');
  })
```

Any reason why we shouldn't do that?  Do you think it would work in Travis?

rchavarria:
It could work on Travis, but exercism users would also need that dependency. We have two options:

1. Include that dependency on every `package.json` file (we're not able to customize that file for each exercise yet -- see XXX)
2. Include the CryptoJS code in the exercise folder. It will be deployed to users machines, and tests would include it as a module (`import CryptoJS from './crypto.js'`). 

If CryptoJS has more dependencies, second option could be more and more complicated. I would go for the first one for now.

matthewmorgan:
CryptoJS has no other dependencies, but on the other hand `md5` is very quick to install.  It's not going to slow things down noticeably for the users.

rchavarria:
Ok. So, let's modify the main `package.json` (it will be spread over each individual exercise's `package.json`).

Let's see how it works!

matthewmorgan:
Do you want to change the PR on this one?  Or shall I?

rchavarria:
Please, go for it!

matthewmorgan:
OK, I'm having an issue trying to add the `md5` module.  It works from within the exercise directory with `gulp test` but not from the repo root with `make` or `make test-assignment ASSIGNMENT=beer-song`.

It's complaining that it can't find the `md5` module, but it is in `node_modules`, so what am I missing?

matthewmorgan:
@kytrinyx any pointers on adding a module?  I must be missing something simple.

rchavarria:
I think I understand your problem. It's as if `make` and running `gulp test` on individual exercises use different NodeJS modules.

If you added `md5` to the main `package.json` just execute `npm install` in the root directory and it will install missing modules. Then, `make` command should work.

matthewmorgan:
No, that is what I am saying-- I installed the module at the root level but I am still getting an error saying it can't be found.  The module is in the main `package.json`, and in the main `node_modules` directory. It *is* installed, yet when running `make` I get this error:

```
Running tests for beer-song assignment
[06:57:21] Using gulpfile ~/xecmascript/gulpfile.js
[06:57:21] Starting 'lint'...
[06:57:21] Starting 'babel'...
[06:57:22] Finished 'lint' after 228 ms
[06:57:22] Finished 'babel' after 214 ms
[06:57:22] Starting 'test'...
[06:57:22] 'test' errored after 19 ms
[06:57:22] Error in plugin 'gulp-jasmine'
Message:
    Cannot find module 'md5'
Details:
    code: MODULE_NOT_FOUND
make[1]: *** [test-assignment] Error 1
make: *** [test] Error 1
```

rchavarria:
I think I had a similar problem when working on Grains exercise (it uses an external library, big-integer).

It seems that our `make` command don't let individual exercises to use NodeJS modules. This command creates a temporal folder, and moves files to that temporal folder. Maybe, we should modify this *build* process to run `npm install` on those temporal folders to get installed all dependencies.

I created this issue to discuss and work on this: see #182 (feel free to improve the description)

matthewmorgan:
Thanks.  I've been doing some research and it could be the problem is
related to using a task runner.  I'm looking at it right now.

On Wed, Dec 30, 2015 at 7:39 AM, Rubén Chavarría <notifications@github.com>
wrote:

> I think I had a similar problem when working on Grains exercise (it uses
> an external library, big-integer).
>
> It seems that our make command don't let individual exercises to use
> NodeJS modules. This command creates a temporal folder, and moves files to
> that temporal folder. Maybe, we should modify this *build* process to run npm
> install on those temporal folders to get installed all dependencies.
>
> I created this issue to discuss and work on this: see #182
> <https://github.com/exercism/xecmascript/issues/182> (feel free to
> improve the description)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/exercism/xecmascript/pull/178#issuecomment-167990533>.
>



-- 
Matt Morgan
Director of Technology
Dulcet Software LLC
(734) 926-8142

matthewmorgan:
Well, still not sure what's happening.  I added my `sed` command from above, and the `package.json` for `beer-song` includes `md5`.  I've added `md5` to the root `package.json` file, re-run `npm install`, still same errors.

The `md5` module is in `node_modules`, though I did notice it is not in `node_modules/bin`.  I don't know if that means anything.

rchavarria:
I don't know. Maybe some NodeJS modules are installed as binaries, and an executable is placed under `node_modules/bin`.

I think the problem is that exercises are *moved* to a temporal folder. In that folder there is no `node_modules` folder, so exercises can not import node_modules as if a user was solving the exercise.

matthewmorgan:
Well for now maybe we should use Crypto JS. It could be copied via the
makefile like bigint. I could work on that tonight

kytrinyx:
@matthewmorgan I have no idea how to add node modules.

For what it's worth, I think that tests should have as little logic as possible. For this exercise my preferred test is a big-ass string with 300 lines in it containing the actual lyrics to the song.

matthewmorgan:
@kytrinyx interesting.  Things should be a simple as possible but no simpler.  I appreciate that.

matthewmorgan:
@rchavarria looking at the ruby version of the test, it looks like it just tests the compatibility between the `verses` method and the `sing` method:
```
  def test_the_whole_song
    skip
    assert_equal song.verses(99, 0), song.sing
  end
```

kytrinyx:
Weird. I thought I fixed that waaaay back. I'll go fix it to just have the lyrics (I'm writing a book about this problem http://signup.99bottlesbook.com/ in Ruby, and it's got the 300 line test there).

matthewmorgan:
OK @rchavarria I took @kytrinyx 's suggestion on this one.  I did of course spend a bunch of time trying to figure out how to add a module to the test environment, but no success!

kytrinyx:
I updated the Ruby one here: https://github.com/exercism/xruby/pull/236

rchavarria:
Ok, no problem for me. If adding that new library is causing more problems than solutions, let's go for the whole song. Wasn't it your first approach, @matthewmorgan?

matthewmorgan:
@rchavarria yes, though I was building it dynamically, using a loop for most of the verses.  That would violate the principle of 'as little logic as possible,' for sure.

Shall I merge this change then?

rchavarria:
I think so, go for it!

matthewmorgan:
done.  Thanks @rchavarria @kytrinyx for hashing this one out!

Issue #177, opened 2015-12-21
Robot name: add test to check names are generated randomly


rchavarria:
`README` file says that the solution should generate robot names randomly, but it's not forced by any test. Add such a test that check, at least, that robot names are not generated sequentially.

matthewmorgan:
Just wondering how we can test this?  I think it may not be possible. 

It's entirely possible (though very unlikely) that a set of randomly generated names could in fact be sequential.  How can we tell the difference?

imagine putting 10 names in a hat and drawing them 'randomly'.  It is certainly conceivable that the names could be drawn in alphabetical order.

rchavarria:
Absolutely. I think that discussion already happened on xjavascript track (well kind of [xjavascript#105](https://github.com/exercism/xjavascript/issues/105))

Let's assume names are generated randomly. Testing two robot names has a certain probability to be sequential: `AA123` and `AA124`. But, testing three names, the probability will be much lower: `AA123`, `AA124` and `AA125` is much more difficult to get than the previous one.

I'm not sure how to mathematically calculate each probability. 

And it might not be so simple to test that two names are sequential. It's easy with `AA123` and `AA124`, but it's not so easy with `AA999` and `AB000`. It will make the test even more difficult than the solution itself.

matthewmorgan:
Yeah, I guess it's just a little weird-- I could have a functioning solution that generates a random sequence, and it could fail bc my solution generates two (or three) names that just happen to be sequential.

BTW I think the chance of three successive sequential names is (1/675999)*(1/675998), or 2.19 x 10^-12.  About 2 in a trillion.  Really small.

rchavarria:
Agree. The probability to have 3 succesive sequential names is really small. It could work.

But, the problem is how to check that two given names are sequential. I can only think of a solution that almost solves the exercise itself, and that should not be the goal of tests.

matthewmorgan:
How about something like this?  Obviously the test isn't set up to test three names in this example, but this is the principle:

```
  it('should test for sequential names', () => {
    expect(areSequential("AA000", "AA001")).toEqual(true);
    expect(areSequential("AA000", "AA002")).toEqual(false);
    expect(areSequential("AA999", "AB000")).toEqual(true);
    expect(areSequential("AA999", "AA998")).toEqual(true);
    expect(areSequential("AA999", "AA997")).toEqual(false);
    expect(areSequential("AB000", "AA999")).toEqual(true);
[...]

PR #180, opened 2015-12-24
Add new test to catch sequential names

matthewmorgan:
As discussed in #177

### DIFF
--- a/robot-name/robot-name.spec.js
+++ b/robot-name/robot-name.spec.js
@@ -60,4 +60,40 @@ describe('Robot', () => {
     expect(modifyInternal).toThrow();
   });
 
+
+  xit('new names should not be sequential', () => {
+    const name1 = robot.name;
+    const name2 = (new Robot()).name;
+    const name3 = (new Robot()).name;
+    expect(areSequential(name1, name1)).toBe(true);
+    expect(areSequential(name1, name2)).toBe(false);
+    expect(areSequential(name2, name3)).toBe(false);
+  });
+
+  xit('names from reset should not be sequential', () => {
+    const name1 = robot.name;
+    robot.reset();
+    const name2 = robot.name;
+    robot.reset();
+    const name3 = robot.name;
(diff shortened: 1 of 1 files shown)


robot-name/robot-name.spec.js
```diff
@@ -60,4 +60,33 @@ describe('Robot', () => {
...
 
+
+  xit('should test for sequential names', () => {
+    const name1 = robot.name;
+    robot.reset();
+    const name2 = robot.name;
+    robot.reset();
+    const name3 = robot.name;
+
+    expect(areSequential(name1, name2)).not.toBe(true);
```
rchavarria: You can expect falsy values in different ways in Jasmine.

```
expect(areSequential(name1, name2)).not.toBe(true);
expect(areSequential(name1, name2)).toBe(false);
expect(areSequential(name1, name2)).toBeFalsy();
```

I'm not very fan of the first one. What do you think? What do you think is the most readable?

matthewmorgan:
I had to look up the syntax on `not.toBe()` myself.  I have seen it used and wanted to try it!

But I don't really feel strongly about it-- option 2 would be fine with me otherwise.

rchavarria:
Ok, let's use second option.

Another question to generate names. What about generate them this way?

```
const name1 = robot.name; // the first robot instance is already generated
const name2 = new Robot().name;
const name3 = new Robot().name;
```

matthewmorgan:
I propose we test both `Robot.reset().name` and `new Robot().name`.  In the event someone were using different techniques to generate vs resetting, we would want to test both for non-sequential output, right?

rchavarria:
I see your point, some users can create and reset names differently.

So, let's create then two tests: one to check new names, one to check reset names. Both tests should look for non-sequential names. Do you agree?

matthewmorgan:
done!

rchavarria:
There is a new commit because I rebased it onto master, but I changed no code.

Thanks a lot for this code. Those new tests look awesome!!

Issue #183, opened 2016-01-16
Add new exercise acronym


matthewmorgan:
I will add this modeled on the Ruby exercise

https://github.com/exercism/xruby/tree/master/acronym

rchavarria:
Closed by #184

PR #184, opened 2016-01-16
Add new exercise acronym

matthewmorgan:
As described in #183

### DIFF
--- a/acronym/acronym.spec.js
+++ b/acronym/acronym.spec.js
@@ -1,0 +1,19 @@
+import Acronyms from './acronym';
+
+describe('acronyms', ()=>{
+  const PAIRS = {
+    'Portable Network Graphics' : 'PNG',
+    'Ruby on Rails' : 'ROR',
+    'HyperText Markup Language': 'HTML',
+    'First In, First Out' : 'FIFO',
+    'PHP: Hypertext Preprocessor' : 'PHP',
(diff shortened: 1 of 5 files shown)


rchavarria:
Thanks!! 

Long time without a new exercise.

The test suite consist only of one test. Do you like it this way? I usually prefer having several *specs*, kind of:

```
it('takes the first letter of each word', () => {
  expect(Acronyms.parse('Portable Network Graphic').toEqual('PNG');
});
it('returns all acronyms uppercase', () => {
  expect(Acronyms.parse('Ruby on Rails').toEqual('ROR');
});
// ...
```

I know there is some duplicated code, but it lets the user implement the solution bit by bit. What do you think?

matthewmorgan:
I agree (and I knew you would say that!)  I simply duplicated the structure of the Ruby test suite.  But I think successive tests would be more interesting for the solver.

I thought also about creating some more tests (EG empty or invalid inputs) but there is a json file for this exercise that specifies the tests to run, so I think I shouldn't add new test without putting in a PR to x-common.  I also don't know if this creates a problem for any existing test suites in other tracks-- @kytrinyx would we create problems by adding tests to the metadata in x-common?

I think it would be better to break out the tests, and I'd be happy to do that.  If it makes sense to add new ones I can do that as well.

Issue #170, opened 2015-12-14
Clean up underscores from meetup exercise tests and code


matthewmorgan:
Underscores serve no real purpose in the `example.js` for this exercise, and are not needed in the test descriptions.  Let's remove them.

rchavarria:
Closed by #185

PR #185, opened 2016-01-17
Remove underscores from meetup test and example code

matthewmorgan:
As described in #170


rchavarria:
It can be read much better now, thank you so much!!

PR #195, opened 2016-02-16
Acronym: disable all tests but not the first one

rchavarria:
Disable all tests except the first one to help users to implement the solution bit by bit.


matthewmorgan:
Thanks @rchavarria !  (for fixing what must have been my mistake 😳)

rchavarria:
Don't blame yourself. It could be just a small distraction. That's why it's better to have several people do the programming :wink:

matthewmorgan:
You're too kind.

Issue #196, opened 2016-02-16
Make a distinction between JavaScript and ECMAScript tracks


rchavarria:
As discussed in [exercism.io#2750](https://github.com/exercism/exercism.io/issues/2750).

PR #197, opened 2016-02-23
Write a message trying to make it clear the intention of this track

rchavarria:
This PR adds the notice message discussed in #196.


matthewmorgan:
@rchavarria Thanks again for doing this!  Seems like the track is gaining momentum.

rchavarria:
Do you think those little messages are working? That would be awesome!

PR #201, opened 2016-03-03
Add tests for emptying list, then adding elements

jcshih:
Add tests cases to ensure the linked list's state is properly reset when popping/shifting its last element.

### DIFF
--- a/exercises/linked-list/example.js
+++ b/exercises/linked-list/example.js
@@ -13,16 +13,23 @@ export default class Deque {
     if (!this.head) {
       return undefined;
     }
-    const result = this.head.value;
-    this.head = this.head.prev;
-    return result;
+
+    const value = this.head.value;
+    if (this.head.next) {
+      this.head = this.head.next;
+      this.head.prev = null;
+    } else {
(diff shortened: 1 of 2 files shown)


matthewmorgan:
@jcshih -- thanks for catching this (especially since that was my `example.js` ;) !!)

I think it could be written so the logic is a little more clear.  In hindsight I think the ternaries were not the best choice. What do you think about this, for example:
```
  shift() {
    if (!this.tail) {
      return undefined;
    }
    const value = this.tail.value;
    if (this.tail.next){
      this.tail = this.tail.next;
      this.tail.next = null
    } else {
      this.tail = null;
      this.head = null;
    }
    return value;
  }
```

Also, we usually like to see one commit per PR if possible.  If you need help squashing the commits please let me know, I'd be happy to assist.

jcshih:
@matthewmorgan Thanks for the tips. I agree that conditionals are more readable than ternaries in this case.

I went ahead and refactored and squashed all the changes. I also switched next/prev because it made more sense to me.

matthewmorgan:
Thanks @jcshih !

PR #210, opened 2016-04-04
Update tests and example to be case-agnostic

matthewmorgan:
The test metadata for this exercise specifies that counts should be independent of the case of the input words.

This PR brings this exercise in to conformance with that requirement.

### DIFF
--- a/exercises/word-count/word-count.spec.js
+++ b/exercises/word-count/word-count.spec.js
@@ -28,7 +28,7 @@ describe('words()', () => {
     expect(words.count('testing 1 2 testing')).toEqual(expectedCounts);
   });
 
-  xit('respects case', () => {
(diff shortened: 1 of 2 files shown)


exercises/word-count/word-count.spec.js
```diff
@@ -29,12 +29,12 @@ describe('words()', () => {
   });
 
   xit('respects case', () => {
```
rchavarria: Same as xjavascript :smile:
I think this description does not reflect what the spec does now. What about normalizes to lower case? @matthewmorgan, if you agree, I can do the change while merging the PR.

matthewmorgan: Absolutely! Thanks for catching that and please change it if it's not too
much trouble.

Issue #204, opened 2016-03-13
Verify "Largest Series Product" exercise implementation


kytrinyx:
There was some confusion in this exercise due to the ambiguous use of the term _consecutive_ in the README. This could be taken to mean _contiguous_, as in consecutive by position, or as in _consecutive numerically_. The the README has been fixed (https://github.com/exercism/x-common/pull/200).

**Please verify that the exercise is implemented in this track correctly** (that it finds series of contiguous numbers, not series of numbers that follow each other consecutively).

If it helps, the canonical inputs/outputs for the exercise can be found here:
https://github.com/exercism/x-common/blob/master/largest-series-product.json

If everything is fine, go ahead and just close this issue. If there's something to be done, then please describe the steps needed in order to close the issue.

rchavarria:
None of the tests included in this track violates rules stated in the `README`.

But, there are interesting cases in the canonical inputs/outputs that could be added:

```
[
// ...
    {
        "description": "can get the largest product of a big number (Project Euler)",
        "digits": "7316717653133062491922511967442657474235534919493496983520312774506326239578318016984801869478851843858615607891129494954595017379583319528532088055111254069874715852386305071569329096329522744304355766896648950445244523161731856403098711121722383113622298934233803081353362766142828064444866452387493035890729629049156044077239071381051585930796086670172427121883998797908792274921901699720888093776657273330010533678812202354218097512545405947522435258490771167055601360483958644670632441572215539753697817977846174064955149290862569321978468622482839722413756570560574902614079729686524145351004748216637048440319989000889524345065854122758866688116427171479924442928230863465674813919123162824586178664583591245665294765456828489128831426076900422421902267105562632111110937054421750694165896040807198403850962455444362981230987879927244284909188845801561660979191338754992005240636899125607176060588611646710940507754100225698315520005593572972571636269561882670428252483600823257530420752963450",
        "span": 13,
        "expected": 23514624000
    },
    {
        "description": "reports zero if all spans include zero",
        "digits": "99099",
        "span": 3,
        "expected": 0
    },
    {
        "description": "rejects invalid character in digits",
        "digits": "1234a5",
        "span": 2,
        "expected": -1
    },
    {
        "description": "rejects negative span",
        "digits": "12345",
        "span": -1,
        "expected": -1
    }
]
```
And, one case that explains the weird (I found it weird the first time I read the specs) spec that returns `1` (the spec already exists, but not the explanation):

```
{
    "#": [
[...]

PR #211, opened 2016-04-05
Add new cases from metadata

matthewmorgan:
Adds some test cases form the canonical `json` file as discussed in #204


exercises/largest-series-product/largest-series-product.spec.js
```diff
@@ -27,6 +35,18 @@ describe('Series', () => {
     expect(new Series('99099').largestProduct(3)).toBe(0);
   });
 
+  xit('rejects invalid character in input', ()=> {
+    expect(()=> {
+      new Series('1234a5').largestProduct('2')
+    }).toThrow(new Error('Invalid input.'));
```
rchavarria: The canonical specs talk about returning `-1` instead of throwing error. But, other existing specs were throwing, so I'd keep these throws for now.

exercises/largest-series-product/largest-series-product.spec.js
```diff
@@ -27,6 +35,18 @@ describe('Series', () => {
...
+  });
+
+  xit('rejects negative span', ()=> {
+    expect(() => {
+      new Series('12345').largestProduct(-1)
+    }).toThrow(new Error('Invalid input.'));
```
rchavarria: Same as line 41

rchavarria:
A-we-some!

Issue #213, opened 2016-04-07
binary: improve tests for invalid numbers


kytrinyx:
We should have separate tests for:

- alphabetic characters at the beginning of a valid binary number
- alphabetic characters at the end of a valid binary number
- alphabetic characters in the middle of an otherwise valid binary number
- invalid digits (e.g. 2)

If the test suite for binary has test cases that cover these edge cases, this issue can safely be closed.

See https://github.com/exercism/x-common/issues/95

matthewmorgan:
Of the items listed above, this track lacks a test for:

invalid characters in the middle of an otherwise valid binary number.

matthewmorgan:
Closed by #217

PR #217, opened 2016-04-07
Binary: add missing test

rchavarria:
Addresses #213 and adds the missing test: alphabetic characters in the number makes it to be decimal 0.


matthewmorgan:
Thanks again!  It's nice to keep this track maintained well.  Cheers :beers:

PR #218, opened 2016-04-11
Clock: synchronize tests with canonical ones

rchavarria:
Addresses #216 

There are lots of differences. Even the track solution needed some changes. But, it synchronizes tests with canonical ones: [`clock.json`](https://github.com/exercism/x-common/blob/master/clock.json).

xjavascript track should be very similar to this one.

### DIFF
--- a/exercises/clock/clock.spec.js
+++ b/exercises/clock/clock.spec.js
@@ -2,59 +2,216 @@ import at from './clock';
 
 describe('Clock', () => {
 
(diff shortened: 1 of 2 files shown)


matthewmorgan:
This was a big PR.  Thanks for doing the work!

I did find one typo which I fixed ('substracts -> subtracts').  Otherwise it looked great. Thanks @rchavarria !

PR #221, opened 2016-04-30
Assertion not executed in Circular Buffer tests

ryanplusplus:


exercises/circular-buffer/circular-buffer.spec.js
```diff
@@ -31,7 +31,7 @@ describe('CircularBuffer', () => {
     buffer.write('1');
     buffer.write('2');
     buffer.clear();
-    expect(buffer.read).toThrowError;
```
ryanplusplus: This was the wrong assertion, but it wasn't noticed because it wasn't actually executed.

rchavarria:
Nice catch! Many thanks!

PR #223, opened 2016-05-06
Makefile: Replace xdescribe with describe in addition to replacing xit with it

ryanplusplus:
I noticed that any spec with an `xdescribe` was not tested correctly with `make test`. This ensures that these `describe`s are enabled along with all `it`s.


rchavarria:
Very well caught @ryanplusplus!

Just for reference, it was the `leap` exercise who has the `xdescribe` thing. Thanks a lot for this PR.

PR #225, opened 2016-05-11
(fix for #224) Fix/enhance word-count test.

### LINKED ISSUE #224: word-count: 'prototype' is not a property on the Object prototype
Relevant test: 'handles properties that exist on Object’s prototype'

As 'prototype' isn't a standard property on the Object prototype,
the test doesn't actually check the case of dealing with a property name collision.

The 'constructor' property is a standard, however:

es5: http://www.ecma-international.org/ecma-262/5.1/#sec-15.2.4
es6: http://www.ecma-international.org/ecma-262/6.0/#sec-properties-of-the-object-prototype-object

There has already been a PR merged for this on the JavaScript track:

https://github.com/exercism/xjavascript/pull/199

ZacharyRSmith:
fixes #224

### DIFF
--- a/exercises/word-count/word-count.spec.js
+++ b/exercises/word-count/word-count.spec.js
@@ -59,8 +59,8 @@ describe('words()', () => {
   });
 
   xit('handles properties that exist on Object’s prototype', () => {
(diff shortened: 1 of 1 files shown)


matthewmorgan:
Thanks @ZacharyRSmith !  Sorry it took so long to get this merged, but your contribution is appreciated.

PR #230, opened 2016-06-05
Accumulate spec: add missing semicolons

ryanplusplus:


matthewmorgan:
Thanks @ryanplusplus !  Keeping us moving forward...

PR #232, opened 2016-06-15
Wrap the function to test

yanick:
... such that if `buffer` is an object, `this` is set to the proper value when `read` is invoked.

### DIFF
--- a/exercises/circular-buffer/circular-buffer.spec.js
+++ b/exercises/circular-buffer/circular-buffer.spec.js
@@ -7,14 +7,14 @@ describe('CircularBuffer', () => {
 
   it('reading an empty buffer throws a BufferEmptyException', () => {
     const buffer = circularBuffer(1);
-    expect(buffer.read).toThrow(bufferEmptyException());
+    expect(()=>buffer.read()).toThrow(bufferEmptyException());
   });
 
   xit('write and read back one item', () => {
     const buffer = circularBuffer(1);
     buffer.write('1');
     expect(buffer.read()).toBe('1');
-    expect(buffer.read).toThrow(bufferEmptyException());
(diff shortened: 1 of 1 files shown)


rchavarria:
The internal code that makes these tests to pass doesn't use `this` inside the `read` method (see [`example.js`](https://github.com/exercism/xecmascript/blob/master/exercises/circular-buffer/example.js) file). So, if we take that code into account, it would be unnecessary to do any modification.

But exercism users provide their own solutions, that can be totally different from the code in this exercise's `example.js`. and they may have problems with `this` inside `read`. So I would like to discuss a little bit your submission.

Did you find any problem while solving the exercise on exercism.io? Do you know anybody that found that problem? Please, don't get me wrong. I'd like to know just why do you think your changes are necessary.

yanick:
The way the interface is presented, one (well, I, at least :-) ) would expect

```
const buf1 = circularBuffer(2);

const buf2 = circularBuffer(1);

buf1.write(1);
buf1.write(2);

```

to work. But it doesn't because the example code uses a singleton (and the test using the un`this`ified function hamstring the uses of encapsulated objects).

If, for the exercise, we /really/ want to implement a singleton class, the calls should be less ambiguous:

```
import circularBuffer from 'crircularBuffer';

circularBuffer.setSize(2);

circularBuffer.write(1);

etc, etc, etc
```

rchavarria:
Yep, I see.

The current example doesn't fail because of the current implementation of tests, and tests don't fail because of the current example. But I see your point when you say

> the test using the un`this`ified function hamstring the uses of encapsulated objects

So, this PR is ready to merge. Thanks for your contribution and this interesting conversation!

yanick:
Yay! Thanks! :-)

PR #234, opened 2016-07-14
Trinary misses an illegal case

rchavarria:
Trinary numbers should contain only digits from `0` to `2`

Trinary numbers have a base of `3`, and they can only be
expressed with digits from `0` to `n < BASE`

### DIFF
--- a/exercises/trinary/example.js
+++ b/exercises/trinary/example.js
@@ -7,8 +7,17 @@ export default class Trinary {
   }
 
   toDecimal () {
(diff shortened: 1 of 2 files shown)


rchavarria:
@matthewmorgan, here it is what we talked through exercism.io

It adds a test case, to check that only valid digits are parsed.

matthewmorgan:
Nice work!  I like how the `isNaN()` test catches negatives, decimals, letters, etc.

rchavarria:
It's not 100% mine. I got some ideas from the xjavascript track ;)

PR #236, opened 2016-08-09
Update Node.js download/installation links

rzoller:
Update link to "Installing-Node.js-via-package-manager" instructions which are now hosted on nodejs.org.
Fix instructions for Windows/OS X and add link to download installer.


matthewmorgan:
@rzoller Thanks for the update!

PR #240, opened 2016-09-08
Add isogram problem to exercises

Roshanjossey:
Added relevant tests and example.js


Roshanjossey:
@rchavarria: I didn't understand how this branch got conflicts with master. Could you share some insights in to this so that I can resolve this and prevent it from happening in my future PRs

matthewmorgan:
@Roshanjossey It doesn't appear that this branch conflicts with `master`.  What makes you think that it does?

Let me know and I can try to help!

Roshanjossey:
@matthewmorgan: Thank you for looking in to this. I resolved it. You see I made another commit after that comment.

The problem was that I added two problems in separate branches. Both were editing `config.json`. When I submitted PR for each, they had no merge conflicts. When one PR was merged, other one had conflict.

Could you tell me if there are any ways to avoid such scenarios?

matthewmorgan:
@Roshanjossey I think that in your situation, the problem occurred because each of your PRs was modifying the same part of the same file -- `config.json`-- , but with different exercise names.  @rchavarria and I had this problem frequently when we were building out this track, as we were alternating contributing exercises, so each new PR was adding a new exercise to the end of `config.json`.

There are three different solutions that I can suggest:
1) Wait to make the second PR until the first is merged.  EG just keep it on your fork or locally.
2) Add new exercises in different parts of `config.json`.  For this to work, the changes must not be on adjacent lines.  The disadvantage to this approach is that the order of the files in `config.json` determines the order in which they are served to users, so you may need to submit a second PR later to correct the order.
3) After the first PR is merged, pull `master` and rebase your changes onto master.  This replays your changes on top of the newest version of master:

```
git master
git pull
git checkout MY-NEW-BRANCH
git rebase master
git push -f
```

Let me know if that all makes sense.

exercises/isogram/example.js
```diff
@@ -0,0 +1,11 @@
+export default class Isogram {
+
+  constructor(string) {
+    this.string = string.replace(/ |-/g, '');
+  }
+
+  isIsogram() {
+    const uniqueLetters = this.string.toLowerCase().split('').filter((element, index, self) => self.indexOf(element) === index);
+    return uniqueLetters.length == this.string.length;
```
matthewmorgan: I would suggest the use of `===` here.

Roshanjossey:
@matthewmorgan: Thank you for your detailed reply. That was exactly the problem. This time I did what was explained in option 3 but I think going forward, I'll use a combination of (1) and (3). ie, only send one PR at a time, have changes in my local and send them only after pulling from upstream master and rebasing. (That way I don't have to use `-f` 😉 ).

I like the suggestion you made. I'll make that change in a moment.

matthewmorgan:
Looks great!  Thanks for the contribution!

rchavarria:
@matthewmorgan Impressive mate! What a wonderful explanation about avoiding conflits. If you have a blog, you should publish that. If not, you should start one ;)

Thanks for taking care of this PR. And thanks @Roshanjossey for your contribution!

PR #241, opened 2016-10-03
Simple cipher misses an illegal case.

marocchino:
It was pass string like '123ABCxyz'.


rchavarria:
Nice point! You're right. The regex was not strong enough.

Thanks for fixing it.

PR #242, opened 2016-10-12
Add test to Grade School

mgmatola:
Grade School has a test to prevent modifications when calling the roster() method. However, there is no test to prevent modifications when calling the grade() method. This PR adds a test to address the grade() scenario.

My code in this submission passes the existing tests, but allows modifications through grade():
http://exercism.io/submissions/971cbf70d109466ba098fa6c0195eb00

### DIFF
--- a/exercises/grade-school/grade-school.spec.js
+++ b/exercises/grade-school/grade-school.spec.js
@@ -70,4 +70,11 @@ describe('School', () => {
     expect(school.roster()).toEqual(expectedDb);
   });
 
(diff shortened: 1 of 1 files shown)


matthewmorgan:
@mgmatola Thanks!

mgmatola:
Grade School in the JavaScript doesn't have either test. Is that intentional? If not, I can do a PR adding those to the JavaScript track.

matthewmorgan:
Why not?  Sounds like a great idea.

Issue #220, opened 2016-04-26
Grains: the provided BigInt implementation fails linter


ryanplusplus:
Running `npm test` with an empty `grains.js` implementation yields:

```shell
~/exercism/ecmascript/grains$ npm test

> xecmascript@0.0.0 test /home/ryan/exercism/ecmascript/grains
> gulp lint test

[20:17:28] Using gulpfile ~/exercism/ecmascript/grains/gulpfile.js
[20:17:28] Starting 'lint'...
[20:17:28] Starting 'babel'...
[20:17:29] 
big-integer.js
   260:18  error  Move variable declaration to function body root  no-inner-declarations
   366:18  error  Move variable declaration to function body root  no-inner-declarations
   505:13  error  Move variable declaration to function body root  no-inner-declarations
   509:17  error  Move variable declaration to function body root  no-inner-declarations
   576:13  error  Unexpected assignment within an 'if' statement   no-cond-assign
   581:9   error  Unexpected constant condition                    no-constant-condition
   614:14  error  Move variable declaration to function body root  no-inner-declarations
   754:6   error  Unnecessary semicolon                            no-extra-semi
   784:14  error  Move variable declaration to function body root  no-inner-declarations
   785:13  error  Move variable declaration to function body root  no-inner-declarations
   823:16  error  Gratuitous parentheses around expression         no-extra-parens
   824:13  error  Gratuitous parentheses around expression         no-extra-parens
   890:14  error  Move variable declaration to function body root  no-inner-declarations
   961:14  error  Move variable declaration to function body root  no-inner-declarations
   962:13  error  Move variable declaration to function body root  no-inner-declarations
   963:13  error  Move variable declaration to function body root  no-inner-declarations
   983:13  error  Move variable declaration to function body root  no-inner-declarations
   988:17  error  Move variable declaration to function body root  no-inner-declarations
  1035:13  error  Move variable declaration to function body root  no-inner-declarations
  1075:17  error  Move variable declaration to function body root  no-inner-declarations
  1085:17  error  Move variable declaration to function body root  no-inner-declarations
  1089:17  error  Move variable declaration to function body root  no-inner-declarations
  1090:17  error  Move variable declaration to function body root  no-inner-declarations
  1132:10  error  Move variable declaration to function body root  no-inner-declarations

✖ 24 problems (24 errors, 0 warnings)

[20:17:29] 'lint' errored after 1.06 s
[20:17:29] ESLintError in plugin 'gulp-eslint'
Message:
[...]

PR #206, opened 2016-03-16
Added a test script to package.json

andreasolund:
Resolves issue #205 with an npm script which simplifies running the test suite.

Issue #205, opened 2016-03-14
Missing test script in package.json


andreasolund:
I noticed that all excercises are missing an npm test script in their package.json. Probably a suffiecient solution is to add: `"test": "gulp test"` in the script block in the master [package.json](https://github.com/exercism/xecmascript/blob/master/package.json) and re-run the _makefile_ that replaces each exercises package.json. 

Note that this was previously discussed in [Pull Request #13](https://github.com/exercism/xecmascript/pull/13/files)

rchavarria:
Yeah, I didn't notice that.

Tests are thought to be run with `gulp test` or `gulp lint test` commands instead of `npm test`. But, there is no a big difference of using one or another. I would be happy to hear some reasons on why we should have it (please, sorry for my poor English vocabulary if this sounds too rude).

@matthewmorgan Any thought about this? I don't mind to have a `test` script on `package.json` files, but I don't see it as necessary.

matthewmorgan:
@rchavarria I have no strong feelings either for or against it.  It would not be a big deal to change it.

@andreasolund I think your suggestion would work-- maybe you would like to create a PR for this change?

rchavarria:
I see it as a common practice for node projects, so it's ok for me too.

andreasolund:
Sure, I can try to make a PR for it tonight. I agree with @rchavarria that the script should be: `"test": "gulp lint test"` to get the desired result. I think it would be great to provide a widely accepted solution that makes the experience feel familiar between languages, like the exercism exercises in Go where you use `go test`!

matthewmorgan:
Closed by #206.  Thanks @rchavarria and @andreasolund !

PR #245, opened 2016-11-21
Update setup to not install `gulp` globally

### LINKED ISSUE #246: Create separate npm scripts for testing and testing+linting

### LINKED ISSUE #247: Don't suggest that gulp be installed globally
Instead point users to the `npm test` script

ryanplusplus:
`gulp` doesn't need to be installed globally because it's set up as an `npm` script in `package.json` for all of the exercises. However, the documentation currently instructs people to install it globally and then run tests directly using `gulp` instead of `npm test`.

Closes #246
Closes #247

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -7,16 +7,6 @@ http://exercism.io/languages/ecmascript
 
 ## Requirements
 
-They are already described in the link above, but just as a
-quick reference:
-
-Install globally a tool to run [Gulp](http://gulpjs.com) if
-it is not installed yet:
-
-```bash
-$ npm install -g gulp-cli
-```
-
 Install assignment dependencies:
 
 ```bash
@@ -28,7 +18,7 @@ $ npm install
 Execute the tests with:
 
 ```bash
-$ gulp test
+$ npm test
 ```
 
 In many test suites all but the first test have been skipped.
--- a/exercises/accumulate/package.json
+++ b/exercises/accumulate/package.json
@@ -18,7 +18,8 @@
     "yargs": "~3.27.0"
   },
   "scripts": {
-    "test": "gulp lint test"
+    "test": "gulp test",
+    "lint-test": "gulp lint test"
   },
   "licenses": [
     "MIT"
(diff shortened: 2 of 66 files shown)


SETUP.md
```diff
@@ -7,16 +7,6 @@ http://exercism.io/languages/ecmascript
 
 ## Requirements
 
-They are already described in the link above, but just as a
```
ryanplusplus: None of these steps are described in the link above so I removed this chunk.

rchavarria:
Ok, I see your point, and you're right.

Some time ago, a user complained about having to lint all the code, so there are two separated tasks on the gulp file: one to lint the code, and one to run the tests.

`npm test` will run both on them. Could you add a new script in `package.json` to run just the tests? They would look something like:

    "lint-test": "gulp lint test"
    "test": "gulp test"

You could add this to this PR. What do you think?

ryanplusplus:
@rchavarria sure, I can do that. Is there a related issue?

rchavarria:
Not specifically to address those two npm scritps, but there have been some discussions around linting and tests:

- You had problems with `big-integer.js` in the grains problem: #220 
- Creating the `test` npm script: #205 and #206 

If they are not what you want, there is no problem of creating a new issue.

ryanplusplus:
@rchavarria I believe that all scripts have been updated. I'm not particularly bothered about creating and immediately closing an issue unless you are.

rchavarria:
Ok, please, create the issue.

Yep, all scripts seem correct, thanks.

Before merging, do you mind squashing the two commits into a single one?

ryanplusplus:
@rchavarria issues created. It is not necessary to squash on my end, you can do it when you merge (see https://github.com/blog/2141-squash-your-commits and https://help.github.com/articles/merging-a-pull-request/).

ryanplusplus:
Thanks!

rchavarria:
Hmmm, I'm not very happy with how the commits have been merged into `master`.

I used the GitHub button, the one that squash PR commits into a new commit on `master`. But I don't like it, because PR author's commits are *lost* in the `master` history. In the attached image, I like the three first merges, but not the last one (this PR) :cry: 

4-merges

Sorry @ryanplusplus, I would prefer having your several commits in `master` than none of them.

PR #249, opened 2016-12-02
Rna-Transcription: Handle invalid nucleotides in strand

### LINKED ISSUE #248: rna-transcription has missing test cases
As per x-common's canonical data, some tests which handle invalid input are missing: https://github.com/exercism/x-common/blob/master/exercises/rna-transcription/canonical-data.json#L40-L54

tejasbubane:
As per [x-common canonical data](https://github.com/exercism/x-common/blob/master/exercises/rna-transcription/canonical-data.json#L40-L54)

Fixes #248

### DIFF
--- a/exercises/rna-transcription/example.js
+++ b/exercises/rna-transcription/example.js
@@ -7,7 +7,14 @@ const DNA_TO_RNA = {
 
 export default class Transcriptor {
   toRna(dna) {
-    return dna.replace(/./g, nucleotide => DNA_TO_RNA[nucleotide]);
+    let rna = dna.replace(/./g, nucleotide => DNA_TO_RNA[nucleotide])
+
+    if (rna.length !== dna.length) {
+      // invalid characters in the strand
+      throw new Error('Invalid input DNA.');
+    }
+    else {
+      return rna;
+    }
   }
 }
-
--- a/exercises/rna-transcription/rna-transcription.spec.js
+++ b/exercises/rna-transcription/rna-transcription.spec.js
@@ -24,4 +24,22 @@ describe('Transcriptor', () => {
         .toEqual('UGCACCAGAAUU');
   });
 
+  xit('correctly handles invalid input', () => {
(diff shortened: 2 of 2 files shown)


matthewmorgan:
@tejasbubane thanks for doing this.  I am wondering if we shouldn't be throwing an exception here instead of returning `null` though?  What are your thoughts?

exercises/rna-transcription/example.js
```diff
@@ -7,7 +7,14 @@ const DNA_TO_RNA = {
...
-    return dna.replace(/./g, nucleotide => DNA_TO_RNA[nucleotide]);
+    let rna = dna.replace(/./g, nucleotide => DNA_TO_RNA[nucleotide])
+
+    if (rna.length !== dna.length) {
+      // invalid characters in the strand
+      return null;
+    }
+    else {
+      return rna;
+    }
```
tejasbubane: Not sure if this is the most elegant solution, but it works 😺

tejasbubane:
@matthewmorgan Yep. I think throwing an exception would be better. Conceptually it is a wrong input and returning `null` makes little sense to the consumer of the API. Also [hamming throws an error](https://github.com/exercism/xecmascript/blob/master/exercises/hamming/example.js#L8) for similar reasons.

tejasbubane:
Strangely, `xruby` changed its behaviour from throwing error to silently returning an empty string. https://github.com/exercism/xruby/pull/358/files. Could not find any history with `x-common` since the file seems to be moved into `exercises` directory.

matthewmorgan:
Are you willing to make the change to throwing an exception then?  If not, I am happy to do it.  Sounds like we agree it should be done.

tejasbubane:
> Were this part of a larger app, that would be a better way to indicate nonsensical input to the consumer.

Agreed. 👍  Will make the change

tejasbubane:
@matthewmorgan fixed.

matthewmorgan:
Lovely!  Thanks again, keep up the good work.

PR #250, opened 2016-12-08
dibs: I will implement all-your-base

javaeeeee:
dibs: I will implement all-your-base


tejasbubane:
PS: We will need to deprecate `binary`, `trinary`, `octal` and `hexadecimal` exercises after we merge this in. ref: https://github.com/exercism/x-common/issues/279 Not sure of the exact method of deprecating though.

matthewmorgan:
@tejasbubane right about that.  @rchavarria the only thing about that process that still seems unclear to me is how to create an association between the old deprecated exercises and this new one.  It's not clear that there's any way to do that at this point, and I think it requires a change to the CLI globally.  We would also need to change this track's config.json file afterward.

Does it make sense to merge this exercise now, or do we need to wait until we have a way to deprecate the old ones?

tejasbubane:
> Does it make sense to merge this exercise now, or do we need to wait until we have a way to deprecate the old ones?

There would be absolutely no harm in merging. This would be served as just another exercise.

In terms of deprecating, I think just adding the exercises to `deprecated` in `config.json` should do the job as per [this document](https://github.com/exercism/x-common/blob/fbae1b26da79a1295048421c9bf00f0f668d074a/CONTRIBUTING.md#configjson). I am not sure though.

rchavarria:
I agree with @tejasbubane: it'll make no harm to add a new exercise.

What to do with `binary` and friends? According to https://github.com/exercism/x-common/issues/279 there is still some work to do, but there are some tracks that have already deprecated them (https://github.com/exercism/x-common/issues/279#issuecomment-264938769). 

I would move `binary` and friends to `deprecated` in `config.json`, and I'd keep exercise implementations, as suggested in https://github.com/exercism/x-common/issues/279#issuecomment-265214041

matthewmorgan:
Well that sounds good to me.

@javaeeeee would you like to move `binary`, `trinary`, and `hexadecimal` to `deprecated`?  If not I can do it as part of another PR.

javaeeeee:
Done!

tejasbubane:
Looks like you forgot to remove the `octal` exercise @javaeeeee

javaeeeee:
Thank you @tejasbubane, OKAY

rchavarria:
Fantastic! Just one little detail. The PR consist of multiple commits. Do you mind squashing them into a single one?

javaeeeee:
Done

rchavarria:
Hmmm, I think there are more commits now. But don't worry, I'll merge them into a single commit into branch.

It seems to me you struggled with git and branches during the squash process. The other day, I talked to another contributor about how to squash commits (see https://github.com/exercism/xjavascript/pull/309#issuecomment-264655066). Let me know if you find the link and the commands useful.

javaeeeee:
Thank you, I'll take a look

PR #252, opened 2016-12-13
Fix non-compiling default palindrome-products file

ErikSchierboom:


rchavarria:
Whoops! Nice catch. It took me a while to see what was the problem. Dashes in the source code and dashes in the diff view seem mixed... I'm too tired.

Thanks for the fix.

I don't know what was the problem with the Travis CI build, but it doesn't matter. I think everything's ok.

PR #253, opened 2016-12-14
Make it more clear that a default constructor is called in queen-attack

ErikSchierboom:


matthewmorgan:
Thanks @ErikSchierboom !

PR #265, opened 2017-03-19
SVG for badge instead of PNG

rootulp:


rchavarria:
Nice! Thanks for the contribution.

As it was merged in the JavaScript track, I think it's ok to merge it here.

I'm just curious: does it improve anything?

rootulp:
> It doesn’t matter about your screen size, zoom level, resolution or whether or not you have a ‘retina’ display. SVG is vector based, paths, shapes and fills. This allows for infinite resolution, much like when creating artwork for print and using shapes, the same principles apply!

taken from: [5 reasons you should be using SVG’s over PNG’s](https://watb.co.uk/5-reasons-you-should-be-using-svgs-over-pngs/)

matthewmorgan:
@rchavarria @rootulp I checked this by zooming that tiny logo file to a very large size, and it never got jaggy.  Looks good!

rchavarria:
Oh, yeah! That's *the* difference between vector and bitmap graphics :smiley: 

Many, many thanks @rootulp !!

PR #271, opened 2017-04-10
Update tooling to Babel 6, Node 7.8

### LINKED ISSUE #267: Update infrastructure to use NodeJS 6+, `latest` babel presets
@exercism/javascript We are using somewhat older versions of Node and Babel.  I was going to suggest to a contributor that he update his solution to incorporate generators, but he wouldn't have been able to do that without an update on our end.

Any concerns?  If not, let's make it happen!

matthewmorgan:
Resolves #267 @rchavarria mind taking a look at this one for me?  Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-- '5.7'
+- '7.8'
 sudo: false
 script:
 - bin/fetch-configlet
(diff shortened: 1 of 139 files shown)


rchavarria:
Travis CI took a look before me, and he says it's ok :wink: 

The changes look ok. I would have used Node v6.10 instead of v7.8, but it's not a big issue. The real change is to use a newer version of Babel.

Thanks a lot for solving this. I really really appreciate it!!

matthewmorgan:
It was fun!  I hope will enjoy using the latest toys...

PR #273, opened 2017-04-18
Make hello-world more simple

tejasbubane:
by always expecting "Hello, World!"

Change in x-common: [here](https://github.com/exercism/x-common/pull/544)

Upcoming new exercise to kinda replace `hello-world`: [here](https://github.com/exercism/x-common/pull/759)

Since this `hello-world` is just meant to be a sanity check for everything working, I also went ahead and removed the use of class for the sake of simplicity.
We can introduce classes with the next `two-fer` exercise.

### DIFF
--- a/exercises/hello-world/example.js
+++ b/exercises/hello-world/example.js
@@ -1,9 +1,6 @@
 
-class HelloWorld {
-    hello(name = 'World') {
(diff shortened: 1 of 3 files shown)


matthewmorgan:
Does what it says on the box!  Nice.

PR #276, opened 2017-04-26
Maintain class consistency

tejasbubane:
* As per @rchavarria's comment https://github.com/exercism/xecmascript/issues/274#issuecomment-297439261, do not make users switch between classes and functions for solutions. Only `hello-world` and `leap` exercises were expecting functions, convert their specs to expect classes instead.
* Also took the opportunity to remove some specs from `leap` as per the `x-common`. Keep only what are absolute necessary.

Kept the commits separate since I think they do different things.

### DIFF
--- a/exercises/hello-world/example.js
+++ b/exercises/hello-world/example.js
@@ -1,6 +1,7 @@
-
-const helloWorld = () => {
-  return 'Hello, World!';
+class HelloWorld {
+  hello() {
+    return "Hello, World!";
+  }
 }
 
-export default helloWorld;
+export default HelloWorld;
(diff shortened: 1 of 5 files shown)


rchavarria:
:+1: to simplified Leap year exercise. Yep, x-common changed specs some time ago.

I'm ok with changing to classes the two first exercises. But I'd like to wait just in case somebody at @exercism/javascript has something to add.

PR usually consist on just one commit, but you mentioned explicitly you want two commits. No a big issue.

tejasbubane:
> PR usually consist on just one commit, but you mentioned explicitly you want two commits. No a big issue.

If you want, I can split these into two PRs. Let me know.

matthewmorgan:
Nicely done!  I especially like the simplification of the test cases.

Issue #272, opened 2017-04-17
[Proposal] Use the Ava Test runner


tejasbubane:
[Ava](https://github.com/avajs/ava) is a new javascript test runner that claims to be futuristic. It built-in babel which allows us to [write and run tests in ES2017](https://github.com/avajs/ava#es2017-support) without the need of any other tool. 

Only thing here is that it does not compile the source code, just the tests. As mentioned in ava's README, this can be easily fixed by using [babel-register](https://babeljs.io/docs/usage/babel-register/) which hooks into ava and transpiles ESNext code on the fly while running tests. 

All it takes is this small snippet of code in `package.json`:

```json
"ava": {
    "require": [
      "babel-register"
    ]
 }
```

For our use case (of small exercises), this completely eliminates the need for bundling tools like gulp and to maintain the [huge gulpfile](https://github.com/exercism/xecmascript/blob/master/gulpfile.js) and pass it along to the users with each exercise. Only thing we need to pass to users is `package.json`.

To demonstrate what it looks like I ported the first two exercises `hello` and `leap` to use ava tests in [this repository](https://github.com/tejasbubane/xecmascript-ava). `hello` uses the standard `xunit` style that ava gives out of the box. While `leap` uses the BDD style which is provided by [ava-spec](https://github.com/sheerun/ava-spec).

PS: I also took the opportunity to use the latest ES using [babel-preset-env](https://github.com/tejasbubane/xecmascript-ava/blob/master/leap/package.json#L25) which addresses my previous issue #259.

If maintainers agree, I would like to work on this.

matthewmorgan:
@tejasbubane this looks very cool.  Myself I have been thinking about proposing migrating to using [Jest](https://github.com/facebook/jest), which appears to provide similar benefits.

Both projects appear similar in popularity, I'm not sure there's a compelling case for migrating to one versus the other.  My familiarity with Jest comes from its association with React, which is of course enormously popular these days.

I think there is another consideration involved here though: the gulpfile provides an example of what a typical Javascript build pipeline looks like these days, and I think that's invaluable.  Not every user will investigate that, but we should consider that aspect as well.

I have also been thinking about proposing changing the build to use [Webpack](https://webpack.js.org/), which has value for the same reasons that using Gulp has, but is more current technology.

@rchavarria what do you think about this?  Is it better to have a dead simple exercise environment, or to help expose users to build tools?  @tejasbubane what are your thoughts?

tejasbubane:
[...]

PR #277, opened 2017-04-29
Use Jest test runner

matthewmorgan:
As discussed in #272 , this is a variation of a terrific idea from @tejasbubane .

It greatly simplifies our scaffolding around this project.  My goals were:

1. Replace Jasmine, Gulp, and Babel with Jest and Babel-Jest.
2. Retain eslint for linting support

As suggested by @rchavarria, the babel config is in the `package.json` file.  I was also able to move the eslint config here as well.

I had one issue with an array matcher in the "Palindrome Products" exercise that was not behaving, so I had to add some JS functions to specfile to get the desired test results. 

This is a big PR in appearance, but mostly removes some existing scaffolding.  The changes are conceptually small but are duplicated many times.

UPDATE: I realized I missed changing two of the `/doc` files, so I have updated those.


exercises/all-your-base/package.json
```diff
@@ -9,22 +9,70 @@
...
-    "gulp": "~3.9.0",
-    "gulp-babel": "^6.1.2",
-    "gulp-eslint": "^1.1.0",
-    "gulp-jasmine": "~2.4.2",
-    "yargs": "~3.27.0"
+    "eslint": "^3.19.0",
+    "jest": "^19.0.2"
+  },
+  "jest": {
+    "modulePathIgnorePatterns": [
```
matthewmorgan: This was needed to prevent Jest from warning about multiple `package.json` files in the project tree.

tejasbubane:
I wonder what the issue with `palindrome-products` is since jest also has the array `toContain` matcher.

matthewmorgan:
@tejasbubane I don't know either why the matcher isn't working.  It looks like the Jest version of the matcher does not handle nested arrays, EG it doesn't see `[ [3, 3], [1, 9]]` as containing `[3, 3]`.  I had to unpack both arrays and compare them element by element.

matthewmorgan:
One more commit to copy the new `package.json` file with `jest --watch` to all the exercise dirs.

rchavarria:
Awesome! Awesome! Awesome! Many, many thanks for this great work!!!

PR #279, opened 2017-05-01
Throw error for invalid input of perfect-numbers

tejasbubane:
The [canonical-data](https://github.com/exercism/x-common/blob/master/exercises/perfect-numbers/canonical-data.json#L86-L106) says it should be an error. Raise an error instead of just returning
the string.

### DIFF
--- a/exercises/perfect-numbers/example.js
+++ b/exercises/perfect-numbers/example.js
@@ -42,7 +42,7 @@ export default class PerfectNumbers {
 
     // Check if the input is valid
     if (number <= 0) {
-      return 'Classification is only possible for natural numbers.';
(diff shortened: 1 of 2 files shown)


matthewmorgan:
Yeah, @tejasbubane in some contexts, returning any kind of any object (like that string) will suffice for the error handling, but it should be more explicit.  This is a better way.

Looks like the previous PR created a merge conflict.  If you want to fix, I can merge this after.  Thanks!

tejasbubane:
Fixed the conflicts @matthewmorgan

matthewmorgan:
:100: thanks!

PR #281, opened 2017-05-08
Use classes for custom exceptions in `circular-buffer`

### LINKED ISSUE #280: Custom Exceptions are confusing
I was solving `Circular Buffer` today and had a hard time figuring out how to create custom exceptions. That is because I was expecting the matcher to match with a `custom` exception class, but what it actually does is matches with `Error` itself. 

```js
expect(()=>buffer.read()).toThrow(bufferEmptyException());
```
Note `bufferEmptyException` is a function call which gives no idea what its return type should be - `Error`/`CustomError`/`String`??

For custom exceptions in ES6, most [sources](http://exploringjs.com/es6/ch_core-features.html#sec_from-custom-error-to-error-subclass) say that this should suffice: (which is far better than its ES5 equivalent)

```js
class bufferFullException extends Error {}
```

But it does not work since [babel does not allow subclassing built-in classes](https://github.com/facebook/jest/issues/2123) (also mentioned in the [docs](https://babeljs.io/docs/usage/caveats/#classes)). I ended up using the [babel-plugin](https://github.com/loganfsmyth/babel-plugin-transform-builtin-extend) for allowing subclassing. Solution [here](http://exercism.io/submissions/d5f75a974880456db6154f39e39051f6).

Most other exercises just throw a String, while others like `wordy` and `circular-buffer` throw custom exceptions. 
So I have two questions:

* Should we have a standard way of doing exceptions in the track? Is throwing Strings OK or should we go all custom?

* For the exercises that expect custom errors, is `class bufferFullException extends Error {}` better than what we have right now? which is:
```js
const BufferFullException = () => ({
    name: "buffer full exception!",
    message: "can't write to a full buffer!"
  });
```

tejasbubane:
Closes #280.

### DIFF
--- a/exercises/circular-buffer/example.js
+++ b/exercises/circular-buffer/example.js
@@ -1,25 +1,28 @@
 let buffer, bufferMax;
 
-const BufferEmptyException = () => ({
-    name: "buffer empty exception!",
-    message: "can't read from an empty buffer!"
-  });
-
-const BufferFullException = () => ({
-    name: "buffer full exception!",
-    message: "can't write to a full buffer!"
-  });
+export class BufferEmptyError extends Error {
+  constructor(message) {
+    super();
+    this.message = message || 'Buffer is empty.';
+  }
+};
+export class BufferFullError extends Error {
+  constructor(message) {
+    super();
+    this.message = message || 'Buffer is full.';
+  }
+};
 
 const read = () => {
   if (buffer.length === 0){
-    throw BufferEmptyException();
+    throw new BufferEmptyError();
   }
   return buffer.splice(0,1)[0];
 };
 
 const write = (value) => {
   if (buffer.length === bufferMax){
-    throw BufferFullException();
+    throw new BufferFullError();
   }
   value ? buffer.push(value) : null;
 };
@@ -41,9 +44,7 @@ const CircularBuffer = (capacity) => {
     write: write,
     forceWrite: forceWrite,
     clear: clear
-  }
+  };
 };
 
 export { CircularBuffer as default };
-export { BufferFullException as bufferFullException };
-export { BufferEmptyException as bufferEmptyException };
--- a/exercises/accumulate/package.json
+++ b/exercises/accumulate/package.json
@@ -11,6 +11,7 @@
   "devDependencies": {
     "babel-jest": "^20.0.1",
     "babel-preset-env": "^1.4.0",
+    "babel-plugin-transform-builtin-extend": "^1.1.2",
     "eslint": "^3.19.0",
     "jest": "^20.0.1"
   },
@@ -22,6 +23,11 @@
   "babel": {
     "presets": [
       "env"
+    ],
+    "plugins": [
+      ["babel-plugin-transform-builtin-extend", {
+        "globals": ["Error"]
+      }]
     ]
   },
   "scripts": {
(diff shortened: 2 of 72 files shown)


exercises/circular-buffer/example.js
```diff
@@ -1,25 +1,28 @@
...
+  constructor(message) {
+    super();
+    this.message = message || 'Buffer is empty.';
+  }
+};
+export class BufferFullException extends Error {
+  constructor(message) {
+    super();
+    this.message = message || 'Buffer is full.';
+  }
```
tejasbubane: The constructor part is optional. But it reads nicely in the stacktrace with the message.

tejasbubane:
For now, I have added the plugin to global `package.json` and the one of `circular-buffer`.
One problem with using the plugin in our setup is we will need to copy the main `package.json` file over in all exercises for consistency even though all won't be using custom exceptions.

exercises/circular-buffer/example.js
```diff
@@ -1,25 +1,28 @@
...
-const BufferEmptyException = () => ({
-    name: "buffer empty exception!",
-    message: "can't read from an empty buffer!"
-  });
-
-const BufferFullException = () => ({
-    name: "buffer full exception!",
-    message: "can't write to a full buffer!"
-  });
+export class BufferEmptyException extends Error {
```
tejasbubane: On second thought, should this be `BufferEmptyException` or `BufferEmptyError` (since it inherits from `Error`)?

matthewmorgan: `Error` is probably better semantically.

matthewmorgan:
@tejasbubane once this PR is merged I can re-run the makefile, which will copy the global `package.json` to each exercise, so that will fix that.  Thanks!

exercises/circular-buffer/example.js
tejasbubane: Changed it to `Error`.

tejasbubane:
> I can re-run the makefile, which will copy the global package.json to each exercise, so that will fix that.

Do you think I should do that in this PR itself?

matthewmorgan:
@tejasbubane for sure-- if you're up for that, I think it's fine in this PR.

tejasbubane:
> if you're up for that, I think it's fine in this PR.

Done. @matthewmorgan

rchavarria:
Sorry. I can't have an opinion on this, because I didn't try myself Jest yet. This is something I must do. I must update locally this track and run all the tests.

matthewmorgan:
@rchavarria if you want us to wait before merging this PR until you've had a chance to run it locally, please let me know.  I'll hold off until tomorrow to give you some more time.

joelwallis:
Same to me. Haven't tried Jest yet :(

matthewmorgan:
@tejasbubane--do you mind resolving the merge conflicts, then I will merge it in?  Thanks!

rchavarria:
Don't worry, don't wait for me. I'll catch up.

matthewmorgan:
@tejasbubane bump :)

tejasbubane:
@matthewmorgan Done. Sorry for the delay.

rchavarria:
I'm having an error running `make test`, but as it passes on Travis, I'll merge this.

```
~/xecmascript • make test
Makefile:19: recipe for target 'test-package-files' failed
make[1]: *** [test-package-files] Error 1
Makefile:34: recipe for target 'test' failed
make: *** [test] Error 2
```

rchavarria:
Sorry for my last comment, I didn't read #302

PR #286, opened 2017-05-12
Upgrade Jest to v20

tejasbubane:
* Also upgrade couple other packages like `babel-jest` and `babel-preset-env`.

* Check-in the yarn.lock file in root directory.


tejasbubane:
This might conflict with #281

matthewmorgan:
Let's do this first, we can always change it later.   Thanks!

PR #289, opened 2017-05-12
Use local jest instead of global in Makefile

tejasbubane:
The Makefile was relying on global installation of jest to run the
tests. This failed for me since I did not have jest globally
installed. Also the README does not mention about installing globally
which is anyways not a good idea (might cause issues due to different versions).

Hence, change Makefile to use locally installed jest instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,4 @@ sudo: false
 script:
 - bin/fetch-configlet
 - bin/configlet .
-- npm install
 - make test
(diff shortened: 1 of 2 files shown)


tejasbubane:
Looks like travis does not recognize `./node_modules` path 😞

matthewmorgan:
@tejasbubane I think your directory path is wrong.  This works for me:

```bash
@./node_modules/.bin/jest $(OUTDIR)/temp.*.spec.js
```
Since I can't push to your fork, want to give that a try?

tejasbubane:
Travis was not happy with the `./` in the beginning. Removing that worked 😄 .

> @./node_modules/.bin/jest $(OUTDIR)/temp.*.spec.js

I forgot node_modules creates these bins (they are just symlinks to the filepath I have given). Better to use these than to dig inside the jest-cli directory. Will change it.

matthewmorgan:
Great!  Good to get this cleaned up.

tejasbubane:
I also removed the `npm install` from `travis.yml` since travis will automatically do that when it finds `package.json` in project root.

PR #291, opened 2017-05-13
Run jest only once while running all tests

tejasbubane:
I am using a single directory named `tmp_exercises` and dumping all exercise and test files there. This makes it easy to run `jest` command over the entire directory in one-shot instead of running it over single exercise.

Speeds up the tests considerably - which means faster builds! 🎉 

**Before:**
`make test  106.12s user 24.49s system 101% cpu 2:09.03 total`

**After:**
`make test  20.53s user 2.75s system 106% cpu 21.872 total`

This is possible because jest parallelizes tests across workers.

Also added default target to Makefile. Which means we can run `make` instead of `make test`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.DEFAULT_GOAL := test
+
 # assignments
 ASSIGNMENT ?= ""
 IGNOREDIRS := "^(\.git|docs|bin|node_modules|.idea)$$"
--- a/Makefile
+++ b/Makefile
@@ -4,24 +4,29 @@ IGNOREDIRS := "^(\.git|docs|bin|node_modules|.idea)$$"
 ASSIGNMENTS = $(shell find ./exercises -maxdepth 1 -mindepth 1 -type d | cut -d'/' -f3 | sort | grep -Ev $(IGNOREDIRS))
 
 # output directories
-OUTDIR ?= "exercises/$(ASSIGNMENT)/tmp"
+OUTDIR ?= "tmp_exercises/$(ASSIGNMENT)"
 
 # language specific config (tweakable per language)
 FILEEXT := "js"
 EXAMPLE := "example.$(FILEEXT)"
 TSTFILE := "$(subst _,-,$(ASSIGNMENT)).spec.$(FILEEXT)"
 
-test-assignment:
+copy-assignment:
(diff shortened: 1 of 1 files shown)


matthewmorgan:
@tejasbubane this is interesting, and the performance gain is impressive!  I have a couple of questions:

1. Do you think there's value in a contributor being able to run the tests locally in a way that's grouped by exercise?  Because with this change, we would get tests in a non-deterministic order.  Perhaps what we might want to do is to enable this mode for Travis, but turn it off locally...IE, Travis doesn't care, fast builds are great, but helpful feedback is great too. :)
2. When I run `make` locally, I get this output:

```bash
Test Suites: 132 passed, 132 total
Tests:       623 skipped, 755 passed, 1378 total
Snapshots:   0 total
Time:        4.657s
Ran all test suites matching "tmp_exercises/".
```

I _think_ what is happening is that Jest is running all the tests in `/tmp_exercises`, but also all the tests in `/exercises/*/`.  That should probably be fixed, no?

matthewmorgan:
@tejas let me know if I can help with this one.

tejasbubane:
Thanks @matthewmorgan! There is certainly some issue with the numbers there. I did not find time to get back to this one. Will fix it tonight and let you know if any issue.

matthewmorgan:
@tejasbubane have you had a chance to look at this?  Is there anything we can do to help you finish it?

BTW I did some checking and it looks like Travis supplies the env `TRAVIS=true` by default.  That could be the gatekeeper for running the tests as a group rather than per-exercise.

matthewmorgan:
@tejasbubane are you planning on implementing this, or should we close the PR for now?  Either way is OK, please let me know if I can help.

joelwallis:
The work @tejasbubane did in this PR is a really good one. Faster builds are always [GG](http://www.urbandictionary.com/define.php?term=GG). Let's keep this one alive.

@tejasbubane if you don't have the available time or something that might block you from getting this one merged I would be happy to help you on it. Your commits are currently in your fork, so I can't commit in your branch. You could add me or @matthewmorgan as contributor to your fork, so we can continue from where you are.

matthewmorgan:
@tejasbubane Thanks for initiating this work.  However, because the PR has been open for three months now, I'm going to close it in favor of a newer one that addresses some of the concerns here.

PR #292, opened 2017-05-17
Add the connect exercise

arthurchipdean:


matthewmorgan:
@arthurchipdean thanks again for your contribution!  

I'd like to review your code and suggest some changes that will make it more "ES-6-y".  None of these changes are functionally critical, but I think that the example solutions are examined by some as guidance on ES6 vs ES5, so there are some things you could change to make it useful to the community in that regard.

I'm tied up for a bit here but will get some feedback to you shortly.

exercises/connect/connect.spec.js
```diff
@@ -0,0 +1,110 @@
...
+      ". . . . .",
+      " . . . . .",
+      "  . . . . .",
+      "   . . . . .",
+      "    . . . . ."
+    ];
+    expect(new Board(board).winner()).toEqual("");
+  });
+
+  it("X can win on a 1x1 board", () => {
```
matthewmorgan: All tests except the first one should be declared with `xit` so they're skipped by default.  I won't call this out everywhere, but it should be changed for this test and all the remaining ones.

matthewmorgan: All tests except the first one should be declared with xit so they're skipped by default. I won't call this out everywhere, but it should be changed for this test and all the remaining ones.

exercises/connect/example.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+  startPositions(XorO) {
+    return XorO === "X" ?
+      this.board.map((pos, i) => ({x:i,y:i})) :
+      this.board[0].map((pos, i) => ({x:0, y:i}));
+  }
+  winningSpot({x,y},XorO) {
+    return XorO === "X" ?
+      y === this.board[0].length - 1 + x:
+      x ===  this.board.length - 1;
```
matthewmorgan: Looks like you have an extra space here right of `===`.

exercises/connect/example.js
```diff
@@ -0,0 +1,68 @@
...
+    if(! this.matches(pos, XorO)) {
+      return false;
+    }
+    if(this.winningSpot(pos, XorO)) {
+      return true;
+    }
+    checked = checked.slice(0);
+    checked.push(pos);
+    var matches = this.neighbors(pos).filter(({x,y}) =>{
+      return  this.matches({x,y}, XorO) && checked.filter((spot) => spot.x === x && spot.y === y).length === 0;
```
matthewmorgan: There's an extra space here right of `return`.

exercises/connect/example.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * "Player O" plays from top to bottom, "Player X" plays from left to right.
+ * @param board
+ */
+export default class {
+  constructor(board) {
+    this.board = board.map(b => b.split(""));
+  }
+  winner() {
+    var players = ['X','O'];
```
matthewmorgan: Let's use `let` instead of `var`.

exercises/connect/example.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * "Player O" plays from top to bottom, "Player X" plays from left to right.
+ * @param board
+ */
+export default class {
+  constructor(board) {
+    this.board = board.map(b => b.split(""));
```
matthewmorgan: You can use the (spread)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator] operator here in place of `split`.

exercises/connect/example.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+  winner() {
+    var players = ['X','O'];
+    for(let i = 0; i < 2; i++) {
```
matthewmorgan: Let's use [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) in place of an index, since we're only using the index to get each element.

exercises/connect/example.js
```diff
@@ -0,0 +1,68 @@
...
+      if(this.checkWin(players[i])) {
+        return players[i];
+      }
+    }
+    return "";
+  }
+  checkWin(player) {
+    let positions = this.startPositions(player);
+    for(var i = 0; i < positions.length; i++) {
```
matthewmorgan: Let's use [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) in place of an index, since we're only using the index to get each element.

exercises/connect/example.js
```diff
@@ -0,0 +1,68 @@
...
+  search(pos, XorO, checked) {
+    if(! this.matches(pos, XorO)) {
+      return false;
+    }
+    if(this.winningSpot(pos, XorO)) {
+      return true;
+    }
+    checked = checked.slice(0);
+    checked.push(pos);
+    var matches = this.neighbors(pos).filter(({x,y}) =>{
```
matthewmorgan: `let`, or `const` if you never expect to reassign the pointer.

exercises/connect/example.js
arthurchipdean: @matthewmorgan  How would I use the spread operator to split a string?

matthewmorgan:
Awesome! Thanks again!

rchavarria:
GREAT job you both @arthurchipdean and @matthewmorgan !

PR #293, opened 2017-05-19
Add first patch labels to probot exempt

tejasbubane:
We want to keep these issues open indefinitely for first-timers.
Tell probot to ignore them so that they are never closed
automatically.

`pinned` and `security` are default. I added `good first patch` and
`first-timers only`.


matthewmorgan:
@tejasbubane would you like to do this for Javascript as well?

PR #294, opened 2017-05-24
Validates exercises' package.json files

### LINKED ISSUE #290: Validate package.json in all tracks
We are relying on copying `package.json` in all exercises. It is difficult to check all those file changes in the PR eg. #286 has 68 file changes. Although the copying is automated using Makefile, I think we should be extra-cautious since these files are directly delivered to users.

We can add some CI check similar to [configlet](https://github.com/exercism/xecmascript/blob/master/bin/fetch-configlet) but for our numerous `package.json` files. We can check for json structure and that all files are identical. This will make sure someone did not change anything unexpected in the PR.

This of-course depends on https://github.com/exercism/xecmascript/issues/235. If we are unable to find any way to remove those numerous package.json files, we should make sure to keep them valid and checked in CI.

joelwallis:
**TL;DR:** Basically, I'm checking if all exercises' `package.json` MD5 hashes are the same as the one in project's root directory.

This PR closes #290


joelwallis:
This might fail because today we already have some files that are different than the `package.json` in root.

joelwallis:
FYI: I'm not really a Makefile programmer 😂 So I tried my best to adapt the `Makefile` to make that checking - but any suggestions be appreciated! ☺️

rchavarria:
CI reports no errors, so something must not be working according to your plans 😢 
[CI logs](https://travis-ci.org/exercism/xecmascript/builds/235483976?utm_source=github_status&utm_medium=notification), look at lines 230-231

```
/bin/sh: 1: md5: not found
xargs: md5: No such file or directory
```

joelwallis:
@rchavarria Hm, that's probably because Travis is running the CI on a Linux machine that doesn't have the `md5` utility available in macOS.

I'll research about hashing utilities in Debian/Ubuntu Linux today to fix it.

rchavarria:
I just realize that, on the xjavascript track, we're installing `jasmine` in Travis CI. Look at the *section* `install` in the [`.travis.yml` file](https://github.com/exercism/xjavascript/blob/master/.travis.yml).

Maybe, you can get `apt-get` installing an `md5` utility.

tejasbubane:
I was wondering if we can use [cmp](https://linux.die.net/man/1/cmp) instead.

joelwallis:
There's a `md5sum` available in Debian based distros - and Travis uses Ubuntu as OS. I wrote a script that encapsulates that call.

@tejasbubane I'll check if `cmp` would be better than a MD5 approach. The best option would be to use solutions that are readily available from popular OSes.

joelwallis:
Finally got some time to finish this! 🎉

I just broke the build - and it's weird to be happy for breaking the build! 😂 It's currently breaking because the file `exercises/connect/package.json` is not the same as the `package.json`:

```
❯ find ./exercises/*/* -name package.json | xargs md5 | grep -v "`md5 -q ./package.json`"
MD5 (./exercises/connect/package.json) = 61ca18ec644e8c9170124e986ed3068d
```

Should I be fixing it too directly from this PR?

joelwallis:
Also, another very important thing is that our scripts residing in the `bin/` folder does not follow the indentation rules described in `.editorconfig`. I got it working by adding  the following lines to the file:

```
###########################################################
; make
###########################################################

[Makefile]
indent_style = tab

[makefile]
indent_style = tab
```

Here's the output of `git diff ./.editorconfig`:

```
diff --git a/.editorconfig b/.editorconfig
index 7b1c78f..f97ad9a 100644
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,6 +18,13 @@ indent_style = space
 indent_size = 2

 ###########################################################
+; bash
+###########################################################
+
+[{bin/*,*.sh}]
+indent_size = 4
+
+###########################################################
 ; make
 ###########################################################
```

Should I include this in the PR too?

matthewmorgan:
@joelwallis thanks, this is great!  Good call on checking the hashes first, to make the build fail quickly if they don't make.  Also, I think updating `.editorconfig` and the `connect` exercise are fine.  

Nice work!

rchavarria:
I'd prefer using spaces instead of tabs for indenting.

I recall to try to use spaces in `Makefile`, but then the `make` command failed :( I don't know why. 

Maybe now, that we have a Joel *bash master* Wallis, we could try to change `Makefile` to use spaces. Just saying.

joelwallis:
Omg! You just made my day! Hahahaha

Btw, I know nothing about Makefiles, but I can try to make it use spaces
instead of tabs. I would rather prefer spaces too! They are more consistent
across editors and our `.editorconfig` already enforces it as default. So
yeah it makes sense.

Joel Wallis Jucá
joelwallis.com

barraponto:
AFAIK, tabs are mandatory: https://stackoverflow.com/questions/2131213/can-you-make-valid-makefiles-without-tab-characters

rchavarria:
Thanks @barraponto! If they're mandatory, there is no discussion.

PR #297, opened 2017-05-31
Simplify ESLint configuration by extending a default set of rules

rchavarria:
This PR creates a complete ESLint configuration, based on a default set of rules provided by ESLint itself.

Then, some rules are disabled becaused they report a relative big number of errors and I prefer discussing them before spending time fixing them (just in case we decide we don't want those rules).

Anyway, some reported errors are fixed, but not too many.

Maybe, this PR has conflicts, because it modifies `package.json` (and `package.json` on each individual exercise), but I'd like to start discussing ESLint rules.


matthewmorgan:
👍 @rchavarria!

PR #302, opened 2017-06-02
Fix test package files task

### LINKED ISSUE #301: Improvements to `test-package-files`
* **Checks for package.json in `node_modules` as well.**

`test-package-files` task from Makefile checks for [all package.json from exercises and its subdirectories](https://github.com/exercism/xecmascript/blob/master/Makefile#L16) which also includes any `node_modules` inside. eg. `exercises/phone-number/node_modules/{some-package}/package.json`. This is usually not the case, but happened with me since I had created the phone-number exercise and had tested in the same directory. The task fails since these file obviously don't match.

* **Error does not show filename.**

Filename which does not match would be helpful here
```
⇒  make test
make[1]: *** [test-package-files] Error 1
make: *** [test] Error 2
```

Action items:

- [x] Show filenames in error msg for `test-package-files`.
- [x] Ignore `node_modules` inside exercises.

tejasbubane:
* Show filenames in error msg for test-package-files
* Ignore node_modules inside exercises

Fixes #301

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ TSTFILE := "$(subst _,-,$(ASSIGNMENT)).spec.$(FILEEXT)"
 
 # package.json MD5 hash
 SOURCE_PKG_MD5 ?= "`./bin/md5-hash ./package.json`"
-PKG_FILES= $(shell find ./exercises/*/* -name package.json)
+PKG_FILES= $(shell find ./exercises/*/* -maxdepth 1 -name package.json)
 
 test-package-files:
 	@for pkg in $(PKG_FILES); do \
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,9 @@ SOURCE_PKG_MD5 ?= "`./bin/md5-hash ./package.json`"
 PKG_FILES= $(shell find ./exercises/*/* -name package.json)
 
 test-package-files:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,10 +13,12 @@ TSTFILE := "$(subst _,-,$(ASSIGNMENT)).spec.$(FILEEXT)"
 
 # package.json MD5 hash
 SOURCE_PKG_MD5 ?= "`./bin/md5-hash ./package.json`"
-PKG_FILES= $(shell find ./exercises/*/* -name package.json)
+PKG_FILES= $(shell find ./exercises/*/* -maxdepth 1 -name package.json)
 
 test-package-files:
-	@for pkg in $(PKG_FILES); do ! ./bin/md5-hash $$pkg | grep -qv $(SOURCE_PKG_MD5) || exit 1; done
+	@for pkg in $(PKG_FILES); do \
```
rchavarria: This looks much better than a oneliner, thanks!

rchavarria:
Now, `make test` works perfectly.

Well, I had to install a newer version of NodeJS. Otherwise, `circular-buffer` exercise was failing. Just in case anyone has the same issue with exercises. I installed NodeJS v7.10.0

matthewmorgan:
@tejasbubane Thanks!  @rchavarria thanks for your input also!

PR #311, opened 2017-06-29
Lint using Airbnb style guide

rchavarria:
First, we were talking about using `standard` as a guide for linting, but it relies on ASI (Automatic Semicolon Insertion) and it expects JavaScript code not to finish each line with `;`. Currently, our code is just the opposite. I feel more confortable using `;` at the end of lines (Java background) and I felt that was the feeling too for other maintainers (maybe I'm wrong, sorry).

So, I decided to use Airbnb as the base to lint code in this track. I needed to add some new dependencies in `package.json`. Then, the ESLint configuration extends from the `airbnb` one. Search in `package.json` files for `"extends": "airbnb"` in the `eslint` section.

There are some rules that are still disabled, because they report a big number of errors. The idea is to enable them one by one.

There are other rules that MUST be disabled: `import/no-unresolved` and `import/extensions` so far. They must be disabled because they check ES6 imports. In tests, we have imports such as `import Bob from './bob'`. It looks for a file called `bob.js` in the same folder. But, during development, `bob.js` (`<exercise>.js`) is called `example.js`. So, ESLint reports an error for each `import` statement like that. And we can't modify them to `import Bob from './example'` because it will make tests to fail.

Disabled rules that have to be enabled:

- [ ] no-unused-expressions
- [ ] one-var
- [ ] no-param-reassign
- [ ] no-plusplus
- [ ] eqeqeq
- [ ] no-mixed-operators
- [ ] no-restricted-properties
- [ ] no-return-assign
- [ ] class-methods-use-this
- [ ] no-new
- [ ] no-use-before-define
- [ ] no-bitwise
- [ ] no-shadow
- [ ] no-confusing-arrow
- [ ] consistent-return
- [ ] array-callback-return
- [ ] no-multi-str
- [ ] import/newline-after-import
- [ ] no-underscore-dangle
- [ ] prefer-const
- [ ] no-array-constructor
- [ ] radix
- [ ] no-multi-assign
- [ ] no-restricted-syntax
- [ ] default-case
- [ ] import/no-duplicates
- [ ] valid-typeof
- [ ] prefer-rest-params
- [ ] func-names
- [ ] no-prototype-builtins


rchavarria:
This is proved to be a wrong approach. I've spend more time resolving conflicts than doing interesting work. And latest conflicts look very tedious to me.

I'm going to close this in favor of #317

PR #312, opened 2017-06-30
Add exercise Flatten-Array

matthewmorgan:
I used the same position in `config.json` as on the `javascript` track.  Thought 5 might be a good difficulty because of the recursion-- what do people think?


exercises/flatten-array/example.js
```diff
@@ -0,0 +1,7 @@
+export default class Flattener {
+  flatten(arr) {
+    return arr
+      .reduce((acc, el) => Array.isArray(el) ? acc.concat(this.flatten(el)) : acc.concat(el), [])
```
joelwallis: I would spread the expressions in the ternary operator on multiple lines to improve the readability of the solution:

```js
.reduce((acc, el) => {
  return Array.isArray(el)
    ? acc.concat(this.flatten(el))
    : acc.concat(el)
}, [])
```

Btw, that's a very elegant usage of recursion to solve this issue, Mat!

matthewmorgan: I think that's a great suggestion @joelwallis !  The more I do this, the more I favor readability.

joelwallis:
Mat, you might want to add test cases for `undefined` values too.

matthewmorgan:
@joelwallis I had considered that, but decided against it.  The canonical tests do have `null` values, but not `undefined`.

Having said that, `undefined` is an important concept in Ecmascript, so I could be convinced otherwise.  Do you think it would add value for the users, beyond the null checks?  If anyone else has thoughts, please feel free to weigh in.

rchavarria:
This is great! It's awesome there are still exercises that can be added to these tracks. Nice job @matthewmorgan!

Before the last two specs, I would add a least one or two:

```
expect(flattener.flatten([1, 2, null])).toEqual([1, 2]);
```

I also think `undefined` is an important value in JavaScript, so I'd add this one two:

```
expect(flattener.flatten([1, 2, undefined]).toEqual([1, 2]);
```

I think these specs will introduce users smoothly into *empty* values for this exercise. In fact, I'd write them at the beginning of the specs, but it's not a big deal, so I don't care too much.

matthewmorgan:
OK, well, that's 2 votes for adding in `undefined`.  I did that, and also fixed a couple of typos in the test.

BTW I realized I am using `it` and `xit` here instead of Jest's `test` and `xtest`.  To support `test` and `xtest` I think we'll need to alter the Makefile.  I'll open an issue to discuss that.

joelwallis:
I've just noticed that `undefined` is *not* in the [spec](https://github.com/exercism/problem-specifications/blob/master/exercises/flatten-array/description.md). But I still think it worth being included in the specs:

```js
const a = [1, 2, 3]
a[3] = this['numberFour'] // numberFour is undefined

new Flattener().flatten(a) // [1, 2, 3, undefined]
```

The example above doesn't seem be what a Flattener user would expect. Actually, it would be an issue that's very hard to debug, since setting undefined attributes like this does not throw an error.

joelwallis:
Mat already implemented the `undefined` value. So, I think this one is good to merge.

matthewmorgan:
Thanks for the feedback gentlemen!

rchavarria:
Thank you for all the awesome work

PR #313, opened 2017-06-30
Add exercise List-Ops

matthewmorgan:
A few thoughts:

1. I let myself use `pop` and `push`, using arrays to store the data. No Array iterator methods, no `length` property.  I did cheat a little and use the array index for the `foldr` method.  If anyone has any better ideas for this, I'm open.
2. I intentionally didn't mutate the inputs.  If we were handling objects instead of primitives, we might want to copy the objects before adding them to the list.
3. I made `length()` a method rather than a property.  I could be persuaded to change that, but it does mean that the `length` property would have to be updated with every relevant operation, so, it seems nicer to calculate it when needed.
4. Where it made sense, I made the methods chainable.  Some of the tests expect this, but I think that's OK since we're trying to emulate functional iterator methods, so it seems like the right thing to do.

Any feedback would be appreciated!

### DIFF
--- a/config.json
+++ b/config.json
@@ -314,6 +314,15 @@
       ]
     },
     {
+      "slug": "list-ops",
+      "difficulty": 5,
+      "topics": [
+        "Data structures",
+        "Lists",
+        "Loops"
+      ]
+    },
+    {
       "slug": "binary-search-tree",
       "difficulty": 1,
       "topics": [
--- a/exercises/list-ops/example.js
+++ b/exercises/list-ops/example.js
@@ -1,0 +1,77 @@
+class List {
+
+  constructor(arr){
+    this.values = arr || [];
+  }
+
+  append(otherList){
+    for (let el of otherList.values){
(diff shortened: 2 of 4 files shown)


rchavarria:
1. `pop` and `push` are basic methods of `Array`. Without them, I'm not sure you're able to do anything with Arrays. :+1: 
2. :100: for immutable
3. I'm ok with the method. We don't need to imitate exactly the JavaScript API. This is just for learning.
4. Perfect, fluent interfaces are quite readable.

This is a terrible work! Thanks @matthewmorgan! 

In `config.json`, I'd add some other topics: `For loops`, `While loops` or simply `Loops`.

matthewmorgan:
OK @rchavarria, thanks for the feedback.  I've added those, and also changed `test` to `it/xit` in the specfile.

rchavarria:
Oops, I didn't realized about `test` instead of `it`. Nice!

PR #314, opened 2017-06-30
Add exercise Diffie-Hellman

matthewmorgan:
I'm not an expert at crypto, but thought this might be fun.

I tried to keep the example simple by limiting the possible modulo and base to the first 1000 primes, but this isn't a condition of the tests.  In fact, that code could probably be removed entirely.

There was no canonical test data for this one.  I built the tests based on the steps in the README.  I'm open to adding more if that makes sense.

Feedback appreciated!


exercises/diffie-hellman/diffie-hellman.spec.js
```diff
@@ -0,0 +1,67 @@
+import DiffieHellman from './diffie-hellman';
+
+describe('diffie-hellman', () => {
+  const p = 23;
+  const g = 5;
```
rchavarria: I'd give better names to `p` and `g`, but that is the terms used in the `README`, so they're ok for me.

exercises/diffie-hellman/diffie-hellman.spec.js
```diff
@@ -0,0 +1,67 @@
...
+  let diffieHellman = new DiffieHellman(p, g);
```
rchavarria: Could it be also a `const`?

exercises/diffie-hellman/diffie-hellman.spec.js
```diff
@@ -0,0 +1,67 @@
...
+
+  test('throws an error if the constructor arguments are not prime', () => {
+
+    expect(() => {
+      new DiffieHellman(10, 13)
+    }).toThrow();
+
+  });
+
+  test('throws an error if private key >= p-1', () => {
```
rchavarria: `README` says:

> [...] picks a private key, `a`, greater than 1 and less than p.

Let `a` be the given private key, I'd write the specs for these cases:

1. if `a < 0`, throw (negative integers)
2. if `a == 0`, throw
3. if `a == 1`, throw
4. if `a > 1 && a < p`, do not throw. Maybe, this test case is covered in next specs so that it doesn't need a test here.
5. if `a == p`, throw
6. if `a > p`, throw

exercises/diffie-hellman/diffie-hellman.spec.js
```diff
@@ -0,0 +1,67 @@
...
+
+  test('when given a different private key, returns the correct public one', () => {
+
+    expect(diffieHellman.getPublicKeyFromPrivateKey(bobPrivateKey)).toEqual(bobPublicKey);
+
+  });
+
+  test('can generate a shared secret from our private key and their public key', () => {
+
+    const aliceCalculatesSharedSecret = 2;
```
rchavarria: If both Alice and Bob will calculate the same secret, do we need two values for the same secret?

exercises/diffie-hellman/example.js
```diff
@@ -0,0 +1,29 @@
+export default class DiffieHellman {
+
+  constructor(p, g) {
+    if (p < 2 || g < 2 || p > 7919 || g > 7919) {
```
rchavarria: Does `7919` the last prime used in `PRIMES`? Why not use `PRIMES[PRIMES.length - 1]` instead of a hardcoded value?

exercises/diffie-hellman/example.js
```diff
@@ -0,0 +1,29 @@
...
+      throw Error('Constructor arguments must be between 2 and 7920!');
+    }
+    if (!PRIMES.includes(p) && PRIMES.includes(g)) {
```
rchavarria: This and previous `if` statements could be extracted to a function. That function might throw errors, and the constructor will be like:

```
constructor(p, g) {
  validate(p);
  validate(g); // or validate(p, g), whatever
  this.p = p;
  this.g = g;
}
```

exercises/diffie-hellman/diffie-hellman.spec.js
matthewmorgan: I think adding additional tests is a good idea.  But, why do you suggest separate tests for some for the cases, EG (1, 2, 3) could be: `a < 2` , and (5, 6) could be `a < p`?

Can you share your thoughts on what you think the value is in having the extra tests?

exercises/diffie-hellman/diffie-hellman.spec.js
matthewmorgan: No!  I will take that out.

matthewmorgan:
@rchavarria thanks for the feedback.  I'm going to push the changes without squashing them so they're easier to see.

exercises/diffie-hellman/diffie-hellman.spec.js
matthewmorgan: @rchavarria I took a look at the other existing test suites (C#, F#, Elixir, Go) and they all just have one test for the range, similar to the one I wrote.

Would you be OK with just adding two more tests or do you feel that all six should be added?

rchavarria: I usually favor one test for one assert, or just one assert by test. But it's not a golden rule. In this case, it makes sense to check `-1`, `0` and `1` in the same test (and `p` and `p + 1` in another one).

I commented about more than one case, not just `a < 2` because negative numbers, `0`  and `1` have *special* meanings. It's not that you have to test all possible values: `-100`, `-99`,... But `0`, `1` and so on are kind of edge cases, or very common edge cases.

rchavarria:
:cry: there is a conflict in `./exercises/package.json`.

But I think `./exercises/package.json` can be deleted. `./package.json` can't. `./exercises/<each exercise/package.json` can't. But, `./exercises/package.json` can. Don't you think?

exercises/diffie-hellman/diffie-hellman.spec.js
matthewmorgan: @rchavarria I'm convinced.  I'll make those changes 😁

Also, I thought it was interesting, some of the other test suites check big primes (dozens of digits), or have their Diffie-Hellman class generate random private keys and test for randomness.  Seems like those tests are not needed for this sort of exercise.

matthewmorgan:
@rchavarria I've incorporated your feedback.  Please let me know if this looks good to merge!

rchavarria:
This is impressive!! Merging...

PR #316, opened 2017-07-04
Update test methods

### LINKED ISSUE #315: Update tests and makefile to use Jest's "test" method
Jest is capable of using Jasmine's `it` and `xit` methods, but has its own `test` and `xtest`.  I opened this issue to see what @exercism/ecmascripts think about switching from the Jasmine style to the Jest style for this track.  Should we bother, or leave things as they are?

matthewmorgan:
Closes #315.

I've left this as three separate commits for visibility, we of course should squash on merge!

This change looks bigger than it is, pretty much just find-and-replace!


exercises/flatten-array/package.json
```diff
@@ -48,23 +53,71 @@
...
+      "brace-style": [
+        "error",
+        "1tbs",
+        {
+          "allowSingleLine": true
+        }
+      ],
+      "object-curly-spacing": [
+        "off",
+        "always"
```
tejasbubane: Should these changes be in this PR?

matthewmorgan: It looks like the formatting of these three `package.json` files was out-of sync with `exercises/package.json`.  This change brings them in line.  

I didn't intentionally add it, (Makefile copied it over) but I think it needed to be done and can remain here rather than making a separate PR.

tejasbubane: Ok, if it is done by Makefile then thats fine I guess.

matthewmorgan:
@tejasbubane thanks!  @exercism/ecmascripts any other concerns, or are we good to merge?

PR #317, opened 2017-07-07
Chose Airbnb style guide as code standard enforced by ESLint

rchavarria:
#311 doesn't look like a good approach to introduce Airbnb as code standard.

It requires lots of changes on `package.json` and conflicts happen all the time.

Let's change: this PR just configure airbnb style guid as code standard. No rule is disabled. But there are more than 2000 reported errors. Instead of disabling rules (modifying `package.json` all over the place), we will live with ESLint errors. They can be fixed at exercise level. In theory, there won't be more changes to `package.json` files, just in the two or three files in an exercise. If a conflict arise, it will be in just 2 or 3 files. Not in all `package.json` ones.

But this PR should be merged soon to be able to start working on fixing ESLint errors.

### DIFF
--- a/exercises/accumulate/package.json
+++ b/exercises/accumulate/package.json
@@ -10,9 +10,13 @@
   },
   "devDependencies": {
     "babel-jest": "^20.0.1",
-    "babel-preset-env": "^1.4.0",
     "babel-plugin-transform-builtin-extend": "^1.1.2",
+    "babel-preset-env": "^1.4.0",
     "eslint": "^3.19.0",
+    "eslint-config-airbnb": "^15.0.1",
+    "eslint-plugin-import": "^2.2.0",
+    "eslint-plugin-jsx-a11y": "^5.0.1",
+    "eslint-plugin-react": "^7.0.1",
     "jest": "^20.0.1"
   },
   "jest": {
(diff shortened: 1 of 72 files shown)


matthewmorgan:
OK @rchavarria I think we can do it this way.  

NB if you run `eslint --fix .` you get down to 317 errors, which seems manageable to me.  A lot of these are find-and-replace type of issues.

Thanks for doing the investigation!

matthewmorgan:
BTW I found this interesting comment from an ESlint maintainer (who is a local Ann Arbor guy) that suggests to me that we should not be installing ESlint globally, but rather locally for each exercise to avoid some weird issues.  I think we should consider changing our docs for the contributors: https://github.com/eslint/eslint/issues/3782#issuecomment-211366784

joelwallis:
Why do we need `eslint-plugin-react` as dev dependency?

matthewmorgan:
@joelwallis both the `eslint-plugin-react` and `eslint-plugin-jsx-a11y` appear to be peer dependencies of the `eslint-config-airbnb` module.

It's not clear to me why they're needed, but the linting fails without them.

@rchavarria did you have any insight into this?

rchavarria:
Ok to change documentation to suggest uses to install ESLint locally. In fact, there is a npm script in `package.json` to run ESLint. So, there is no need to install it globally.

About `eslint-*` dependencies... Yeah, `eslint-config-airbnb` needed them. It seems the Airbnb style guid is not just for vanilla JavaScript, it covers some aspects of React ans JSX. I think those packages implement some ESLint rules enforced by the style guide.

rchavarria:
@exercism/ecmascripts, would it be ok to merge this PR?

joelwallis:
@rchavarria I think it is. It's opened for five days and we already had some conversation on it. Just merge it then :) let's start working on these 2K warnings over there 😄

matthewmorgan:
Hey guys, I'll merge it-- just wanted to make sure all of @joelwallis's questions were answered!

PR #327, opened 2017-07-27
Added exercise/bowling, requesting code review

AakashMallik:
completed.... standing by for code review

### DIFF
--- a/exercises/bowling/bowling.spec.js
+++ b/exercises/bowling/bowling.spec.js
@@ -1,0 +1,184 @@
+import Bowling from './bowling.js';
+
+describe('Bowling', () => {
+  let game;
+
+  const playGame = (rolls) => {
+    rolls.map(pins => game.roll(pins));
+  };
+
+  beforeEach(() => {
+    game = new Bowling();
+  });
+
+  test('Calculate score for a game with all zeros', () => {
+    const rolls = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+    playGame(rolls);
+    expect(game.score()).toBe(0);
+  });
+
+  xtest('Calculate score for a game with no strikes or spares', () => {
+    const rolls = [3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6];
+    playGame(rolls);
+    expect(game.score()).toBe(90);
+  });
+
+  xtest('A game with a spare followed by zeros is worth ten points', () => {
+    const rolls = [6, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+    playGame(rolls);
+    expect(game.score()).toBe(10);
+  });
+
+  xtest('Points scored after spare are counted twice', () => {
+    const rolls = [6, 4, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+    playGame(rolls);
+    expect(game.score()).toBe(16);
+  });
+
+  xtest('Consicutive spares get one round bonus each', () => {
+    const rolls = [5, 5, 3, 7, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+    playGame(rolls);
+    expect(game.score()).toBe(31);
+  });
+
+  xtest('A spare in the last frame gets a bonus roll that is counted ones', () => {
+    const rolls = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 7, 3, 7];
+    playGame(rolls);
+    expect(game.score()).toBe(17);
+  });
+
+  xtest('A strike earns ten points in a frame with a single roll', () => {
+    const rolls = [10, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+    playGame(rolls);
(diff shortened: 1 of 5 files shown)


matthewmorgan:
@AakashMallik I'll have a look at this for you.  Thanks for doing it.

AakashMallik:
I'll get the code clean up done in a day or two.... Thanks for your time @matthewmorgan

exercises/bowling/bowling.spec.js
```diff
@@ -0,0 +1,467 @@
...
+  });
+
+  xtest('Consicutive spares get one round bonus each', () => {
+    const rolls = [5, 5, 3, 7, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+    playGame(rolls);
+    expect(game.score()).toBe(31);
+  });
+
+  xtest(
+    'A spare in the last frame gets a bonus roll that is counted ones',
```
AakashMallik: Damn these typos..... I laughed so hard at this one....

exercises/bowling/bowling.spec.js
```diff
@@ -0,0 +1,467 @@
...
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        7,
+        3,
+        7,
```
AakashMallik: My Eslint VScode plugin is causing this issue.... i'll fix this em all

matthewmorgan: I'm not sure how we have our ESlint rules set up for this project-- do you think your plugin is doing this because of our rules, or in spite of them?  Maybe we need to address this in our config for the repo.

matthewmorgan:
@AakashMallik thanks!  I want to make sure you know that refactoring the `score()` method is not a requirement-- I just think it would be a good idea.  I always like to have opportunities to improve my code.  @rchavarria often calls out things just like this in my PRs :) so I have plenty of practice fixing them!

exercises/bowling/bowling.spec.js
AakashMallik: nope.... its my linter... i changed the config of my linter...

AakashMallik:
@matthewmorgan I am really sorry, I am caught up in this internship process being held in our college.... I'll get this work as soon as I am free....

Btw is there any internship vacancies that you might know about and that may need a web app developer ( React, Angular )?

matthewmorgan:
@AakashMallik I understand that life is busy!  If you think it will be a while before this work will be done we should close the PR.  But we don't have to do that today.  

I don't know of any internships at the moment-- I don't know where you're looking, but in the US (where I am) most of the companies seem to start looking for interns in January to fill positions that start in May, which is the summer school vacation.

Were you hoping to find local work (if so where) or remote?  LMK and I will keep my ears open.

AakashMallik:
I was looking fr local work...
Large MNCs in India usually go the colleges looking for interns during this
time of the year... our summer break starts frm May as well....
Lemme know if there is some company or startup looking for interns...

I think i'll close the PR in a day or two...

AakashMallik:
@matthewmorgan I think refactoring is not a good idea as I feel that this is the type of code where being imperative will generate a far thinner code as compared to declarative. Declarative code will only thicken the code base. I mean even declarative code has an imperative code deep down the hierarchy. I guess this is that deep level. I am open to suggestions though.

matthewmorgan:
@AakashMallik as i mentioned before, refactoring `score()` is not a requirement, as example code is not required to follow any certain pattern or philosophy, except that it should be clear and easy to maintain.  Having said that, I don't really understand your explanation as to why you think refactoring is not a good idea.

It seems to me that code that is easy to read, especially small functions that do something that is easy to understand at a glance, is always a good goal.  

Just curious what you're thinking here.

AakashMallik:
@matthewmorgan I am sorry if I was too abstract about my idea back there.... actually a few months ago I was working with a startup using react codebase and there came a time where I had to write the implementation code for all the declarative code we had used... My mentor told me to push the implementations as low as possible and stick to the declarative style as much as possible. But what I did was I used declarative code when it was time to implement pushing the implementation further down and hence fattening the code base with unnecessary abstraction. Thats why I was reluctant to refactoring the imperative code I wrote for this example.... 
I hope I conveyed my point of view here....

Btw I see another PR trying to implement the bowling exercise!! Has someone started working on it in parallel??

rchavarria:
@AakashMallik - don't worry about that PR, @matthewmorgan prevented that person to continue working because you were already working on this. You can continue with this one.

About the refactoring, I agree with @matthewmorgan about refactoring the code to make it cleaner. But it doesn't mean to pollute the code with abstractions. In your job, you may find a large code base, so you'll have multiple layers of abstractions and lots of opportunities to move and rewrite code up and down those layers.

In this project, each exercise is quite focused and isolated, so the code you need to keep in mind is very little. So, in this case, *refactoring* doesn't mean doing great changes. Just simplify some conditions, extract some variables or methods, even renaming some variables.

AakashMallik:
@rchavarria Well that makes sense... I understood where I was going wrong..... 
I will get back to refactoring ASAP, I am stuck in some work though so I will take some time if that's okay...

matthewmorgan:
@AakashMallik I agree with @rchavarria.  What I had in mind is that some of the conditionals might get factored out to named methods, like the example I gave above.  Just so the code reads a bit more like a story...does that make sense?

AakashMallik:
Yes @matthewmorgan ..... I see where the refactoring needs to be done now.... Conditions indeed can be refactored to named functions.... I will get to it ASAP. Thanks again for your time.

matthewmorgan:
@AakashMallik thank you for _your_ time :100:

rchavarria:
@AakashMallik How is this going? It would be really bad if, after all the hard work, this PR can't be part of the track. Is there anything we can do to help you to move it forward?

matthewmorgan:
Looks like this is going to be addressed by #355.

@AakashMallik please consider contributing again in the future when you have time.

PR #336, opened 2017-08-04
Run Jest only once on Travis for faster builds

matthewmorgan:
This PR builds on work begun by @tejasbubane , who had the excellent idea that builds should run faster by invoking Jest once to run all tests.  

1. Builds on Travis now run the `test-travis` make task, which speeds up build times.  One potential disadvantage is that tests are not run sequentially, which can make debugging a challenge. 
2. Local build, run with `make test`, will still run the suite for each exercise sequentially, making debugging more straightforward.
3. `.babelrc` has been removed, as it's not needed.  Configuration is now in `package.json`.
4. Some dependencies have been updated.

This PR looks more extensive than it is, because every `package.json` file has been updated.  Really, the changes are in the main `package.json` file and the `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,12 @@
+.DEFAULT_GOAL := test
+
 # assignments
 ASSIGNMENT ?= ""
 IGNOREDIRS := "^(\.git|docs|bin|node_modules|.idea)$$"
 ASSIGNMENTS = $(shell find ./exercises -maxdepth 1 -mindepth 1 -type d | cut -d'/' -f3 | sort | grep -Ev $(IGNOREDIRS))
 
 # output directories
-OUTDIR ?= "exercises/$(ASSIGNMENT)/tmp"
+OUTDIR ?= "tmp_exercises/$(ASSIGNMENT)"
 
 # language specific config (tweakable per language)
 FILEEXT := "js"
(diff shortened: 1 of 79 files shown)


Makefile
```diff
@@ -20,17 +22,26 @@ test-package-files:
...
+
+# To be run as: make test-assignment ASSIGNMENT=hello-world
+test-assignment:
+	$(MAKE) -s copy-assignment
+	@node_modules/.bin/jest $(OUTDIR)
+	@rm -rf $(OUTDIR)
+
+test-travis:
+	@echo "Preparing tests..."
+	@for assignment in $(ASSIGNMENTS); do ASSIGNMENT=$$assignment $(MAKE) -s copy-assignment || exit 1; done
```
rchavarria: Could lines 41 to 43 be replaced by a single line that uses the `test-assignment` task?

```
@for assignment in $(ASSIGNMENTS); do ASSIGNMENT=$$assignment $(MAKE) -s test-assignment || exit 1; done
```

Otherwise, I don't see any usage of `test-assignment`.

Makefile
```diff
@@ -20,17 +22,26 @@ test-package-files:
...
+	@node_modules/.bin/jest --bail $(OUTDIR)
```
rchavarria: This line is quite similar to L36. What's the right command? `--bail` or not `--bail`? (sorry, I don't know what's the purpose of that Jest option)

exercises/accumulate/package.json
```diff
@@ -36,9 +36,10 @@
             "Error"
           ]
         }
-      ]
+      ],
```
rchavarria: 💓 Much nicer having all Babel config in one place

Makefile
matthewmorgan: If we combine them we would lose the ability to run the tests in parallel, so we wouldn't want to do that.

Makefile
matthewmorgan: `--bail` exits the test runner on the first error.  I thought this would be best for the situations in which the contributor was running `test` for all exercises, and in the `test-travis` task.

When someone works on  _single_ exercise (usu a new contribution or refactor) they would want to use `test-assignment`, and perhaps it would be useful to see all tests for that exercise run.

rchavarria: yep, absolutely, I agree

Makefile
rchavarria: ok. I see now the difference between `test` and `test-travis`. One process each exercise individually. The other, copy first every exercise and then run all tests.

matthewmorgan:
@rchavarria thanks for taking the time to review! 😁

PR #344, opened 2017-08-23
Updated tests & example for the Custom Set problem

apapirovski:
Hi there! It didn't seem like anyone was working on it so I took care of https://github.com/exercism/ecmascript/issues/229

I've added all the canonical tests + a few of the tests that were in the JavaScript version that seemed useful. There were some bugs in the existing example so I also fixed those up and re-structured slightly. Let me know if anything needs to be adjusted!

### DIFF
--- a/exercises/custom-set/custom-set.spec.js
+++ b/exercises/custom-set/custom-set.spec.js
@@ -1,115 +1,222 @@
 import CustomSet from './custom-set';
 
 describe('CustomSet', () => {
-  test('can delete elements', () => {
-    const expected = new CustomSet([1, 3]);
-    const actual = new CustomSet([3, 2, 1]).delete(2);
-    expect(actual.eql(expected)).toBe(true);
+  describe('empty: returns true if the set contains no elements', () => {
+    xtest('sets with no elements are empty', () => {
+      const actual = new CustomSet().empty();
+      expect(actual).toBe(true);
+    });
 
-    const expected2 = new CustomSet([1, 2, 3]);
-    const actual2 = new CustomSet([3, 2, 1]).delete(4);
-    expect(actual2.eql(expected2)).toBe(true);
+    xtest('sets with elements are not empty', () => {
+      const actual = new CustomSet([1]).empty();
+      expect(actual).toBe(false);
+    });
   });
 
-  xtest('can check for difference', () => {
-    const expected = new CustomSet([1, 3]);
-    const actual = new CustomSet([3, 2, 1]).difference(new CustomSet([2, 4]));
-    expect(actual.eql(expected)).toBe(true);
-    const expected2 = new CustomSet([1, 2, 3]);
-    const actual2 = new CustomSet([1, 2, 3]).difference(new CustomSet([4]));
-    expect(actual2.eql(expected2)).toBe(true);
+  describe('contains: sets can report if they contain an element', () => {
+    xtest('nothing is contained in an empty set', () => {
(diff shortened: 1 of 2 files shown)


matthewmorgan:
Looks good to me! Any other comments from @exercism/ecmascripts ?

tejasbubane:
Can we have groups of tests for each operation/function like `empty`, `contains`, etc? Just like in the canonical json file: https://github.com/exercism/problem-specifications/blob/master/exercises/custom-set/canonical-data.json

apapirovski:
@tejasbubane Definitely — the JS version didn't so I wasn't sure if it was desired or not. Will update tonight or tomorrow.

apapirovski:
Updated with groupings & descriptions based on the canonical tests.

exercises/custom-set/custom-set.spec.js
```diff
@@ -1,115 +1,257 @@
 import CustomSet from './custom-set';
 
 describe('CustomSet', () => {
-  test('can delete elements', () => {
-    const expected = new CustomSet([1, 3]);
-    const actual = new CustomSet([3, 2, 1]).delete(2);
-    expect(actual.eql(expected)).toBe(true);
+  test('size: knows its size', () => {
```
tejasbubane: `size`, `toList` and `clear` are custom functions which this track has added and not there in the [canonical.json](https://github.com/exercism/problem-specifications/blob/master/exercises/custom-set/canonical-data.json).
Let us try to match the canonical data in terms of sequence and keep these custom ones at the bottom so we get nice progression from easier to more difficult tasks.

apapirovski: My thinking here was that we might want to have them build `size` first because they will use it in, at the very least, empty but potentially also in `eql`.

tejasbubane: RIght. Makes sense 👍

matthewmorgan: I think you two have hit upon an important concern: do we want to force a certain implementation on the user?  Or should we only care about the  public API of the class?  I am more inclined to favor the latter.  

My recollection is that there has been an attempt in Exercism to get away from testing private or helper methods, so I'm leabing toward @tejasbubane 's original request that these tests be removed.  I think the methods can remain in the example, since that's an implementation-specific choice.

apapirovski:
Updated to remove the non-canonical tests as per request above.

exercises/custom-set/example.js
```diff
@@ -1,20 +1,38 @@
...
 
   delete(el) {
     delete this.data[el];
     return this;
   }
 
+  clear() {
+    this.data = {};
+    return this;
+  }
```
tejasbubane: `clear` function can be removed from the implementation since it is not being used anywhere now.

apapirovski:
Should be good to go. Let me know if you want me to squash the commits.

matthewmorgan:
@apapirovski nice work.  Looking great.

We do like to have the commits squashed if you're up for it.

If this looks good to you @tejasbubane I feel like it's ready.

apapirovski:
Commits squashed. Let me know if anything else is needed. Thanks!

apapirovski:
Can also probably close https://github.com/exercism/ecmascript/issues/229 now.

matthewmorgan:
Ah!   Good catch.  For future reference, if you put `Closes #229` in the description body for your PR, I believe it will close automatically when your branch is merged.

PR #345, opened 2017-08-23
Add simple-linked-list exercise, ported from JS

apapirovski:
This is a pretty straightforward port, adjusted for ES. Let me know if any changes are needed. Thanks! 

https://github.com/exercism/ecmascript/issues/333

### DIFF
--- a/exercises/simple-linked-list/example.js
+++ b/exercises/simple-linked-list/example.js
@@ -1,0 +1,111 @@
+export class ElementValueRequiredException extends Error {}
+export class ElementNextNotInstanceException extends Error {}
+
+export class Element {
+  constructor(value, next) {
+    if (value === undefined) {
+      throw new ElementValueRequiredException();
+    }
+
+    if (next !== undefined && !(next instanceof Element)) {
+      throw new ElementNextNotInstanceException();
+    }
+
+    this.value = value;
+    this.next = next;
+  }
+}
+
+export class List {
+  static fromArray(array) {
+    const list = new List();
+    array.forEach(item => list.push(new Element(item)));
+
+    return list;
+  }
+
+  push(value) {
+    const newEl = (value instanceof Element)
+      ? value
+      : new Element(value);
+
+    if (!this.head) {
+      this.head = newEl;
+      return;
+    }
+
+    let lastEl = this.head;
+    while (lastEl.next) {
+      lastEl = lastEl.next;
+    }
+
+    lastEl.next = newEl;
+  }
+
+  unshift(value) {
+    const newEl = (value instanceof Element)
+      ? value
+      : new Element(value);
+
+    newEl.next = this.head;
+    this.head = newEl;
+  }
+
+  shift() {
+    if (!this.head) {
+      return;
+    }
+
+    this.head = this.head.next;
+  }
+
+  pop() {
+    if (!this.head) {
+      return;
+    }
+
+    let penultEl;
+    let lastEl = this.head;
+
+    while (lastEl.next) {
+      penultEl = lastEl;
+      lastEl = lastEl.next;
+    }
+
+    if (!penultEl) {
+      this.head = undefined;
+    } else {
+      penultEl.next = undefined;
+    }
+  }
+
+  reverse() {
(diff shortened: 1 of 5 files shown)


matthewmorgan:
@apapirovski thanks for the contribution!

@joelwallis maybe you want to CR this as the JS version was yours?  If not, I'm happy to do it.

exercises/simple-linked-list/example.js
```diff
@@ -0,0 +1,115 @@
...
+
+    this.head = previous;
+  }
+
+  toArray() {
+    const array = [];
+
+    while (this.head) {
+      array.push(this.head.value);
+      this.shift();
```
tejasbubane: I am not sure of the javascript implementation, but this operation is destructive. Since the head is moving forward, once we do `list.toArray()`, there is no way to get back the linked list - which should not be the case. Correct me if I am wrong.

exercises/simple-linked-list/example.js
```diff
@@ -0,0 +1,115 @@
+export class Element {
+  constructor(value, next) {
+    if (value === undefined) {
+      throw new Error('Value required.');
+    }
+
+    if (next !== undefined && !(next instanceof Element)) {
+      throw new Error('A Element instance as next value is required.');
```
tejasbubane: Let us use custom exceptions - https://github.com/exercism/ecmascript/issues/280#issuecomment-299868621

exercises/simple-linked-list/simple-linked-list.spec.js
```diff
@@ -0,0 +1,142 @@
...
+      const el = new Element(1);
+      ll.unshift(el);
+      expect(ll.head).toBe(el);
+
+      const ll2 = new List();
+      const elOne = new Element(1);
+      const elTwo = new Element(2);
+      ll2.unshift(elOne);
+      ll2.unshift(elTwo);
+      expect(ll2.head).toBe(elTwo);
```
tejasbubane: append - `push`
prepend - `unshift`
these names are confusing IMO. I would like to head what others think on this. The public API should be intuitive especially when there are these many functions. Since there is no canonical data for this exercise, we will have to take a call.

exercises/simple-linked-list/simple-linked-list.spec.js
```diff
@@ -0,0 +1,142 @@
...
+      expect(a.length).toBe(2);
+      expect(a[0]).toBe(1);
+      expect(a[1]).toBe(2);
+    });
+
+    xtest('allows you to reverse a List', () => {
+      const ll = new List();
+      ll.push(new Element(1));
+      ll.push(new Element(2));
+      ll.push(new Element(3));
```
tejasbubane: use `List.fromArray`?

exercises/simple-linked-list/simple-linked-list.spec.js
apapirovski: It uses the same API naming as JS arrays, that seems sensible to me. Anyone doing this exercise will know enough JS/ES to know shift/pop & unshift/push, IMO.

apapirovski:
Ok, the naming issue above is outstanding but everything else has been updated. Thanks!

joelwallis:
LGTM 👍

exercises/simple-linked-list/example.js
joelwallis: You're right - and I should port this patch to the JavaScript solution too! Thanks for catching!

exercises/simple-linked-list/simple-linked-list.spec.js
matthewmorgan: I'm in favor of using the JS names.  It seems like the prevailing paradigm.

rchavarria:
@apapirovski Same as #344. Could you squash the commits into a single one? It seems both PRs are good to go, but we prefer having just a single commit before mergin PRs into the repo.
Thanks

apapirovski:
It's a single commit now. Let me know if anything else is needed.

rchavarria:
Great!! Tested locally and it works!

Thanks for the contribution.

PR #346, opened 2017-08-25
Queen attack: force test failure when exception is not thrown

alexashley:
I noticed that this test will pass as long as the `Queens` constructor is implemented, because the assertion is in the catch. Since it was a quick fix, I didn't make an issue.

### DIFF
--- a/exercises/queen-attack/example.js
+++ b/exercises/queen-attack/example.js
@@ -1,10 +1,34 @@
-const W = 8,
-  H = 8,
-  STARTING = { black: [7, 3], white: [0, 3] };
+const W = 8;
+const H = 8;
+const STARTING = { black: [7, 3], white: [0, 3] };
+
+function samePosition({ white, black }) {
+  return white[0] === black[0] && white[1] === black[1];
+}
+
+function buildRow(cell, colCount) {
+  return Array(...Array(colCount)).map(() => cell);
+}
+
+function concatRows(row, rowCount) {
+  return [...Array.prototype.concat.apply(buildRow(row, rowCount)).join('')];
+}
+
+function constructBoard() {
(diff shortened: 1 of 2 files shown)


alexashley:
I think this is failing because the [default implementation] (https://github.com/exercism/ecmascript/blob/418f76a8c5a2f1b48a24aff0344c5b075022aea6/exercises/queen-attack/example.js#L7) is incorrect. 

```js
> console.log([1] === [1])
> false

alexashley:
I tried to follow the existing conventions with my changes to the example, but it added some lint errors 😞 
I also changed the example to throw a string instead of an `Error` since that's what the test was expecting.

matthewmorgan:
@alexashley thanks for taking this on!  I have a couple of thoughts on this PR:

1. If you're getting linting errors, you should be able to automatically fix them using `eslint --fix .`  If this doesn't work, please let me know.
2. The correct pattern to use when expecting an error is to use the `toThrow()` matcher.  As long as you're changing this code, let's do it right!  The Jest docs have some good examples: https://facebook.github.io/jest/docs/en/expect.html#tothrowerror

Please let me know if you have any questions or if I can help in any way.

alexashley:
@matthewmorgan 
1. 👍 No more lint errors in `queen-attack.spec.js` or `example.js`
2. Fixed!

matthewmorgan:
OK, looks great!  One last thing: we like to have the commits squashed to a single commit for this repo.  We can do that as part of the merge, but usually ask contributors to do it themselves if they're game.

Let me know if you need some tips on this, I'm happy to provide them.

alexashley:
@matthewmorgan No problem!

matthewmorgan:
Woohoo!  Thanks @alexashley !

PR #347, opened 2017-09-01
Add test for simple-cipher wrap on decode

nicholejeannine:
Add test to ensure that simple cipher solution correctly wraps characters while decoding.


matthewmorgan:
Great idea!  I see that this problem has no canonical tests, but this seems like a fine addition.  Thanks for your contribution!

PR #351, opened 2017-09-11
Add exercise word-search

matthewmorgan:
I will add the `word-search` exercise to this track.

I have some concerns about the canonical tests not being comprehensive enough.  I'm going to open an issue to get some feedback on https://github.com/exercism/problem-specifications 

If the changes don't get added there, I'll extend the cases here.

UPDATE: it sounds like there is support on the problem spec repo for adding the cases.  Once that PR is accepted, I will open my PR here.

UPDATE: my PR in `problem-specifications` has been merged, so we now have an official updated canonical test set!  I hope to finish the EcmaScript version this weekend and get it pushed here for code review.

UPDATE: Whew!  This was tougher than I expected!  The PR looks big because there was a curly brace out of place in the `package.json` file, so _every exercise_ is updated.

I would welcome feedback or improvements on `example.js` if anyone has thoughts.  The whole exercise is very "loopy" by nature, would love to improve that if there are suggestions.


matthewmorgan:
@exercism/ecmascripts I've made a few changes that reduce code duplication and (I think) make the PR more readable.  

LMK what you think.  Thanks!

matthewmorgan:
@exercism/ecmascripts sounds like everyone is too busy to do any code review?  Since this implements the canonical tests and passes CI, I'll merge it in the morning unless I hear any objections.

LMK if you see any issues, I'm happy to open another PR to correct them.

joelwallis:
Hey @matthewmorgan,

I didn't see this PR before, so I'm giving it a look now. Also, I just noticed that there are 85 files being changed on this PR, and it seems to me that there is an automatic formatting step being performed by your editor (my best guess). Would you mind double-checking it before going any further?

exercises/package.json
```diff
@@ -0,0 +1,69 @@
+{
```
rchavarria: I think we don't need this file. `package.json` in the root folder is ok. `package.json` on each exercise folder is ok. But this is not in the root folder not in exercise specific folder. I think it could be deleted.

exercises/word-search/README.md
```diff
@@ -0,0 +1,25 @@
+In word search puzzles you get a square of letters and have to find specific
```
rchavarria: The README seems ok, but other READMEs include more information about how to run tests and so on.

I think that if a person is going to solve this exercise, she has been solved lots of exercises in exercism previously, and she'll be able to run tests on her own.

Do you think we should include more information as other READMEs do?

exercises/word-search/example.js
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  find(words) {
+    return words
+      .map(word => ({[word]: findWordInAnyDirection(word, this.grid)}))
+      .reduce((acc, oneWord) => Object.assign(acc, oneWord), {});
+  }
+}
+
+function findWordInAnyDirection(word, grid) {
```
rchavarria: Wow! Function names make this piece of code really readable. That's a relief, because there's lots of code in this file (the exercise doesn't feel easy to implement).

exercises/word-search/example.js
```diff
@@ -0,0 +1,146 @@
...
+      if (startCol) {
+        [end, start] = getCoords();
+      }
+    }
+    if (start && end) {
+      return {start, end};
+    }
+    rowIndex++;
+  }
+}
```
rchavarria: Should this function return something explicitly? Even `return undefined`?

Not sure, just to make it explicit, and probably the `or` operator at `findWordInAnyDirection` would be a bit clearer .

exercises/word-search/example.js
```diff
@@ -0,0 +1,146 @@
...
+  };
+}
+
+function flipGrid(grid) {
+  return [...grid[0]]
+    .map((col, c) => grid.map((row, r) => grid[r][c]))
+    .map(row => row.join(''))
+}
+
+function searchDiagonally({word, grid, isReversed = false, fromTop = true}) {
```
rchavarria: I'm not a big fan of optional arguments. But I don't have a better suggestion.

To search horizontally in both directions, you implemented a `flipGrid` function. Would it be useful to search diagonally? Would there be a similar function to search diagonally by *flipping* the grid? I'm just thinking outloud, I'm not asking for implementing it.

matthewmorgan: @rchavarria I had thought about rotating the grid 45 degrees to do the diagonal search, but I couldn't quite figure out how to make that happen.

If it were possible to do that, then you would only need to search in one direction: just rotate the grid each time, and correctly transform the coordinates back to the original system.

Maybe some enterprising exercism user will figure that one out...?

exercises/word-search/example.js
matthewmorgan: I could certainly add that if you think it's better to have it explicit.  I just suggested to a contributor that he take his out!!  :)

exercises/word-search/README.md
matthewmorgan: Right.. good catch.  I forgot the "new" README format is different...this one just comes from the problem spec repo.

exercises/package.json
matthewmorgan: Good catch-- that's an artifact of a broken 'make' run.  I'll remove it.

matthewmorgan:
@joelwallis @rchavarria thanks for taking a look.  I believe I've addressed all of your feedback, please let me know if not.

@rchavarria If we can decide on the question of explicitly returning undefined or not, I can finish this up.  I don't have a strong opinion: I think an argument could be made that it demonstrates this feature of the language, and a contrary argument could be made that an explicit return is clearer.  I guess the question is which is more valuable to someone who is likely to be reading the example in this repo.

exercises/word-search/example.js
rchavarria: Oh, I see. 45 degrees rotation? Interesting point, but not sure how to reflect that. And, I don't know if the exercise forces the grid to be square. Let's keep it as it is.

exercises/word-search/example.js
rchavarria: On one hand, I would like to see all returning points. On the other hand, returning `undefined` is not very helpful and may confuse the reader.

rchavarria:
This is ok for me. If nobody has anything to say...

matthewmorgan:
Thanks @rchavarria!!

@joelwallis I _could_ merge my own PR, but would be happier if one of my esteemed colleagues would do the honors!  Do you have any remaining concerns?

rchavarria:
We can revert if there are any concerns with this PR. Meanwhile, I'm going to merge it.

Thanks @matthewmorgan !!

joelwallis:
@matthewmorgan I spend some time reading the code but I was tired enough to not understand a thing 😂

I'm glad Rubén merged your PR. I deleted your branch, btw. Great work!

Issue #191, opened 2016-01-29
triangle: incorrect test in some tracks


kytrinyx:
Please check if there's a test that states that a triangle with sides 2, 4, 2 is invalid. The triangle inequality states that for any triangle, the sum of the lengths of any two sides must be greater than or equal to the length of the remaining side. If this doesn't affect this track, go ahead and just close the issue.

matthewmorgan:
Tested and legal in this track.

kytrinyx:
+1 thanks!

PR #352, opened 2017-09-17
$ is default command line prompt

rootulp:


matthewmorgan:
@rootulp `%` is the default for csh and zsh, `$` is the default for bash, and `>` is the default for Windows.

I'm not opposed to his change, but by the same token I don't know what would be gained by it.

rootulp:
@matthewmorgan thanks for the info regarding various shells! My reason for making the change is because I tend to see `$` more frequently in docs and therefore assume that it could more recognizable for new contributors.

matthewmorgan:
You know, I think it might be best to omit the command line prompt altogether.  I see that is what the [Ruby track](https://github.com/exercism/ruby/blob/master/README.md) does.

I see how it could be confusing, but by changing it, I think we are the same risk on confusing a different set of users.  

Would you mind just getting it out entirely?

rootulp:
I checked the [python](https://github.com/exercism/python) track and I'm on board with removing the prompt all together

matthewmorgan:
Thanks @rootulp !

rchavarria:
I'm late on this, but on the JavaScript track we discussed to get rid of `$` documenting the different commands (https://github.com/exercism/javascript/pull/191)

rootulp:
Sorry @rchavarria I should've updated the PR title sooner but in commit 26658f9 I got rid of the `$`

PR #353, opened 2017-09-21
Fix link and remove incorrect version

rootulp:
Node LTS is currently 6.11 but stating a version number here will become outdated so I think recommending the latest LTS is safer

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Exercism exercises in ECMAScript 6
 ## Running the Unit Test Suite
 
 [Node.js](https://nodejs.org) must be installed. Follow these [instructions](http://exercism.io/languages/ecmascript/installation) for installing nodejs.
-We recommend using the LTS version.
+We recommend using the latest LTS version.
 
 Use `npm` to install all required dependencies:
 
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ Exercism exercises in ECMAScript 6
 
 ## Running the Unit Test Suite
 
-[Node.js](https://nodejs.org) must be installed. Follow [these instructions](http://exercism.io/languages/ecmascript/installing) for installing nodejs.
-We recommend using the latest stable version (v7 at the moment).
+[Node.js](https://nodejs.org) must be installed. Follow these [instructions](http://exercism.io/languages/ecmascript/installation) for installing nodejs.
+We recommend using the LTS version.
 
 Use `npm` to install all required dependencies:
 


rchavarria:
You're absolutely right, and thanks for fixing the link!

README.md
```diff
@@ -7,8 +7,8 @@ Exercism exercises in ECMAScript 6
 
 ## Running the Unit Test Suite
 
-[Node.js](https://nodejs.org) must be installed. Follow [these instructions](http://exercism.io/languages/ecmascript/installing) for installing nodejs.
-We recommend using the latest stable version (v7 at the moment).
+[Node.js](https://nodejs.org) must be installed. Follow these [instructions](http://exercism.io/languages/ecmascript/installation) for installing nodejs.
+We recommend using the LTS version.
```
rchavarria: Why not *latest LTS*  instead of just *LTS*? I'm not a native English speaker, so I don't know if just *LTS* refers to the latests one, sorry.

matthewmorgan: We switched to v7 a while back, so I was concerned that moving back wasn't the right way to go.  But `babel-jest` is transpiling everything before running the tests, so perhaps that doesn't matter.  I would feel better if we used the same version in Travis as we recommend to the users, so I recommend changing both or neither.

And I agree with Rubén-- I think "latest LTS" would be the correct language if we do change.  Right now v4 and v6 are both "LTS", with v4 in LTS 'maintenance' status, and v6 in LTS 'active.'

rchavarria: October? I think it's ok to wait a little bit and set v8 for Travis and recommend it to users.

rootulp: Agreed on _latest LTS_ and recommending the same version we use on Travis. 
On Travis we currently use `7.8`. Is the consensus to bump Travis (and this README) to `8.x`? Or wait until Node marks an `8.x` version as active LTS?

matthewmorgan: @rootulp I think @rchavarria is suggesting that we wait until v8 goes to active LTS and change both then.  Do I have that right @rchavarria ?  That seems like a good idea to me, since it looks like it will days sometime between a few days and a few weeks.

If the v8 is 8.5 or above, that would be great-- native `async/await!`

rchavarria: Yep, just wait a little bit to use the same version in both places. Anyway, if the docs say *latest LTS version* it'll be v8 very soon. I think it will work on v6 (current LTS) until then. I would replace `LTS` with `latest LTS` and go with this PR.

rootulp:
Recommended latest LTS. Ready!

tejasbubane:
Looks like we are good to go with this as per the discussion. Thanks @rootulp 👍

PR #356, opened 2017-10-15
Implement exercise collatz-conjecture

alpakhomov:
I think it's ready, please review:)

### DIFF
--- a/config.json
+++ b/config.json
@@ -355,6 +355,21 @@
       ]
     },
     {
+      "uuid": "363b8810-0bd1-c480-b23c-6b5f629975e7452b123",
+      "slug": "collatz-conjecture",
+      "core": false,
+      "unlocked_by": "leap",
+      "difficulty": 3,
+      "topics": [
+        "Control-flow (conditionals)",
+        "Control-flow (loops)",
+        "Recursion",
+        "Exception_handling",
+        "Integers",
+        "Mathematics"
+      ]
+    },
+    {
       "uuid": "4e0e2c30-be33-49b6-b196-213888a93a0c",
       "slug": "clock",
       "core": false,
--- a/exercises/collatz-conjecture/collatz-conjecture.spec.js
+++ b/exercises/collatz-conjecture/collatz-conjecture.spec.js
@@ -1,0 +1,31 @@
+import steps from './collatz-conjecture';
+
+describe('steps()', () => {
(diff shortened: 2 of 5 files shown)


config.json
```diff
@@ -355,6 +355,21 @@
...
+      "difficulty": 3,
+      "topics": [
+        "Control-flow (conditionals)",
+        "Control-flow (loops)",
+        "Recursion",
+        "Exception_handling",
+        "Integers",
+        "Mathematics"
+      ]
+    },
```
alpakhomov: This exercise is somewhat similar to "triangle", so I put it here and as a "triangle" it is unlocked with "leap".

rchavarria: Looks good

exercises/collatz-conjecture/collatz-conjecture.spec.js
```diff
@@ -0,0 +1,31 @@
+import steps from './collatz-conjecture';
```
rchavarria: It seems it follows [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json).

exercises/collatz-conjecture/example.js
```diff
@@ -0,0 +1,14 @@
+export default (n) => {
+  if (n <= 0) {
+    throw new Error('Only positive numbers are allowed');
+  }
+
+  const iter = (num, acc) => {
```
rchavarria: It would be ok to write longer variable names: `number` instead of `num`; `iteration` or `iterate` instead of `iter`.

What would be a better name for `acc`? `count`? `steps`? Any idea?

exercises/collatz-conjecture/example.js
```diff
@@ -0,0 +1,14 @@
+export default (n) => {
+  if (n <= 0) {
+    throw new Error('Only positive numbers are allowed');
+  }
+
+  const iterate = (number, step) => {
```
alpakhomov: Thanks @rchavarria 
I've fixed variable names (now they are `iterable`, `number` and `step`). I think `step` is a good option - readme specifically refers that the result of exercise should be the number of "steps", but I make it singular, so it won't be confused with array. I've also left the parameter for outer function named `n`, again, for consistency with readme.

rchavarria:
Great!! Many, many thanks for the contribution!

PR #359, opened 2017-10-21
Please review 😊; Implement exercise house

laurmurclar:


exercises/house/example.js
```diff
@@ -0,0 +1,63 @@
+class House {
+  constructor() {
+    this.objects = [
```
matthewmorgan: May I suggest moving `this.objects` and `this.actions` outside of the constructor?  They are really more like static properties of the class, rather than properties of each instance.  No reason to create new ones for every object, just as a best practice.    

Also, they are not part of the public API of this class, so if you leave them outside the class, but inside the module, they are not exposed to consumers of the class.

So...

```javascript

const OBJECTS = [
...
];

const ACTIONS = [
...
];

class House {
...
}
```

Then you would later reference these using `OBJECTS` and `ACTIONS` rather than `this.objects` and `this.actions`.

exercises/house/example.js
```diff
@@ -0,0 +1,63 @@
...
+
+  verse(verseNumber) {
+    const lines = [];
+    const totalLines = verseNumber;
+    let itemIndex = verseNumber - 1;
+    for (let lineNumber = 1; lineNumber <= totalLines; lineNumber += 1) {
+      let lineText = '';
+      if (lineNumber === 1) lineText += 'This is';
+      else {
+        lineText += `that ${this.actions[itemIndex]}`;
```
matthewmorgan: +1 for template strings!

exercises/house/house.spec.js
```diff
@@ -0,0 +1,306 @@
+import House from './house';
+
+describe('House', () => {
+  test('verse one - the house that jack built', () => {
+    const house = new House();
```
matthewmorgan: Since you're declaring a `new House()` in every test, you can pull that out into a `beforeEach`like so:

```javascript
let house;

beforeEach(()=> {
  house = new House();
}

test(...)

xtest(...)
```

exercises/house/example.js
laurmurclar: Sounds much more sensible, thanks!

exercises/house/house.spec.js
laurmurclar: Good spot 👍 will update

laurmurclar:
@matthewmorgan ready for another review 😊 I made both class methods static, which removed the need to create new instances in the spec.

rchavarria:
Very nice and clean PR! Thanks @laurmurclar for the hard work, @matthewmorgan for the review.

I'm going to merge this.

PR #362, opened 2017-10-22
Change Exercise is now ECMAScript ready

RobinCsl:


exercises/change/change.spec.js
```diff
@@ -0,0 +1,80 @@
...
+    expect(result).toEqual([1]);
+  });
+
+  xtest('return a single coin', () => {
+    const change = new Change();
+    const result = change.calculate([1, 5, 10, 25, 100], 25);
+    expect(result).toEqual([25]);
+  });
+
+  xtest('multiple coins coin', () => {
```
matthewmorgan: Should this read `multiple coin change`, as in the [canonical file](https://github.com/exercism/problem-specifications/blob/master/exercises/change/canonical-data.json#L18)?

exercises/change/change.spec.js
```diff
@@ -0,0 +1,80 @@
...
+    expect(result).toEqual([4, 4, 4, 5, 5, 5]);
+  });
+
+  xtest('test no coins make 0 change', () => {
+    const change = new Change();
+    const result = change.calculate([1, 5, 10, 21, 25], 0);
+    expect(result).toEqual([]);
+  });
+
+  xtest('test less than the smallest currency represented', () => {
```
matthewmorgan: Should this word `currency` instead be `coin`?

exercises/change/change.spec.js
```diff
@@ -0,0 +1,80 @@
...
+    const change = new Change();
+    const message = 'The total 3 cannot be represented in the given currency.';
+    const test = () => { change.calculate([5, 10], 3); };
+    expect(test).toThrowError(Error, message);
+  });
+
+  xtest('test a large value that the currency cannot represent', () => {
```
matthewmorgan: Should this word `currency` instead be `coin`? (as above)

exercises/change/example.js
```diff
@@ -0,0 +1,139 @@
...
+    return this.coins.length;
+  }
+
+  getSum() {
+    const getSum = (total, num) => total + num;
+    return this.coins.reduce(getSum);
+  }
+}
+
+Change.prototype.calculate = (coinArray, target) => {
```
matthewmorgan: In ES6 we would want to see the `Change` class written something like this:

```javascript

class Change {
  calculate(coinArray, target) {
    ...
  }
}
```

The preferred pattern is like you have done for your `Candidate` class.  The pattern of extending the prototype is an ES5 pattern, so we'll want to use the newer paradigm here.

exercises/change/example.js
```diff
@@ -0,0 +1,139 @@
...
+  // fill the array with 0 to start
+  candidates[target] = 0;
+  candidates.fill(0);
+
+  // save a new candidate to the candidate array
+  const saveCandidate = (candidate) => {
+    const sum = candidate.getSum();
+
+    if (candidate.getSum() <= target &&
+            typeof (candidates[sum]) !== 'number' &&
```
matthewmorgan: I'm wondering if this wouldn't be more readable if the conditional were extracted into a named function instead?

Also, the indenting is a little unusual here.  You could run the linter against your code in "fix mode" to perhaps take care of a lot of these things automatically.  Let me know if I can provide some assistance on how to do that.

exercises/change/example.js
```diff
@@ -0,0 +1,139 @@
...
+      const candidate = new Candidate();
+      candidate.addCoin(temp);
+      saveCandidate(candidate);
+    }
+  };
+
+  // is everthing searched?
+  const isDone = () => {
+    let done = true;
+    for (let i = 0; i < candidates.length; i += 1) {
```
matthewmorgan: Since you don't need the index `i` here, you might consider using the now available `for-of` instead, like so:

```javascript
for (let candidate of candidates){
...
}
```

This would allow you to eliminate the indirection of using the index variable `i` and the temporary variable `temp`.

exercises/change/example.js
```diff
@@ -0,0 +1,139 @@
...
+          break;
+        }
+      }
+    }
+    return done;
+  };
+
+  // get the next unsearched member of the candidate array
+  const getNext = () => {
+    for (let i = 0; i < candidates.length; i += 1) {
```
matthewmorgan: Same remarks about the `for-of` as above.

exercises/change/example.js
```diff
@@ -0,0 +1,139 @@
...
+
+  // for the candidate, generate another candate for each of the possible coins
+  const branch = (current) => {
+    for (let j = 0; j < coinArray.length; j += 1) {
+      // make a new Candidate for coin type
+      const candidate = new Candidate();
+
+      // copy the curent coins into it and add the new coin type
+      for (let i = 0; i < current.getCoins().length; i += 1) {
+        candidate.addCoin(current.getCoins()[i]);
```
matthewmorgan: Again, you have two loops that use index variables that you don't really need.  Hit it with some ES6!!!

RobinCsl:
Thank you very much for all your comments, they were really helpful and constructive. I hope that the code is clearer and better now. Let me know if there is anything else I should improve! :)
(Also, just so you know, I had this error when running `npm run lint` if there were `for...of` loops: 
```
error  iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations  no-restricted-syntax
```
so I did without them, which actually produced more readable code in my opinion.)

exercises/change/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Making the Test Suite Pass
+
+Execute the tests with:
+
+```bash
+$ npm test
+```
+
+In many test suites all but the first test have been skipped.
```
matthewmorgan: Probably should change this to read "In the test suite, all tests but the first have been skipped."  I'm not sure where this language came from, but I know it's present in many exercises.

matthewmorgan:
Awesome!  I'm glad you feel the review was helpful and the code is better for it.

I had one minor language change for the README if you don't mind, otherwise looks ready to merge.

RobinCsl:
Just updated the README file :)

matthewmorgan:
Great!  Mind squashing to one commit?  If you need assistance with this, I'm happy to help.

RobinCsl:
I managed to squash the commits for this one too! I'll try again on #365!

rchavarria:
@matthewmorgan approved the changes, and commits have been squashed. Merging...

PR #364, opened 2017-10-24
Transpose

RobinCsl:
There is a commit comment relating to files which should have been in their own branch, but I worked on master first, and then realised that to make a second PR, I should have used a separate branch... Hopefully that's not a problem!

### DIFF
--- a/exercises/transpose/README.md
+++ b/exercises/transpose/README.md
@@ -1,0 +1,94 @@
+# Transpose
+
+Given an input text output it transposed.
+
+Roughly explained, the transpose of a matrix:
+
+```text
+ABC
+DEF
+```
+
+is given by:
+
+```text
+AD
+BE
+CF
+```
+
+Rows become columns and columns become rows. See <https://en.wikipedia.org/wiki/Transpose>.
+
+If the input has rows of different lengths, this is to be solved as follows:
+
+- Pad to the left with spaces.
+- Don't pad to the right.
+
+Therefore, transposing this matrix:
+
+```text
+ABC
+DE
+```
+
+results in:
+
+```text
+AD
+BE
+C
+```
+
+And transposing:
+
+```text
+AB
+DEF
+```
+
+results in:
+
+```text
+AD
+BE
+ F
+```
+
+In general, all characters from the input should also be present in the transposed output.
+That means that if a column in the input text contains only spaces on its bottom-most row(s),
+the corresponding output row should contain the spaces in its right-most column(s).
+
+## Setup
+
+Go through the setup instructions for ECMAScript to
+install the necessary dependencies:
+
+http://exercism.io/languages/ecmascript
+
+## Requirements
+
+Install assignment dependencies:
+
+```bash
+$ npm install
+```
+
+## Making the Test Suite Pass
+
+Execute the tests with:
+
+```bash
+$ npm test
+```
+
+In the test suite, all tests but the first have been skipped.
+
+Once you get a test passing, you can enable the next one by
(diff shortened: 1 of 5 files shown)


matthewmorgan:
@RobinCsl no worries at all about working on master-- since you're on your own fork, it's not a problem.  I'll review.

exercises/transpose/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Making the Test Suite Pass
+
+Execute the tests with:
+
+```bash
+$ npm test
+```
+
+In many test suites all but the first test have been skipped.
```
matthewmorgan: Same language issue I mentioned on your other PR.  All tests _have_ been skipped except the first, so might as well say so.

exercises/transpose/example.js
```diff
@@ -0,0 +1,12 @@
+export default function transpose(text) {
+  const result = [];
+  text.forEach((line, lineNo) => {
+    line.split('').forEach((value, key) => {
```
matthewmorgan:
```javascript
   [...line].forEach((value, key) => {
```

The `...` is the "spread" operator.  [It's awesome](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator).

exercises/transpose/example.js
```diff
@@ -0,0 +1,12 @@
...
+      if (typeof result[key] === 'undefined') {
+        result[key] = ' '.repeat(lineNo);
+      }
+      result[key] += value;
+    });
+  });
+  return result;
+}
```
matthewmorgan: This all looks good, but it did make me wonder if it's possible to do this entirely functionally-- IE, without having an initial `result` array that gets mutated.  Not critical, but just thinking out loud...

RobinCsl:
I updated the README and used the spread operator (it's really cool indeed!). I tried briefly to think about rewriting the example in a purely functional way, but to no avail at the moment.

matthewmorgan:
Looks fine to me!  Mind squashing the commits?

RobinCsl:
It seems I somehow managed on this one. I'll try to replicate the steps. (But there might be other conflicts since each exercise implementation modifies config.json)

rchavarria:
@matthewmorgan approved changes, commits squased, no conflicts,... merging.

shybyte:
@RobinCsl The expected test result in the last test ("test many lines") is different from the expected test result in https://github.com/exercism/problem-specifications/blob/master/exercises/transpose/canonical-data.json 

So either the canonical test-data is wrong or this PR is wrong.

matthewmorgan:
@shybyte good catch! @RobinCsl I need to remove this exercise for the time being.  Please have a look and let me know if I can help resolving the issue.

RobinCsl:
I had a look for the last hour. I had copy-pasted the code from the old repository and converted it to ES6, but you're right @shybyte, there is an issue with the algorithm. I will work more on that.

shybyte:
Maybe my code for solving the original testcases is helpful:

```javascript
function trimTrailingUndefined(array) {
  const trailingUndefinedCount = [...array].reverse().findIndex(x => x !== undefined);
  return array.slice(0, array.length - trailingUndefinedCount);
}

export default function transpose(input) {
  const maxCol = Math.max(0, ...(input.map(row => row.length)));
  return [...Array(maxCol).keys()].map(col =>
    trimTrailingUndefined(input.map((_v, row) => input[row][col]))
      .map(charOrUndefined => charOrUndefined || ' ')
      .join('')
  );
}
```

PR #365, opened 2017-10-24
Protein translation

RobinCsl:
Same comment as for PR #364, I did some kind of manual rebase so that there shouldn't be any conflicts. Looking forward to having your comments.

### DIFF
--- a/exercises/protein-translation/example.js
+++ b/exercises/protein-translation/example.js
@@ -1,0 +1,45 @@
+const ACID_PROTEIN_MAP = {
+  AUG: 'Methionine',
+  UUU: 'Phenylalanine',
+  UUC: 'Phenylalanine',
+  UUA: 'Leucine',
+  UUG: 'Leucine',
+  UCU: 'Serine',
+  UCC: 'Serine',
+  UCA: 'Serine',
+  UCG: 'Serine',
+  UAU: 'Tyrosine',
+  UAC: 'Tyrosine',
+  UGU: 'Cysteine',
+  UGC: 'Cysteine',
+  UGG: 'Tryptophan',
+  UAA: 'STOP',
+  UAG: 'STOP',
+  UGA: 'STOP',
+};
+
+const getProtein = codon => ACID_PROTEIN_MAP[codon] || 'INVALID';
+
+export default function translate(rnaStrand) {
+  const proteins = [];
+
+  if (rnaStrand) {
+    for (let i = 0; i < rnaStrand.length; i += 3) {
+      const protein = getProtein(rnaStrand.substring(i, i + 3));
+
+      if (protein) {
+        if (protein === 'STOP') {
+          break;
+        }
+
+        if (protein === 'INVALID') {
+          throw new Error('Invalid codon');
+        }
+
+        proteins.push(protein);
+      }
+    }
+  }
+
+  return proteins;
+}
--- a/exercises/protein-translation/protein-translation.spec.js
+++ b/exercises/protein-translation/protein-translation.spec.js
@@ -1,0 +1,59 @@
+import translate from './protein-translation';
+
+describe('ProteinTranslation', () => {
+  test('Empty RNA has no proteins', () => {
+    expect(translate()).toEqual([]);
+  });
+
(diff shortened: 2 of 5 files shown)


exercises/protein-translation/protein-translation.spec.js
```diff
@@ -0,0 +1,59 @@
...
+  xtest('Small RNA strand', () => {
+    expect(translate('AUGUUUUCU')).toEqual(['Methionine', 'Phenylalanine', 'Serine']);
+  });
+
+  xtest('Stop codon ends translation', () => {
+    expect(translate('AUGUUUUCUUAAAUG')).toEqual(['Methionine', 'Phenylalanine', 'Serine']);
+  });
+
+  xtest('Invalid codon throws error', () => {
+    expect(() => { translate('LOL'); }).toThrow(new Error('Invalid codon'));
```
matthewmorgan: We could clean this up a little like so:
```javascript
expect(() => translate('LOL')).toThrow(new Error('Invalid codon'));
```

exercises/protein-translation/example.js
```diff
@@ -0,0 +1,63 @@
...
+
+    case 'UAA':
+    case 'UAG':
+    case 'UGA':
+      return 'STOP';
+
+    default:
+      return 'INVALID';
+  }
+};
```
matthewmorgan: Nothing at all wrong with the massive `switch` statement, but I sometimes like an alternative approach to these sorts of things, using an object.  EG:

```javascript
const ACID_PROTEIN_MAP = {
  'AUG': 'Methionine',
  'UUU': 'Phenylalanine',
  'UUC': 'Phenylalanine',
// and so on...
};

const getProtein = codon => ACID_PROTEIN_MAP[codon] || 'INVALID';
```

It remains very readable and gets what's essentially configuration out of your logic.  In production, you might pull the config out even further: this makes it easy to adjust the mapping without having to redeploy code.

RobinCsl:
Implemented the changes proposed. I tried to extract the config in an external json file, but it seemed to cause more complications, so I finally opted for the easier solution of setting this variable in the same file.

exercises/protein-translation/acid-protein-map.config.json
```diff
@@ -0,0 +1,19 @@
...
+  "UCG": "Serine",
+  "UAU": "Tyrosine",
+  "UAC": "Tyrosine",
+  "UGU": "Cysteine",
+  "UGC": "Cysteine",
+  "UGG": "Tryptophan",
+  "UAA": "STOP",
+  "UAG": "STOP",
+  "UGA": "STOP"
+}
```
matthewmorgan: Gorgeous!  Awesome indeed.

matthewmorgan:
@RobinCsl I see why you're having a problem-- I think the makefile that runs the test suite does not copy your config file over...not a problem to include it in the example file.

It's nice, though, right?  The config is ~ 1/2 the size of the `switch` statement and just as clear.

This PR looks good to me, if you could please squash the commits we'll merge it.

rchavarria:
Extract configuration to external files is usually a good idea in the real world. In exercism we try to get just three files: readme, tests and production/solution. Keeping all tests and all solution code in one file each make solution management much easy.

I think that could be the reason why the `Makefile` doesn't take care of other files. It *only* knows about the files described above. 

Sorry @RobinCsl for causing you some struggles with that. And **thanks** for implementing this exercise for the track.

RobinCsl:
I tried to squash all the commits together, but every time I push there would still be a conflict with master, even though I rebased off upstream/master. So I believe I will need assistance with squashing my commits. Sorry!

RobinCsl:
Looks like I managed now, I guess I had not rebased in the right way.

rchavarria:
Sorry :disappointed: I think merging PRs #362 and #364 created conflicts in this one.

RobinCsl:
No problem, should be fine now :) Thank you!

matthewmorgan:
@RobinCsl glad to hear you got it sorted.  I did not see you were having trouble, hope it wasn't too frustrating.  Sometimes it can be a bit tricky to resolve the conflicts.  

Thanks for the contribution!

PR #366, opened 2017-10-27
Implement exercise isbn-verifier

MattH-be:


exercises/isbn-verifier/example.js
```diff
@@ -0,0 +1,17 @@
+export default class ISBN {
+
+  constructor(isbn) {
+    this.isbn = isbn.replace(/-/g, '');
+  }
+
+  isValid() {
+    if (!this.isbn.match(/^(\d{9}[\dxX])$/)) return false;
+
+    const digits = this.isbn.split('');
```
matthewmorgan: Perhaps use the ES6 spread operator:

```javascript
const digits = [...this.isbn];
```

exercises/isbn-verifier/isbn-verifier.spec.js
```diff
@@ -0,0 +1,81 @@
+import ISBN from './isbn-verifier.js';
+
+describe('ISBN Verifier Test Suite', () => {
+  test('valid isbn number', () => {
+    const isbn = new ISBN('3-598-21508-8');
+
+    expect(isbn.isValid()).toBeTruthy();
```
matthewmorgan: Why '.toBeTruthy()' instead of `.toEqual(true)`?  Did you want to leave the tests open to someone returning something other than a boolean `true`?  Just curious...

exercises/isbn-verifier/example.js
MattH-be: Changed in a commit. Using the spread operator is indeed the correct syntax in this instance.

exercises/isbn-verifier/isbn-verifier.spec.js
MattH-be: You're correct. In this case we only want to get true or false, so I've changed it to all use `.toEqual`

matthewmorgan:
Nice job-- all the canonical tests are covered and the PR otherwise looks great.

One last thing-- would you mind squashing your commits to one?  If you need any hep with this, please let me know, I'd be happy to help.

MattH-be:
That should've done it.

matthewmorgan:
Nice work!  Thanks again 🥇

PR #367, opened 2017-10-27
Implement exercise complex-numbers

MattH-be:


matthewmorgan:
@MattH-be looks like you have a merge conflict in `config.json`, probably because your last PR added text at the same spot in the file.  Would you mind resolving that and pushing up the change?

exercises/complex-numbers/complex-numbers.spec.js
```diff
@@ -0,0 +1,236 @@
+import ComplexNumber from './complex-numbers.js';
+
+xdescribe('Imaginary unit', () => {
```
matthewmorgan: We probably should make the `describe` nesting a little less complicated here.  Here's what I'm thinking:

1. I think it's nice how test blocks are grouped conceptually. But, we don't have any other examples in the entire track with nested `describe` blocks.  Nothing wrong with showing people the possibilities, but that leads me to thought (2):

2. Sometimes we're using `xdescribe` to skip an entire block, but once the block is un-skipped, we have multiple unskipped tests, and possibly multiple failures for the user.  For this reason, I think it would be better to use `xtest`, even inside the `xdescribe` blocks.  Let them proceed by passing one test at a time, instead of an entire block.

rchavarria: Agree. All `test` blocks but the first one should be skipped with `xtest`, even if they are nested in a `describe` block.

About nesting `describe` blocks, the philosophy of the track is to keep it as simple as possible. By default, there shouldn't be any nested `describe`. Exercism users just have to make individual `test`s blocks one by one. I think nesting `describe`s adds noise to that.

Anyway, we can discuss if there is a reason to nest `describe`s.

MattH-be: I grouped the tests in `decribe` blocks as they where grouped in the `canonical-data.json` file in the problem specifications. I however agree that in order to keep things simple for the users, we should only use one `describe` block and just skip tests by using `xtest`.

MattH-be:
All requested changes should've been implemented. Let me know if you need anything else changed!

exercises/complex-numbers/complex-numbers.spec.js
matthewmorgan: @MattH-be @rchavarria I don't have a problem with the nested `describe`s, but I think the suite looks better now, easier to understand.  👍

matthewmorgan:
@MattH-be Looks like we have some merge conflicts still.  Perhaps another PR was merged in and caused that.  

Can you please:
1. Resolve the merge conflicts?
2. Squash to one commit?

Thanks!

exercises/complex-numbers/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "devDependencies": {
+    "babel-jest": "^20.0.3",
+    "babel-plugin-transform-builtin-extend": "^1.1.2",
+    "babel-preset-env": "^1.4.0",
+    "eslint": "^3.19.0",
+    "eslint-config-airbnb": "^15.0.1",
+    "eslint-plugin-import": "^2.2.0",
+    "eslint-plugin-jsx-a11y": "^5.0.1",
+    "eslint-plugin-react": "^7.0.1",
+    "jest": "^20.0.4"
```
tejasbubane: Some dependencies (namely `jest` and `babel-jest`) were recently upgraded on all exercises here: https://github.com/exercism/ecmascript/pull/368 Can you please update them in this file as well?

MattH-be:
Im having some issues with squashing and not running into merge conflicts.

matthewmorgan:
@MattH-be from what I can see, the merge conflict is happening because the ISBN verifier and complex-numbers blocks in the config.json are adjacent.

It's a little complicated to explain how to do this with a fork branch.  There may be more than one way, but I think this approach would work:

```bash
git remote add upstream https://github.com/exercism/ecmascript.git
git checkout master
git pull upstream master   //this should bring in any changes in the original repo master branch
git checkout complex-numbers
git rebase master

/*
I found merge conflicts in *config.json* because the isbn-verifier config block and the complex-numbers block were adjacent.  I found that all I had to do was remove the *===* etc boundaries around the conflicts and save the file, then...
*/

git add .
git rebase --continue
```

At this point you should be able to force-push to your branch on GitHub:
```
git push -f
```

This is a pretty big hassle but also worth knowing how to do.  Care to give it a try?  If it doesn't work out, I should be able to squash it for you-- lmk.

MattH-be:
That fixed it. I was always doing `git rebase -i HEAD~number of commits`. Glad to know there's a better way of doing it. Was kinda wondering if there was, because it was a really devious way of doing things.

matthewmorgan:
Awesome!  Glad that worked out for you.  Sometimes git makes my head spin...

PR #368, opened 2017-10-29
[Housekeeping] Upgrade jest

tejasbubane:
In the process replace `yarn.lock` with `package-lock.json`. In favour of using `npm` instead of additional dependency of `yarn`.


rchavarria:
Great!!

PR #372, opened 2017-10-30
Fix document skipping of tests - implement exercise

gstvribs:
please, tell me if is right


tejasbubane:
Ohh no! That is not what I meant, sorry if I was not clear with the description.

We need those tests to be skipped. All but the first test are skipped. The idea is for the user solving the exercise to go step-by-step unskipping each test and making it pass.
We just want to tell users how to skip them when they solve the exercise.

Your task is to put that info in the testing doc here: https://github.com/exercism/ecmascript/blob/master/docs/TESTS.md

You can refer the wording here: https://github.com/exercism/ruby/pull/755/files This one is for ruby exercises though, make necessary changes for ecmascript instead.

PR #373, opened 2017-10-30
Implement exercise spiral-matrix

MattH-be:


exercises/spiral-matrix/spiral-matrix.spec.js
```diff
@@ -0,0 +1,55 @@
+import SpiralMatrix from './spiral-matrix';
```
rchavarria: Very clear formatting. All different matrix are clearly readable.

exercises/spiral-matrix/example.js
```diff
@@ -0,0 +1,27 @@
+class SpiralMatrix {
+
+  static ofSize(size) {
+    const spiral = [];
+    for (; spiral.length < size; spiral.push([]));
```
rchavarria: Take a look at [`Array.prototype.fill`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill). I think there should be a more clear way of creating a matrix, kind of `new Array(n).fill(new Array(m)` or so.

exercises/spiral-matrix/example.js
```diff
@@ -0,0 +1,27 @@
...
+
+    const totalNumbers = size * size;
```
matthewmorgan: If you like, you can use the new exponentiation operator:

```javascript
const totalNumbers = size**2;
``` 

Node 7 should support it.

MattH-be:
Just pushed the small changes you guys where asking for.

PR #375, opened 2017-10-31
Implement exercise rectangles

MattH-be:


exercises/rectangles/rectangles.spec.js
```diff
@@ -0,0 +1,145 @@
...
+      '| | |',
+      '+-+-+',
+    ]);
+
+    expect(actual).toEqual(expected);
+  });
+
+  xtest('rectangle of height 1 is counted', () => {
+    const expected = 1;
+    const actual = Rectangles.count(['+--+',
```
rchavarria: Should this line be splitted as:

```
const actual = Rectangles.coung([
   '+--+',
```
to be similar to other specs?

exercises/rectangles/rectangles.spec.js
```diff
@@ -0,0 +1,145 @@
+import Rectangles from './rectangles.js';
```
rchavarria: :+1: for following canonical specs.

exercises/rectangles/example.js
```diff
@@ -0,0 +1,38 @@
+class Rectangles {
+
+  static count(diagram) {
+    const rows = diagram.length;
+    const cols = rows !== 0 ? diagram[0].length : 0;
```
matthewmorgan: This could be simply:
```javascript
const cols = rows ? diagram[0].length : 0;
```

If you're concerned about `rows` being a falsy value other than zero, I think a validation check would be appropriate.

MattH-be:
Requested changes where implemented!

rchavarria:
Hmmm, I think recently merged PRs conflict with this one 😞 

@MattH-be - would you mind fixing them? Can we help you with the task?

matthewmorgan:
Thanks again!  Nice job.

PR #381, opened 2017-11-23
Add exercise React

matthewmorgan:
I'm sure there are other ways to do this, but what do people think?  I believe I have covered all the canonical cases.

NB Just realized I needed to `xtest` and remove some debugging code...should be set now.

### DIFF
--- a/config.json
+++ b/config.json
@@ -511,6 +511,18 @@
       ]
     },
     {
+      "uuid": "303c6969-9446-41aa-871a-11223a43e810",
+      "slug": "react",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 8,
+      "topics": [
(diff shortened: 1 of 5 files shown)


matthewmorgan:
Good catch @rchavarria! I have added that file and rebased my branch on to master.

I agree the tests are complex, but the complexity is mostly in setting up the input and compute cells.  At first I experimented with an API that had a manager class, but in order to perform the tests as specified in the canonical data, you still needed to have access to each cell, so the manager served no purpose and didn't make things simpler.

Perhaps one of the users will come up with a simpler approach!

rchavarria:
I'm going to merge this. The code is there just in case anybody at @exercism/ecmascripts have something to say.

Good job mate!!

Issue #377, opened 2017-11-01
Recommended nodejs versions


tejasbubane:
[Nodejs 9.0 was released](https://nodejs.org/en/blog/release/v9.0.0/) yesterday and [8.9 is now LTS](https://nodejs.org/en/blog/release/v8.9.0/).

I am not sure if any discussion on this happened before, I could not find any, hence the issue.

Since we [recommend LTS version in the javascript track](http://exercism.io/languages/javascript/installation), should we recommend the latest version (9.0 as of now) in this track? And use these same recommended versions in CI?

rchavarria:
How stable is considered v9.0? How often does it get updated? Sorry, I'm a bit lazy, and I don't want to update the recommended version every now and then. Maybe, recommending the latest LTS version is enough too in this track.

Anyway, before recommending users to use a version, we should get that version working on the track. I think there won't be any major issue. Just to be sure users won't be the first to face them.

tejasbubane:
@rchavarria Both 9.0 and 8.9 should be pretty stable in my opinion. I think node version does not matter much especially for this track since we are using babel with env preset - which includes pretty much all latest ES features. 

That being said this was a meta issue for both tracks and my concern was mainly with the relation between recommended version and the CI. Although versions don't matter much, I feel both should be same - whatever they are - it is always better to have some convention. 

Currently javascript track recommends latest LTS but CI uses latest node version (which would be 9 - now or soon - depending upon travis support).
This track does not recommend anything. 

Hence to have one convention in the installations instructions, we can mention either:

1. LTS for javascript track and latest for ecmascript track.
2. LTS in both tracks

and use the same versions in CI.

For changing versions - you don't need to do anything on travis since `node` means latest version and `lts/*` stands for latest LTS release - and they get updated themselves.

matthewmorgan:
We had recently discussed this on a different PR with @KamilMalek.  Unfortunately I can't find it, and I'm not sure if it was this track or the JS track.

I'd be in favor of using LTS on both tracks at this point.  I don't see anything (at this point) in Node v9 that seems relevant to the exercises, but correct me if I'm wrong.

Is linting on the JS track enough to prevent users from implementing ES6 or later style solutions on that track?  If the goal is to provide an environment that is ES5 compatible, and nudges people toward ES5-style solutions, maybe we need to be linting as part of the test process, or maybe we need to regress it to an older, ES5-only version of Node. 

[...]

PR #382, opened 2017-11-25
Update node spec to latest stable release

matthewmorgan:
Per https://github.com/exercism/javascript/issues/462

Issue #377

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-- '7.8'
+- 'node'
 sudo: false
 script:
 - bin/fetch-configlet


tejasbubane:
Should we also recommend it in the [installation doc](https://github.com/exercism/ecmascript/blob/master/docs/INSTALLATION.md) just like we have for the javascript track?

rchavarria:
In JavaScript we recommend:

> [...] Choose the LTS (long term support) version. [...]

I'm not sure it useful to recommend the latest version. If I'm new, and I'm going to install a software, I'll choose the latest version. Would it be true for all users?

tejasbubane:
> I'm not sure it useful to recommend the latest version. If I'm new, and I'm going to install a software, I'll choose the latest version. Would it be true for all users?

Yeah I guess so, makes sense 👍

PR #383, opened 2017-11-26
Update makefile to check package.json on CI

### LINKED ISSUE #303: Fix Makefile "test-package-files" task
We made a well-intentioned change to our Makefile that tests to see if each exercise has a `package.json` file that matches the global `xecmascript/package.json` file, and fails the build if not.

However, there is no easy way for a contributor to update all the `package.json` files when there is a change.  It is critical that these files are all updated, and we want this to happen before PRs are submitted, so the build fails if it is not done.  

Two possible solutions:
1. Have the Makefile check its environment.  If `Travis=true` then we're in CI and should always SKIP copying the `package.json` files from global to each exercise.
2. Add a task to the Makefile that would let contributors copy the package.json file to all exercises, EG `make copy-package-to-exercises`.  

Either approach probably warrants a small update to the contributor docs.

Success Criteria:
- Builds fail on Travis if `package.json` files are not the same in each exercise as the global.
- Contributors have a simple way to make sure their PRs are ready for CI testing.
- Documentation is updated.

matthewmorgan:
Closes #303 

Updates Makefile for CI, so that all exercise `package.json` files are checked against the main `package.json` file.  If they are not identical, the build fails.

NB: since the `copy-assignment` task _already_ copies the main `package.json` to a given exercise, and the contributors are meant to run `make test`, all files _should_ be updated before submission, so I did not make a separate task just to copy the `package.json` files as was discussed in #303.

Bonus: Looks like we had many exercises that did not have matching `package.json` files.  This caused the build to fail at first, so I guess the script is working...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,11 +17,6 @@ TSTFILE := "$(subst _,-,$(ASSIGNMENT)).spec.$(FILEEXT)"
 SOURCE_PKG_MD5 ?= "`./bin/md5-hash ./package.json`"
 PKG_FILES= $(shell find ./exercises/*/* -maxdepth 1 -name package.json)
 
-test-package-files:
-	@for pkg in $(PKG_FILES); do \
-		! ./bin/md5-hash $$pkg | grep -qv $(SOURCE_PKG_MD5) || { echo "$$pkg does not match main package.json"; exit 1; }; \
-	done
-
 copy-assignment:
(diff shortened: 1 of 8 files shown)


rchavarria:
:+1: to get some `package.json` updated. Maybe, they were out of date because the exercises were created just before some versions get updated.

Yeah, the tool seems to work and seems to be useful. :tada:

PR #389, opened 2018-02-01
Make babel target current version of nodejs

tejasbubane:
As [suggested by Mathew](https://github.com/exercism/javascript/issues/462#issuecomment-347548041).

Once this is merged, we can close https://github.com/exercism/javascript/issues/462.

PR #390, opened 2018-02-03
Update package.json files and travis.yml for npm install --no-save

matthewmorgan:
~~@tejasbubane looks like the issue with the build in #389 is around some of the exercises having arrow functions where they should have constructors.  _Why_ this is now a problem, but wasn't before, is a mystery to me.~~

~~This PR is purely to demonstrate a fix for #389.  Please make changes in #389 and we can delete this branch after.~~

~~@exercism/ecmascripts if anybody has any insight into why code that was previously passing linting and testing is now failing, please explain!  I'm guessing it's a change in some underlying dependency, but I don't see what.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,6 @@ script:
 - bin/fetch-configlet
 - bin/configlet lint .
 - make test-travis
+install:
+- npm install --no-save
 
--- a/exercises/accumulate/package.json
+++ b/exercises/accumulate/package.json
@@ -26,7 +26,13 @@
   },
   "babel": {
     "presets": [
-      "env"
+      [
+        "env", {
+          "targets": {
+            "node": "current"
+          }
+        }
+      ]
     ],
     "plugins": [
       [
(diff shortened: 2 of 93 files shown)


matthewmorgan:
@tejasbubane @tarunvelli I've just made an additional commit to the PR to determine if using `npm install --no-save` would work, an it sppears that it works as expected-- it does not alter the main `package.json`, so everything works as expected.

I'm going to open a separate issue to discuss if this change makes sense.

.travis.yml
```diff
@@ -7,4 +7,6 @@ script:
 - bin/fetch-configlet
 - bin/configlet lint .
 - make test-travis
+install:
+- npm install --no-save
```
tejasbubane: Just one nitpick, but would you mind moving the `install` block before `script`? I find it more appropriate in order:
* language
* node version
* install
* script

tejasbubane:
Also can you change the PR title to reflect the changes? (since there are no constructor changes now I guess).

.travis.yml
matthewmorgan: Oops...I missed this comment.  I will move it, no problem at all.

PR #396, opened 2018-02-18
babel presets change - to compare to PR #389 & #390

tarunvelli:


matthewmorgan:
@tarunvelli I really appreciate you taking a look at this!

I don't know what's happening-- the main `package.json` file in my PR and the PR of @tejasbubane _does not_ differ from the `package.json` files in the exercises.  These files are identical to one another, and different than the ones in yours only in their formatting:

Ours:
```json
"presets": [["env", { "node": "current"}]],
```

Yours:
```json
    "presets": [
      [
        "env",
        {
          "node": "current"
        }
      ]
    ],
```

You can see, the only difference is formatting, not that the main file differs from the exercise files.  The error message is...erroneous.

So, I have tried removing one little whitespace, re-running `make test`, and committing the result.  However, the build still fails.  

One can see that there _is no_ difference between the main file and the supposedly failing `accumulate/package.json` by copying the main file to the exercise directory and running `git status`-- no differences!

The next thing I'm going to try is to reformat as yours is formatted.  Maybe it will make the linter happy.

matthewmorgan:
Interesting.  I just changed the format to the multi-line version you had instead of the single-line version Tejas and I used, and the CI build passed!! 🎆 🍾 

All I can think is that the single line version somehow is not correctly hashing when the comparison is made, and fails that test.

Anyway, thanks for poking us on this one!

tarunvelli:
Yes initially your main package.json and the ones in exercises are the same. But try running npm install. Upon running npm install, the main package.json file gets linted because of the script in pacakge.json,
From(yours):

"presets": [["env", { "node": "current"}]],

To(mine):

    "presets": [
      [
        "env",
        {
          "node": "current"
        }
      ]
    ],

matthewmorgan:
I think I understand now.  Forgive my nit-picking your help, but as a maintainer of the track (and a curious person) it's important that I understand _exactly_ why this is failing so I can help others in the future.

The problem was not that a commit was needed after `make test`-- I did that.  The problem was that new versions of `npm` can alter the `package.json` file on `install` as described here: https://github.com/npm/npm/issues/17010.  

It appears that to prevent this behavior (and there's no reason we would want it) we need to explicitly call 

```bash
npm install --no-save
```

Thank you again @tarunvelli.  This was very puzzling indeed.

tejasbubane:
@tarunvelli Thanks a lot for looking into this! 👍 It was indeed quite confusing.

PR #398, opened 2018-02-28
Update README.md

Eitan-G:
The test suite specifies that dashes and spaces should be allowed to be repeated, while the readme omits this.

### DIFF
--- a/exercises/isogram/README.md
+++ b/exercises/isogram/README.md
@@ -2,13 +2,14 @@
 
 Determine if a word or phrase is an isogram.
 
-An isogram (also known as a "nonpattern word") is a word or phrase without a repeating letter.
+An isogram (also known as a "nonpattern word") is a word or phrase without a repeating letter, however spaces and hyphens are allowed to appear multiple times.
 
 Examples of isograms:
 
 - lumberjacks
 - background
 - downstream
+- six-year-old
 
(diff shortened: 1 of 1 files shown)


matthewmorgan:
@Eitan-G thank you for your contribution!

I see that this PR would bring the track in line with the change in the problem-specifications repo.  That's great.

I think there are a couple of issues here:
1. The test suite itself does not contain any cases regarding repeated spaces, yet the README would be updated to allow multiple spaces in a phrase.
2. The test suite does not match the cases in `canonical-data.json` from the [problem-specifications](https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json) repo.

I'm feeling like your PR should _also_ contain an additional test to support the presence of spaces, and that the other tests should match the tests in `canonical-data.json`.  Would you be up for making that change here as well?

Additionally, I think it would be good to make a PR to add your new test as a canonical case in https://github.com/exercism/problem-specifications, but that would be a separate PR to that repo.

Would you be up for that?

matthewmorgan:
@Eitan-G do you have any comments or questions about about feedback above?

matthewmorgan:
@Eitan-G I'm going to close this as we haven't heard from you.  Please feel free to re-open if you make further progress.

PR #399, opened 2018-03-05
Add armstrong-numbers exercise

PakkuDon:
[Problem specification and test data](https://github.com/exercism/problem-specifications/tree/master/exercises/armstrong-numbers)

Ported from exercism/javascript#522

### DIFF
--- a/exercises/armstrong-numbers/example.js
+++ b/exercises/armstrong-numbers/example.js
@@ -1,0 +1,11 @@
+const validate = input => {
+  const digits = [...String(input)];
+  const sum = digits.reduce((total, current) => (
+    total + Math.pow(current, digits.length)
+  ), 0);
+  return sum === input;
+}
+
+export default {
+  validate,
+}
--- a/config.json
+++ b/config.json
@@ -1264,6 +1264,17 @@
         "Arrays",
         "Text formatting"
       ]
+    },
+    {
+      "uuid": "b28c34f4-f7af-47db-95c6-f920e020bbba",
+      "slug": "armstrong-numbers",
+      "core": false,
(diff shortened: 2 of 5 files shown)


exercises/armstrong-numbers/example.js
```diff
@@ -0,0 +1,11 @@
+const validate = input => {
+  const digits = (input + '').split('');
```
matthewmorgan: You want to convert `input` from a `Number` to a `String`, right?  It would be nice to have this be more semantic.  We would also prefer the use of the spread operator since this is an ES6-focused track.  That sounds a little contradictory, eh?  But that's the style we would prefer on this track:

```javascript
const digits = [...String(input)];
```

PakkuDon: That's much cleaner. Thanks for the suggestion 😄 
Commit amended

PakkuDon:
@matthewmorgan That's a good point. It's a small algorithm but the tests seem a bit repetitive - maybe there could be a test case for zero or a negative number?
What should we do?

matthewmorgan:
@PakkuDon I think this is OK for now, but if you have any interest in making a PR to https://github.com/exercism/problem-specifications , I think the test cases could use beefing up.  My feeling generally is that testing `10` by itself is not nearly complete coverage for the assertion "There are no 2 digit Armstrong numbers".  It's that sort of thing that could use improvement.

If it were added to problem-specifications, then it would be visible to all the tracks, not just this one.

At any rate, this PR is much appreciated, and if you're feeling ambitious, please think about adding more tests to problem-specifications!  I would be happy to help review that.

PR #403, opened 2018-04-10
Simple Cipher: Add tests for random key generation, 100+ character length

daveyarwood:
_NB: I've made this same PR in the JavaScript track (https://github.com/exercism/javascript/pull/538) and TypeScript track (https://github.com/exercism/typescript/pull/186) and I'm making the rounds. I'd love to see this consistent across all the JS tracks!_

The README for this exercise describes generating random keys consisting of 100+ lowercase letters, but the only constraint enforced by the tests is that it must contain only lowercase letters. One could pass this constraint by simply defaulting the key to a hard-coded string like `lolnotrandom`.

This PR adds tests to enforce that keys are randomly generated (at least a simple test that if you generate 2 keys back to back, they are not the same key), and that they are at least 100 characters long.

### DIFF
--- a/exercises/simple-cipher/simple-cipher.spec.js
+++ b/exercises/simple-cipher/simple-cipher.spec.js
@@ -1,10 +1,24 @@
 import Cipher from './simple-cipher';
 
+describe('Random key generation', () => {
+  xtest('generates keys at random', () => {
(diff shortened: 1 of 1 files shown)


matthewmorgan:
Awesome!  Thanks again @daveyarwood

daveyarwood:
:beers:

PR #405, opened 2018-04-25
Use npm-ci instead of npm-install

tejasbubane:
Looks like `npm-ci` is meant to be used in CI environments.
https://docs.npmjs.com/cli/ci

Also we don't need the `--no-save` flag with this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,10 @@ language: node_js
 node_js:
 - 'node'
 sudo: false
+before_install:
+- npm i -g npm@latest
 install:
-- npm install --no-save
+- npm ci
 script:
 - bin/fetch-configlet
 - bin/configlet lint .


matthewmorgan:
@tejasbubane not sure why this is failing.  I will try to have a look when I can.  Thanks!

tejasbubane:
Looks like travis [decided not to install custom npm](https://github.com/travis-ci/travis-ci/issues/4653#issuecomment-169035161) so we will have to install it ourselves (which is what I have done in this PR) in order to use the new `ci` command. Looks good to me. Merging. Please comment if any issues.

matthewmorgan:
Thanks @tejasbubane , I've been finding it hard to carve out time to work on these issues lately.  Seems like a good step forward.

PR #406, opened 2018-04-26
robot-simulator: fix test that didn't fail if nothing was thrown

jscheffner:
Solves #407:

The robot-simulator exercise includes a test that should fail if in case of an invalid argument
- no error message is thrown
- a wrong message is thrown

In fact, the test only fails in the latter case. So I fixed that. I didn't touch anything else but I really wonder why anybody would throw a String instead of an Error object.

### DIFF
--- a/exercises/robot-simulator/example.js
+++ b/exercises/robot-simulator/example.js
@@ -1,3 +1,10 @@
+export class InvalidInputError extends Error {
+  constructor(message) {
+    super();
+    this.message = message || 'Invalid Input';
+  }
+}
+
 class Robot {
   constructor() {
     this.coordinates = [0, 0];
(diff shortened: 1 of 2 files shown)


jscheffner:
I also updated the example because it didn't implement this behavior.

tejasbubane:
Hi @jscheffner Thanks for the PR and sorry for the delay. The PR looks good to me 👍 

Regarding the custom exception, we had a discussion in https://github.com/exercism/ecmascript/issues/280. We have these string-exceptions scattered across few exercises. We would like to convert them to class-based solutions but just did not find the time to fix all of them. Would you be up to doing it for this one exercise in this PR? You can refer the [circular-buffer exercise](https://github.com/exercism/ecmascript/blob/master/exercises/circular-buffer/example.js).

Let me know if otherwise, this PR looks good to go.

tejasbubane:
@jscheffner Did you get a chance to look at this?

tejasbubane:
This looks good to me 👍

PR #429, opened 2018-07-22
Deprecate binary, trinary, octal and hexadecimal

tejasbubane:
In favour of `all-your-base`. Ref: https://github.com/exercism/problem-specifications/issues/279

We did this in https://github.com/exercism/ecmascript/pull/250 and was later reformatted in https://github.com/exercism/ecmascript/pull/321/files. I guess later while adding topics, the deprecated key was removed.

I have also removed `unlocked-by: binary` for `all-your-base` - since `binary` is deprecated now.

### DIFF
--- a/config.json
+++ b/config.json
@@ -170,7 +170,8 @@
         "mathematics",
         "regular_expressions",
         "strings"
-      ]
+      ],
+      "deprecated": true
     },
     {
       "slug": "prime-factors",
@@ -688,7 +689,8 @@
         "mathematics",
         "regular_expressions",
         "strings"
-      ]
+      ],
+      "deprecated": true
     },
     {
       "slug": "sieve",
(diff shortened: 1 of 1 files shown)


matthewmorgan:
@tejasbubane does this need to be `unlocked-by` something else, otherwise `core`?

tejasbubane:
@matthewmorgan I changed `unlocked-by` to `null` instead of removing it. I am not sure about the exact requirements there, but looking at [ruby's config](https://github.com/exercism/ruby/blob/master/config.json) I guess`unlocked-by: null` and `core: false` should be fine.

matthewmorgan:
OK, looks like in the new UI, these end up unlocked from the beginning in "extra exercises." EG `minesweeper` is the same.

Thanks @tejasbubane !

PR #430, opened 2018-07-23
Update bob exercie

amscotti:
Hi, this is an update to the bob exercise to match the problem-specifications canonical-data.json, also updated `example.js` to fix a failing test for `ending with whitespace` as part of the update to the problem-specifications.

https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json


matthewmorgan:
🎊 nicely done!  Thanks for the help...

PR #433, opened 2018-07-27
Remove core from deprecated exercise

amscotti:
Hi, I'm working my way through the ECMAScript track and just got done with the Space Age exercise, it looks like Prime Factors isn't being unlocked for me in the track. Looking at the `config.json`, it looks like the `binary` exercise before Prime Factors is marked as `"core": true` but also is `"deprecated": true`. This PR is to remove `binary` from core and update the others deprecated exercises to not be unlocked by binary (which `configlet lint .` will error on)

exercism 2018-07-27 09-32-52

### DIFF
--- a/config.json
+++ b/config.json
@@ -159,7 +159,7 @@
     {
       "slug": "binary",
       "uuid": "0ba4d3b9-2519-49ac-bd93-f960aca6c11f",
-      "core": true,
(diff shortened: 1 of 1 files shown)


matthewmorgan:
Nice catch!

Issue #436, opened 2018-08-06
Test cases should not encourage default exports


xarxziux:
When designing an ES6 module, the developer has two ways of exposing internal functions and values: named exports and default exports.  Depending on the method used, the corresponding import statement looks a little different too.  However there are various articles outlining the opinion that using default exports is an anti-pattern should be avoided.

I'm only a little way into the ECMAScript challenges, but many of the test cases I've come across require the default export method to be used, which I don't think should be encouraged for many reasons.

For example take the Armstrong Numbers challenge.  This challenge only requires the developer to export one function, validate.  The test script consumes the function like this:

```JavaScript
import ArmstrongNumber from './armstrong-numbers';
```

It then uses the function like this:

```JavaScript
ArmstrongNumber.validate(input)
```

In order for the submission to pass the unit tests, the validate function has to be exported as a default export.

An alternative way of doing this is to consume the validate function like this:

```JavaScript
import {validate} from './armstrong-numbers';
```

And consume it like this:

```JavaScript
validate(input)
```

For the consuming module, this is a lot less typing.  For the submitter the only change is that they don't have to type 'default'.

The Run Length Encoding exercise does this right:

```JavaScript
import { encode, decode } from './run-length-encoding';
```

Not only is this less typing, but in a real-world scenario the first style will break the tree-shaking routines that are common in production code.  For these and other reasons, I don't think these exercises should be encouraging this pattern and should just consume the named function directly.

I've seen that there was some discussion on here about exporting classes, but I've not seen anything on this particular subject.  I'm happy to submit refactored test cases myself.  Would you be happy to accept them?

matthewmorgan:
I think this is a great idea.  I'd be happy to review any forthcoming pull requests.  :)

coagmano:
Just fired up the leap exercise and noticed that using named exports gives lint errors because of this rule in the `airbnb` preset: 
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/prefer-default-export.md

We should probably disable that specific rule, to avoid giving linting errors for the correct solution

coagmano:
Someone else found the same issue on Simple Cipher, which really confused a student:
https://exercism.io/mentor/solutions/4fd72fa7f4a648fab6a6558aa0902c76

deegeebee:
Hi, I'm very new to both exercism, JS, coding etc, and the same confuses me very much too:
I did 'Hello World' (uses named import).
Then got 'two-fer' as next (side) exercise.
[...]

PR #437, opened 2018-08-09
Fix unit tests for Armstrong Number exercise in accordance with issue #436

xarxziux:
- Change default import `ArmstrongNumbers` to named import `{ validate }`.
- Change calls to imported function from `ArmstrongNumbers.validate(input)` to the simpler `validate(input)`.

The only change for the student is that they don't need to add `default` to their export statement.

### DIFF
--- a/exercises/armstrong-numbers/armstrong-numbers.spec.js
+++ b/exercises/armstrong-numbers/armstrong-numbers.spec.js
@@ -1,43 +1,43 @@
-import ArmstrongNumber from './armstrong-numbers';
+import { validate } from './armstrong-numbers';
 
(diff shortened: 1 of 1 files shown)


xarxziux:
Scrapping this.  I neglected to update the example file.

matthewmorgan:
@xarxziux in the future, no need to close a PR when it requires changes or additions.  You can simply push a new commit.

We usually ask contributors to squash their PRs down to a single commit before we merge them, but multiple commits are not a problem for the review and discussion.

Thanks!

xarxziux:
No problem.  I thought it was simpler to simply cancel this rather than have multiple commits.

PR #439, opened 2018-08-09
Fix unit tests for Armstrong Number exercise in accordance with issue #436:

xarxziux:
- Change default import `ArmstrongNumbers` to named import `{ validate }`.
- Change calls to imported function from `ArmstrongNumbers.validate(input)` to the simpler `validate(input)`.
- Remove default export statement from example.js.
- Change export statement in example.js to a named export.


matthewmorgan:
Great!  Nice improvement in the API , thank you...

PR #440, opened 2018-08-10
Fix unit tests for Accumulate exercise

xarxziux:
As per issue #436:
- Change default import `accumulate` to named import `{ accumulate }`.


matthewmorgan:
🎆 Thanks!

PR #441, opened 2018-08-10
Fix unit tests for Alphametics exercise

xarxziux:
As per issue #436:
- Change default import `solve` to named import `{ solve }`.


matthewmorgan:
Another nice one!

PR #442, opened 2018-08-10
Fix unit tests for Atbash Cipher exercise

xarxziux:
As per issue #436:
- Change default import `atbash` to named import `{ encode }`.


matthewmorgan:
Thanks @xarxziux !

PR #443, opened 2018-08-10
Fix unit tests for Bracket-push exercise

xarxziux:
As per issue #436:
- Change default import `bracket` to named import `{ bracketPush }`.
- Change all function calls in the test suite to `bracketPush`'


matthewmorgan:
👍

PR #479, opened 2018-09-21
Get ESLint rolling

tejasbubane:
Get the ball rolling with eslint :8ball:

Changes in the PR:

* Disable `default-export` rule and eslint and prefer `named-export` instead
* Add eslint to Makefile to travis runs it and add `.eslintignore` file with all exercise directories. Plan is to remove each line from there while fixing eslint for that exercise.
* Use `airbnb-base` instead of complete `airbnb`. Since this is a plain JS repo (no react/JSX) - base is what we need.
* Fix `hello-world` exercise (and in the process change it to not use class).


matthewmorgan:
Thanks @tejasbubane!  Welcome and overdue updates...  :)

Issue #480, opened 2018-09-25
Fix eslint errors in all exercises


tejasbubane:
I have configured eslint in https://github.com/exercism/javascript/pull/479

But have not fixed errors in any exercise apart from `hello-world`. Instead I added all exercises to [.eslintignore](https://github.com/exercism/javascript/blob/master/.eslintignore). 

The plan is to do this:
1. Pick one exercise (You can pick many but we prefer to have separate PRs so that it is easy to review - but if there are not many changes - you can pick 2-3 at once).
2. Remove the exercise name from `.eslintignore` file
3. Run `npm run lint` - This should show errors in those exercises.
4. Fix those errors - commit and raise a PR 🎉 

Note: The `eslint` configuration is not final. I did not have a good look at all errors, just put a sensible default in place. While fixing, if you encounter some error that you feel we should allow(ignore), feel free to comment on this issue.

gargrave:
I'll help out with some of these over the weekend! 👍 Given how small the overall codebase is, it really shouldn't be too big of a deal to get all of these taken of pretty quickly.

gargrave:
@tarunvelli and I have a handful of PRs open for this--do you need review help? I am happy to help review PRs if that will help expedite things.

tejasbubane:
Thanks a lot for the PRs! I am on a vacation for a week and would be able to look at them next week only. Meanwhile some other maintainers might have a chance to review.

matthewmorgan:
@gargrave @tejasbubane @tarunvelli I'm looking at the PRs now.  Thanks for the patience.

ganderzz:
Trying my hand at helping out if there are no issues with it. :) Should be able to bust out a few.

tarunvelli:
I think the hacktoberfest label should be added to this issue, as it might help get this issue resolved faster

Futuro212:
Hi everyone,

So I'm new with this project and when I try to run the lint command, it shows errors about `Expected linebreaks to be 'LF' but found 'CRLF'`. I'm running this on windows. Should I change some configuration in the eslint?

matthewmorgan:
@Futuro212 let's not comment on closed issues please.  Would you mind opening a separate, new issue for this?  Hopefully someone can help.  Thanks!

serixscorpio:
[...]

PR #486, opened 2018-10-01
Fix linting for pangram exercise ( #480 )

tarunvelli:


exercises/pangram/example.js
```diff
@@ -1,10 +1,10 @@
-const notAlpha = /[^a-z]+/gi,
-  alphaLength = 26;
-let cleaned,
-  sortedSet;
+const notAlpha = /[^a-z]+/gi;
+const alphaLength = 26;
+let cleaned;
+let sortedSet;
```
gargrave: Not to split hairs, but is there is any reason `cleaned` and `sortedSet` need to be defined here at all? Seems like it would be better to just declare them as `const`s where they are used, like:

```js
export const isPangram = (candidate) => {
  const cleaned = (candidate.replace(notAlpha, '')).toLowerCase();
  const sortedSet = new Set([...cleaned].sort());
  return sortedSet.size === alphaLength;
};
```

tarunvelli: @gargrave you're right it is not needed to define them there, updated the PR accordingly 👍

tarunvelli:
Updated the PR as suggested by @gargrave

PR #489, opened 2018-10-02
fix: Eslint errors in all-your-base

OrthoDex:


gargrave:
I would personally argue against disabling rules inline, as it kind of defeats the purpose of using ESLint to begin with. 

The `no-plus-plus` rule is easy enough abide by, as you mentioned, by just changing the `++` to `+= 1`. And for the `class-methods-use-this` rule, you could probably just either make that method static, or ideally not use a class at all and instead just export the function directly, since it isn't really needed.

On the second point, I am finding that a lot of these examples are not written in very idiomatic JavaScript, and it kind of seems like they may have been initially prepared by people with more experience in other languages, so sometimes they need a little more rethinking than just fixing a quick linting error.

I'm also not sure what the status of reviews/merges is, as these PRs are starting to pile up.

OrthoDex:
@gargrave agreed, I refrained from changing the code too much, since I am not sure if it would break any upstream dependencies or not.

As for the PRs, I am more than happy to lend a hand in reviewing them. :smiley:

matthewmorgan:
Good conversation on this one.

1. I think this example could use some cleanup.  It's a little confusing that it uses a mixture of class methods and module-scoped helpers.  Sometimes that's OK, but in a simple example like this it feels like unneeded complexity.
2. Let's not disable line rules just so we can pass the linting tests-- fixing the code is just as easy.  If we don't want that lint rule, that's a track-level decision.
3. Locking the node version sounds like a fine idea, though I expect we will continue updating it deliberately.

@OrthoDex would you mind updating this PR to reflect (2) if you're comfortable with that?  I think that restructuring the example and locking the nose version should be separate issues/PRs.

OrthoDex:
@matthewmorgan I have updated the PR to use a function prototype. I feel (1) and @gargrave 's review about writing these examples in idiomatic javascript are related and additional linting rules might help in that regard.

We can also make the helper functions a part of the exported function's prototype, but I see little value there unless they have the potential to be reused in other examples.

PR #500, opened 2018-10-09
queen-attack.spec.js file does not test "edge" case

### LINKED ISSUE #499: queen-attack spec file does not test "edge" case
While working with a friend on queen-attack, I noticed his solution passed all the provided tests in [queen-attack.spec.js](https://github.com/exercism/javascript/blob/master/exercises/queen-attack/queen-attack.spec.js) - but would fail given a specific edge case. This situation would arise if a queen was placed on the last position in a row.

For example, in the spec file; if 

`const positioning = { white: [7, 7], black: [0, 0] };`
and
`expect(queens.toString()).toEqual(board);`

[their solution](https://github.com/AloeareV/Queen/commit/2361af071dc2439af560e7e9aa0ee137aa48c349) would fail--though it currently passes all provided tests.

IndelicateArgot:
Hey @matthewmorgan I have been working with your spec file, and I noticed you were the last one who touched it and seemed the most active! If you'd take a look at this, I would appreciate it.

Issues:
Fixes #499

Problem:
Solutions which current pass all tests would fail in an edge case.

Analysis:
Added a test in queen-attack.spec.js which tests for this edge case.

Tests:
Ran tests with npm against example solution mentioned in issue #499 to confirm that this example solution fails for new edge case test, and that test only.

I was working with @AloeareV to uncover this, thank you!!

### DIFF
--- a/exercises/queen-attack/queen-attack.spec.js
+++ b/exercises/queen-attack/queen-attack.spec.js
@@ -33,6 +33,20 @@ describe('Queens', () => {
     expect(queens.toString()).toEqual(board);
   });
 
+  xtest('toString representation edge case', () => {
+    const positioning = { white: [7, 7], black: [0, 0] };
+    const queens = new Queens(positioning);
+    const board = ['B _ _ _ _ _ _ _',
+      '_ _ _ _ _ _ _ _',
+      '_ _ _ _ _ _ _ _',
+      '_ _ _ _ _ _ _ _',
+      '_ _ _ _ _ _ _ _',
+      '_ _ _ _ _ _ _ _',
+      '_ _ _ _ _ _ _ _',
+      '_ _ _ _ _ _ _ W\n'].join('\n');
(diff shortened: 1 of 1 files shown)


matthewmorgan:
@IndelicateArgot thank you for your contribution.  Testing edge cases is clearly valuable.

Can I suggest that this PR would be valuable in [problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/queen-attack), as a track-wide check?  Maybe some additional boundary conditions should be checked as well.

I'm going to go ahead and merge this with thanks!

IndelicateArgot:
Hi @matthewmorgan!

I just wanted to say thank you for checking in with the PR and moving so quickly! I'll follow up in [problem-specifications](https://github.com/exercism/problem-specifications) and link to this PR, and I'll circle back that way if I find something else along these lines.

Thanks again!

PR #505, opened 2018-10-15
Fix eslint errors in hamming exercise

serixscorpio:
Per #480, fix the following lint errors for the hamming exercise:

```sh
/home/eric/code/javascript/exercises/hamming/example.js
   1:15  error  Block must not be padded by blank lines               padded-blocks
   3:10  error  Expected 'this' to be used by class method 'compute'  class-methods-use-this
  13:12  error  Unary operator '++' used                              no-plusplus
  15:9   error  Unary operator '++' used                              no-plusplus
  22:1   error  Block must not be padded by blank lines               padded-blocks
```
Note:
1. The `while` loop is replaced with `for` to allow easier replacement from `++` to `+=`.
2. The error "Expected `this` to be used by class method `compute`" is addressed by exporting the standalone `compute` function instead of exporting an entire `Hamming` class.

### DIFF
--- a/exercises/hamming/example.js
+++ b/exercises/hamming/example.js
@@ -1,6 +1,5 @@
 class Hamming {
-
-  compute(strand1, strand2) {
+  static compute(strand1, strand2) {
     const len1 = strand1.length;
     const len2 = strand2.length;
     if (len1 !== len2) {
@@ -8,17 +7,15 @@ class Hamming {
     }
 
     let distance = 0;
-    let idx = -1;
     const end = len1; // there could be len2, they're equal
-    while (++idx < end) {
+    for (let idx = 0; idx < end; idx += 1) {
       if (strand1[idx] !== strand2[idx]) {
-        distance++;
+        distance += 1;
       }
     }
 
     return distance;
   }
-
 }
 
 export default Hamming;
--- a/exercises/hamming/example.js
+++ b/exercises/hamming/example.js
@@ -5,13 +5,8 @@ export const compute = (strand1, strand2) => {
     throw new Error('left and right strands must be of equal length');
   }
 
-  let distance = 0;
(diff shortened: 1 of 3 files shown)


tejasbubane:
This changes default class export to named function export which I am ok with as per https://github.com/exercism/javascript/issues/436

exercises/hamming/example.js
```diff
@@ -1,24 +1,17 @@
...
+      distance += 1;
     }
-
-    return distance;
   }
 
-}
-
-export default Hamming;
+  return distance;
```
tejasbubane: Using reduce should remove declaring extra variable `distance`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

Like `return array.reduce()`

I am OK with merging this as is, but would favour the `reduce` version. Would you mind changing it?

exercises/hamming/example.js
```diff
@@ -1,24 +1,12 @@
...
+  const len1 = strand1.length;
+  const len2 = strand2.length;
+  if (len1 !== len2) {
+    throw new Error('left and right strands must be of equal length');
   }
 
-}
-
-export default Hamming;
+  return [...strand1].reduce(
```
matthewmorgan: Since we know the strands must be of equal length, this could also be written, perhaps more semantically, like so:
```suggestion
  return [...strand1].filter((el, idx) => el !== strand2[idx]).length;
```

matthewmorgan:
@serixscorpio @tejasbubane I have one suggestion, but please feel free to take it or leave it.  This PR looks fine as it is to me.

serixscorpio:
@matthewmorgan thanks for your feedback, `filter` seems to be an even better fit! I'll update.

PR #511, opened 2018-10-17
Lint sieve and pig-latin exercises

ntshrocha:
Per #480, fixes the following lint errors for the sieve exercise:
```
/javascript/exercises/sieve/example.js
  15:3   error  Split 'let' declarations into multiple statements  one-var
  15:14  error  Expected variable declaration to be on a new line  one-var-declaration-per-line

/javascript/exercises/sieve/sieve.spec.js
  13:1  error  Line 13 exceeds the maximum line length of 100  max-len
```

Also fixes the following lint error for the pig-latin exercise:
```
/javascript/exercises/pig-latin/example.js
  13:12  error  Expected 'this' to be used by class method 'translate'  class-methods-use-this
```


matthewmorgan:
@ntshrocha looks like there is a merge conflict in your `.eslintconfig`.  Can you please pull in master and resolve the conflict?

No need to open a new PR, just make the fix on this one and push another commit.

Let us know if we can be of any assistance.

exercises/pig-latin/pig-latin.spec.js
```diff
@@ -1,7 +1,7 @@
 import PigLatin from './pig-latin';
```
matthewmorgan: I have no problem going from a class-based solution to an object-based solution, but if we're going to do that, we can dispense with the `PigLatin` reference altogether:

```suggestion
import translator from './pig-latin';

  describe('Pig Latin', () => {
    
    test('translates a word beginning with a', () => {
```

ntshrocha:
@matthewmorgan you're right, that makes sense. Working on it!

PR #530, opened 2018-10-22
lint say example

msomji:
Lint Say example (associated to #480 )

### DIFF
--- a/exercises/say/example.js
+++ b/exercises/say/example.js
@@ -1,79 +1,88 @@
-const smallNumbers = {
-  0: 'zero',
-  1: 'one',
-  2: 'two',
-  3: 'three',
-  4: 'four',
-  5: 'five',
-  6: 'six',
-  7: 'seven',
-  8: 'eight',
-  9: 'nine',
-  10: 'ten',
-  11: 'eleven',
-  12: 'twelve',
-  13: 'thirteen',
-  14: 'fourteen',
-  15: 'fifteen',
-  16: 'sixteen',
-  17: 'seventeen',
-  18: 'eighteen',
-  19: 'nineteen',
-};
-
-const decades = {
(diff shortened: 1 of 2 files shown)


exercises/say/example.js
```diff
@@ -38,42 +38,47 @@ const bigNumbers = {
...
+      result += `${threeDigit(factor)} ${bigNumbers[bigNumber]} `;
+      num.current -= factor * bigNumber;
     }
-    throw new Error('Number must be between 0 and 999,999,999,999.');
   }
+  return result;
 }
+export default class Say {
+}
+Say.prototype.inEnglish = (n) => {
```
tejasbubane: Since you are using prototype, `class` is not really required. A function should be sufficient. eg. https://github.com/exercism/javascript/blob/5062905a6b635623cc1914c519083925e661ea1e/exercises/all-your-base/example.js#L27-L29

matthewmorgan: Should we instead use class, since this is ES6 now?  Wondering if anyone has an opinion on this: `Function.prototype` feels more ES5 and prior to me, although this PR is for linting, so I don't want to conflate the two issues unless it's sensible to address them at the same time.

tejasbubane: > although this PR is for linting, so I don't want to conflate the two issues unless it's sensible to address them at the same time.

Exactly my thoughts. I saw prototype being used on few other exercises as well. Let us create separate issue to remove prototypes from all of them.

matthewmorgan: +1

msomji: @tejasbubane @matthewmorgan I made the changes to use ES6 classes.
- sorry for the delay 👼

matthewmorgan: @msomji thanks for your updates!

PR #537, opened 2018-10-28
high-scores: Implement exercise

PakkuDon:
New problem specification added in exercism/problem-specifications#1378

### DIFF
--- a/config.json
+++ b/config.json
@@ -1361,6 +1361,16 @@
         "bitwise_operations",
         "transforming"
       ]
+    },
+    {
+      "slug": "high-scores",
+      "uuid": "19b41c4a-1466-5c19-b547-40284189fd18",
+      "core": false,
+      "unlocked_by": "acronym",
+      "difficulty": 2,
+      "topics": [
+        "arrays"
+      ]
     }
   ]
 }
--- a/config.json
+++ b/config.json
@@ -1366,7 +1366,7 @@
       "slug": "high-scores",
       "uuid": "19b41c4a-1466-5c19-b547-40284189fd18",
       "core": false,
-      "unlocked_by": "acronym",
+      "unlocked_by": "pangram",
       "difficulty": 2,
       "topics": [
         "arrays"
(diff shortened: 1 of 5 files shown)


config.json
```diff
@@ -1361,6 +1361,16 @@
...
         "transforming"
       ]
+    },
+    {
+      "slug": "high-scores",
+      "uuid": "19b41c4a-1466-5c19-b547-40284189fd18",
+      "core": false,
+      "unlocked_by": "acronym",
+      "difficulty": 2,
+      "topics": [
```
PakkuDon: Using the same `core`, `unlocked_by`, `difficulty` and `topics` as [the Ruby track](https://github.com/exercism/ruby/blob/master/config.json#L225). Do these settings make sense for this track?

PakkuDon: Just saw the [build](https://travis-ci.org/exercism/javascript/builds/447392671?utm_source=github_status&utm_medium=notification). 😅 Apparently not  

> The command "bin/fetch-configlet" exited with 0.
> 0.19s$ bin/configlet lint .
> -> The exercise 'high-scores' is being unlocked by a non-core exercise. Non-core exercises can only be unlocked by core exercises.

Anyone have any suggestions? Would anyone object to `unlocked_by` being set to null?

matthewmorgan: @PakkuDon I'm thinking: have it be unlocked by `pangram`?  Since that's the one that unlocks `acronym`?

exercises/high-scores/example.js
```diff
@@ -0,0 +1,27 @@
+export class HighScores {
+  constructor(scores) {
+    this.scores = scores;
+  }
+
+  latest() {
```
matthewmorgan: I wonder if we should take this exercise as an opportunity to allow [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) or object properties rather than requiring functions to return these values.

I ask because this is a feature of ES 2015 and beyond, and we're trying to make a point of exposing users to them.

@PakkuDon what do you think about that?

PakkuDon: @matthewmorgan TIL, thanks for the article! 
Switched to getters in d14d8c3

matthewmorgan:
Beauteous!

PR #545, opened 2018-11-01
Improve code quality of perfect numbers exercise

### LINKED ISSUE #544: Fix some code quality issues with Perfect Numbers
The [Perfect Numbers exercise](https://github.com/exercism/javascript/tree/master/exercises/perfect-numbers) on this track has a few issues:

1. The suite requires a class, but then only calls a stateless method on the class instance. Either this method should be [static](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static), or better yet, the test should expect a simple function and do away with the ceremony of classes.
2. The suite looks for strings to be thrown for invalid inputs rather than [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) objects.
3. The test suite does not validate input until the end.  This isn't really a good pattern, TDD-wise.  The final `describe` block should be first.
4. By the same token as (3), the deficient 'edge case' test should be the first test in that `describe` block, as it's really a form of input domain validation.

If a first-timer wants to take this on, I would be happy to help.  If no one takes soon, I'll tackle it.

Thanks everyone!

serixscorpio:
closes #544.   Thanks for the detailed suggestions on how to improve the exercise.  The commits largely correspond to each suggestion.
In addition, I included a bit of linting at 08e7135 (more related to #480).  If it's preferred, I can spilt them as separate PRs.  Thanks.

### DIFF
--- a/exercises/perfect-numbers/example.js
+++ b/exercises/perfect-numbers/example.js
@@ -1,12 +1,17 @@
-
-export default class PerfectNumbers {
+/**
+  * Classify a given number as perfect, abundant or deficient.
+  * @param {number} number - number to be classified. Note: if the number is equal or smaller than 0,
+  * then returns an error: Classification is only possible for natural numbers.
+  * @returns {string} - string that contains the number classification (perfect, abundant or deficient).
+  */
+export const classify = (number) => {
 
   /**
    * Calculate all the divisors for a given number.
    * @param {number} number - natural number.
    * @returns {array} array that contains the divisors for a given number NOT including the number itself.
    */
-  getDivisors(number) {
+  function getDivisors(number) {
     let i;
     const divs = new Array();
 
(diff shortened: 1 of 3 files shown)


exercises/perfect-numbers/example.js
```diff
@@ -1,64 +1,49 @@
...
-    } else {
-      result = 'deficient';
-    }
+  // Factorize the current number.
+  const divsArray = getDivisors(number);
 
-    return result;
-  }
+  // Sum the factors.
+  const sum = divsArray.reduce((partialSum, div) => partialSum + div, 0);
```
matthewmorgan: One small thing here:  `Array.reduce` will use the value of first element of the list as the initial accumulator value when you don't supply an initial value explicitly.  

In effect, instead of adding zero to the first element in the first pass, you start by adding the first element to the second one, and continue from there:

```suggestion
const sum = divsArray.reduce((partialSum, div) => partialSum + div);
```

exercises/perfect-numbers/example.js
```diff
@@ -1,64 +1,49 @@
...
-        divs.push(i);
-      }
-    }
-
+/**
+  * Calculate all the divisors for a given number.
+  * @param {number} number - natural number.
+  * @returns {array} that contains the divisors for a given number excluding the number itself.
+  */
+function getDivisors(number) {
```
matthewmorgan: I like the way you've set this method up!  At first I was puzzled why you weren't returning `[1]` for every case, but now I see what you're doing.

exercises/perfect-numbers/example.js
serixscorpio: Thanks for your suggestion! Unfortunately, it is possible for `divsArray` to be empty (consider edge case when classifying number `1`).  When `divsArray` is empty, the initial accumulator `0` is needed in order to arrive at the correct aliquot sum.  Without the initial accumulator, the `reduce` would raise an error on an empty array.

matthewmorgan: Good point.  I did not consider that. 😁

PR #546, opened 2018-11-02
Update example and tests for simple-linked-list

matthewmorgan:
This exercise was implemented in a way that really deviated from the README in the problem-specifications repo.

The README is very simple in its requirements:
1. Implement List and Element classes.
2. Implement the List class so that it can:
a) be reversed
b) return an array of element values
c) accept an array as initial data

There's nothing about `pop`, `shift`, etc. The existing implementation is relatively complicated compared to these requirements, as is the test suite, which requires validation that isn't specified in the metadata.

This exercise is already a bit of a mind-bender without all the extra complications-- let's make it simpler by focusing on the core problem.  There are two other exercises that focus on list implementations, so I think we're pretty well covered here.

### DIFF
--- a/exercises/simple-linked-list/example.js
+++ b/exercises/simple-linked-list/example.js
@@ -1,111 +1,45 @@
-export class ElementValueRequiredException extends Error {}
-export class ElementNextNotInstanceException extends Error {}
-
 export class Element {
-  constructor(value, next) {
-    if (value === undefined) {
-      throw new ElementValueRequiredException();
-    }
-
-    if (next !== undefined && !(next instanceof Element)) {
-      throw new ElementNextNotInstanceException();
-    }
-
+  constructor(value) {
     this.value = value;
-    this.next = next;
+    this.next = null;
   }
 }
 
 export class List {
-  static fromArray(array) {
-    const list = new List();
-    array.forEach(item => list.push(new Element(item)));
-
-    return list;
-  }
-
-  push(value) {
-    const newEl = (value instanceof Element)
-      ? value
-      : new Element(value);
-
-    if (!this.head) {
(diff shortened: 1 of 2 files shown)


tejasbubane:
Do we have to expose `Element` class to users in tests? Can we have `list.add(1)` directly? 
I think that will give users flexibility to have some other way of implementing the exercise. eg. Instead of `class`, one could use plain JS object with two properties `value` and `next` - or even arrays for that matter, though I would not treat the later as linked list.

exercises/simple-linked-list/example.js
```diff
@@ -1,111 +1,52 @@
...
-
-    if (!penultEl) {
-      this.head = undefined;
-    } else {
-      penultEl.next = undefined;
+  countElements(c, element) {
+    let count = c;
+    while (element.next) {
+      count += 1;
+      return this.countElements(count, element.next);
```
tejasbubane: You can use either one of loop or recursion. Just like you did with `toArray`. A return condition before recursion call should work.

matthewmorgan:
@tejasbubane I exposed the `Element` class and used it in tests because the README requires implementing it.

Normally I would agree that this implementation detail should be hidden from consumers of `List`, but because of both the limited scope of the exercise and the explicit call out of the requirement in the README, I thought it should be like this.

Does that make sense?

exercises/simple-linked-list/example.js
matthewmorgan: Yes!  I think `reverse` can be recursive also.

I'm going to make those changes as soon as I have a few minutes.

tejasbubane:
Ok it makes sense after reading the README. 👍

PR #548, opened 2018-11-03
Clean up indent errors

xarxziux:
Clean up indent errors as part of issue #480
- Fix 30 cases of ESLint issue `indent`.

As an added bonus, I've been able to remove the ISBN Identifier and Matrix exercises from .eslintignore.


xarxziux:
Sorry about that.  I'd assumed that the two couldn't affect each other.  Fixed.

matthewmorgan:
I didn't see exactly where the conflict was, but any PRs that modify adjacent lines can have this problem-- it doesn't have to be the same lines.

Thanks!

PR #549, opened 2018-11-06
Clean-up of ESLint errors

xarxziux:
Clean-up of small, one-line ESLint errors as part of issue #480:
- Hexadecimal: fix `no-plusplus` error.
- Kindergarten Garden: fix `no-shadow` error.
- Luhn: fix `no-param-reassign` error.
- Rotational Cipher: fix `operational-linebreak` error.
- Transpose: fix `implicit-arrow-linebreak` error.

### DIFF
--- a/exercises/hexadecimal/example.js
+++ b/exercises/hexadecimal/example.js
@@ -4,7 +4,7 @@ export default function (hex) {
   this.toDecimal = () => {
     const hexCharacters = [...this.hex];
 
-    for (let i = 0; i < hexCharacters.length; i++) {
+    for (let i = 0; i < hexCharacters.length; i + 1) {
       if (/[^0-9a-fA-F]/.exec(hexCharacters[i])) { return 0; }
     }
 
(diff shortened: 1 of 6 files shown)


exercises/hexadecimal/example.js
```diff
@@ -4,7 +4,7 @@ export default function (hex) {
   this.toDecimal = () => {
     const hexCharacters = [...this.hex];
 
-    for (let i = 0; i < hexCharacters.length; i++) {
+    for (let i = 0; i < hexCharacters.length; i + 1) {
```
serixscorpio: I think this might have just been a typo, perhaps it's intended to be `i += 1` instead of `i + 1`?  This may have been the cause of the CI timing out (due to infinite loop), cc @tejasbubane

tejasbubane: Ahh yes! I missed this. Good catch @serixscorpio

xarxziux: Nice catch.  That was it!

PR #550, opened 2018-11-07
`concat` should be applied to a list of lists.

paparomeo:
`List.concat` in the JavaScript track is just a duplicate of `List.append`.

This PR makes it more useful by expecting a list of lists as the argument instead, aligning it with, for example, the [test suite in the Python track](https://github.com/exercism/python/blob/master/exercises/list-ops/list_ops_test.py#L26).

### DIFF
--- a/exercises/list-ops/example.js
+++ b/exercises/list-ops/example.js
@@ -10,8 +10,11 @@ class List {
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


matthewmorgan:
Thanks @paparomeo !

PR #552, opened 2018-11-09
lint proverb exercise

serixscorpio:
per #480, lint proverb exercise.


matthewmorgan:
Thanks @serixscorpio !!

PR #556, opened 2018-11-13
Use named export for largest series product

serixscorpio:
Per #436, convert default export to named export for largest series
product exercise. Change from class style `new Series(x).largestProduct(y)`
to `largestSeriesProduct(x, y)` since there is no need to keep state. See 8a38f2f.

As part of this change, I also simplified and linted the pre-existing logic in the example solution (
c2d080c, e9e737b, c450f53 and 40504cf.  If these are too noisy to be within the same PR, please let me know.  I can break these into a separate PR.  Thanks!


tejasbubane:
I think it would be better if you move the named export changes to another PR and leave this one for eslint only.

matthewmorgan:
@serixscorpio we appreciate your contributions, truly!  But I am in agreement with @tejasbubane that the PRs should be logically separate.  Just makes it easier for the easily confused among us to follow the plot :)

serixscorpio:
Thanks for your feedback, I'll open a separate PR to lint first, and another one to move to named export (in the order suggested) to reduce the cognitive load.  Thanks.

PR #563, opened 2018-11-18
Fix ESLint errors in OCR Number test script

xarxziux:
As per issue #480:
- Fix all ESLint errors in ocr-numbers.spec.js file.

I have NOT removed the OCR Numbers entry in .eslintconfig as the example,js file has other errors which require non-trivial fixes.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -14,7 +14,6 @@ list-ops
 meetup
 minesweeper
 nth-prime
-ocr-numbers
 palindrome-products
 point-mutations
 prime-factors
--- a/exercises/ocr-numbers/example.js
+++ b/exercises/ocr-numbers/example.js
@@ -41,6 +41,7 @@ const PATTERNS = {
     '   '],
 };
 export default class Parser {
+  // eslint-disable-next-line class-methods-use-this
   convert(text) {
     return Parser.splitIntoRows(text).map(Parser.valuesInRow).join(',');
   }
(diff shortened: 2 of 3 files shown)


tejasbubane:
@xarxziux For non-trivial fixes you can choose to disable specific rules of eslint in the file. https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

tejasbubane:
I would suggest fixing example.js (if required by disabling rules) in this PR itself.

xarxziux:
> @xarxziux For non-trivial fixes you can choose to disable specific rules of eslint in the file. https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

I've been meaning to ask what the convention was on that.

tejasbubane:
> I've been meaning to ask what the convention was on that.

There has been no discussion I think. I expect to discuss these on PRs as need arises.

tejasbubane:
@xarxziux I think you can actually change the `convert` function to be `static` - since there is not initializer. You will then need to change the test cases as well.

xarxziux:
If I'm changing the `convert` function to be `static`, I may as well drop the class altogether and just export a function.  There's no real need for this solution to export a class anyway as there's no internal state and you're only calling a single function.

I wasn't planning on re-designing the whole example, but I don't mind doing so if you're OK with that.

tejasbubane:
I am OK with that - in fact I encourage it. You have already done similar changes for few exercises as per #436

tejasbubane:
Another way would be to have an API like `new Ocr(text).convert`. This way text will be stored as a property in constructor and all static methods could be converted to normal instance methods.

tejasbubane:
Giving this another thought, I think we need to discuss this separately. Similar pattern can be seen in another PR I just merged (https://github.com/exercism/javascript/pull/566/files#diff-3413c81ea73421151342fbdca24dc471).

If you have not done any API changes I would suggest to hold on - I will merge this as is and we will discuss the API change in a separate issue. Let me know.

xarxziux:
I've done a full conversion of the OCR exercise and raised a separate pull request: #567.  There shouldn't be a problem merging this PR, while we weigh up the merits of the other one.

PR #567, opened 2018-11-22
OCR convert to function

xarxziux:
As discussed in issue #563, this is s full rewrite of the OCR exercise, ditching the class design of the original.

### DIFF
--- a/exercises/ocr-numbers/example.js
+++ b/exercises/ocr-numbers/example.js
@@ -40,50 +40,44 @@ const PATTERNS = {
     ' _|',
     '   '],
 };
-export default class Parser {
-  // eslint-disable-next-line class-methods-use-this
-  convert(text) {
-    return Parser.splitIntoRows(text).map(Parser.valuesInRow).join(',');
-  }
 
-  static valuesInRow(row) {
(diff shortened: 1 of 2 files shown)


SleeplessByte:
Hi there @xarxziux, 

Thank you for this work. Would you be so kind to update this to the latest master (and resolve the conflicts)? We'll be happy to review (and merge) it soon thereafter.

xarxziux:
OK, it took me a little while to figure out what I was doing here, but I think I have it now!

tejasbubane:
For the record, this is done because all methods in [the Parser class were static]( https://github.com/exercism/javascript/blob/9c009439465975142db21dd62d3185916174348f/exercises/ocr-numbers/example.js) - so it did not make much sense to keep class.

SleeplessByte:
(sidenote: @tejasbubane do we run example.js on CI to make sure there is a working implementation?)

tejasbubane:
> do we run example.js on CI to make sure there is a working implementation? 

Yes we run the test cases against examples. This Makefile does all the heavy lifting: [ref](https://github.com/exercism/javascript/blob/master/Makefile#L25)

PR #568, opened 2018-11-22
Fix various trivial ESLint issues

xarxziux:
As per issue #480, fix the following ESLint errors:
- clock/example.js: 1 `no-param-reassign` and 1 `max-len` error.
- prime-factors/example.js: 1 `no-param-reassign` and 1 `no-plusplus` error.
- rectangles/example.js: 9 `no-plusplus` errors.
- strain/example.js: 1 `no-plusplus` error.


exercises/rectangles/example.js
```diff
@@ -1,37 +0,0 @@
...
-          }
-        }
-      }
-    }
-
-    return rectangles;
-  }
-}
-
-export default Rectangles;
```
tejasbubane: I wonder why this diff shows for entire file? I first thought the file is deleted 🤔

PR #569, opened 2018-11-26
Minesweeper: change default export to named export

serixscorpio:
per #436, change default export to named export.  The test code now
imports and consumes differently.  Also note the `class` structure was
unnecessary from the example solution, so just export the `annotate`
function directly instead.

```js
// before
import Minesweeper from './minesweeper'
Minesweeper.annotate(...) // example usage

// after
import { annotate } from './minesweeper'
annotate(...) // example usage
```


matthewmorgan:
Great improvements!  thanks @serixscorpio !!

PR #576, opened 2018-12-07
matrix: change from default to named export

serixscorpio:
Per #436 , for `matrix` exercise, change from default to named export.


matthewmorgan:
Just what we needed!  thanks @serixscorpio

PR #580, opened 2018-12-10
Fix ESLint errors in Allergies, Linked List and Point Mutations

### LINKED ISSUE #480: Fix eslint errors in all exercises

thanhcng:
This PR fixes #480.
Sorry if I have fixed too many exercises since I'm doing this PR for an assignment in college.
Thanks!

### DIFF
--- a/exercises/allergies/example.js
+++ b/exercises/allergies/example.js
@@ -15,6 +15,7 @@ class Allergies {
   }
 
   list() {
+    // eslint-disable-next-line no-bitwise, no-restricted-properties
     return possibleAllergies.filter((allergy, i) => this.allergenIndex & Math.pow(2, i));
   }
 
--- a/exercises/linked-list/example.js
+++ b/exercises/linked-list/example.js
@@ -14,11 +14,12 @@ export default class Deque {
       return undefined;
     }
 
-    const value = this.head.value;
+    const { value } = this.head;
     if (this.head.next) {
       this.head = this.head.next;
       this.head.prev = null;
     } else {
+      // eslint-disable-next-line no-multi-assign
       this.head = this.tail = null;
     }
 
@@ -42,11 +43,12 @@ export default class Deque {
       return undefined;
     }
 
-    const value = this.tail.value;
+    const { value } = this.tail;
     if (this.tail.prev) {
       this.tail = this.tail.prev;
       this.tail.next = null;
     } else {
+      // eslint-disable-next-line no-multi-assign
       this.head = this.tail = null;
     }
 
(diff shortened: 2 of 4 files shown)


exercises/linked-list/example.js
```diff
@@ -14,11 +14,12 @@ export default class Deque {
       return undefined;
     }
 
-    const value = this.head.value;
+    const { value } = this.head;
     if (this.head.next) {
       this.head = this.head.next;
       this.head.prev = null;
     } else {
+      // eslint-disable-next-line no-multi-assign
```
tejasbubane: please don't disable properties which can be easily fixed.

exercises/allergies/example.js
```diff
@@ -15,6 +15,7 @@ class Allergies {
   }
 
   list() {
+    // eslint-disable-next-line no-bitwise, no-restricted-properties
     return possibleAllergies.filter((allergy, i) => this.allergenIndex & Math.pow(2, i));
```
tejasbubane: That `&` should actually be replaced by `&&` - this will remove fix the `no-bitwise` error.

exercises/linked-list/example.js
```diff
@@ -42,11 +43,12 @@ export default class Deque {
       return undefined;
     }
 
-    const value = this.tail.value;
+    const { value } = this.tail;
     if (this.tail.prev) {
       this.tail = this.tail.prev;
       this.tail.next = null;
     } else {
+      // eslint-disable-next-line no-multi-assign
```
tejasbubane: same here

tejasbubane:
Please remove `fixes` from your commit/PR message. It will close the issue which we don't want right now. Issue will be closed after all the exercises are fixed.

exercises/allergies/example.js
matthewmorgan: Is that what we want here?  I think we actually want the bitwise operator, don't we?

exercises/linked-list/example.js
matthewmorgan: Agreed.

PR #584, opened 2019-01-03
raindrops: convert to function

cmccandless:
Similar in nature to #567, but for the `raindrops` exercise. There is no stateful information used in this exercise, so a class-based implementation is not necessary.


tejasbubane:
Ref issue: https://github.com/exercism/javascript/issues/436

PR #595, opened 2019-01-23
robot-name: change optional test to enforce better implementation

SleeplessByte:
According to the instructions, you may never generate a name that has already been used. Using random makes this improbable, but not impossible. This changes the final test so that a different approach is required to satisfy the entire exercise.

This is in line with the ruby tests (even though those have issues).

### DIFF
--- a/exercises/robot-name/robot-name.spec.js
+++ b/exercises/robot-name/robot-name.spec.js
@@ -15,13 +15,6 @@ const areSequential = (name1, name2) => {
   return Math.abs(totalDiff) <= 1;
 };
 
-// Change these when adding / changing / removing tests in order to keep the
-// final test not exhausting all the names that can be generated.
-
-const NAMES_USED_IN_TESTS = 9 // beforeEach
-  + 3 // new Robot()
-  + 10003; // reset()
-
 const TOTAL_NUMBER_OF_NAMES = 26 // A-Z
   * 26 // A-Z
   * 10 // 0-9
@@ -33,6 +26,9 @@ describe('Robot', () => {
 
   beforeEach(() => {
     robot = new Robot();
+  });
+  afterEach(() => {
+    Robot.releaseNames();
   });
 
   test('has a name', () => {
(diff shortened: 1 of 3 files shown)


exercises/robot-name/robot-name.spec.js
```diff
@@ -15,6 +15,19 @@ const areSequential = (name1, name2) => {
   return Math.abs(totalDiff) <= 1;
 };
 
+// Change these when adding / changing / removing tests in order to keep the
+// final test not exhausting all the names that can be generated.
+
+const NAMES_USED_IN_TESTS = 9 // beforeEach
+  + 3 // new Robot()
+  + 10003; // reset()
```
tejasbubane: Can we add `reset` to `afterEach` so that the state gets cleared and every test is independent?

SleeplessByte: Unfortunately, `reset` is _not_ a method that clears the state in this exercise. Rather, it just assigns a new name. So calling `reset` after each test actually pollutes the state even more.

SleeplessByte:
@tejasbubane any other comments on this?

exercises/robot-name/robot-name.spec.js
tejasbubane: I meant `release_names`

tejasbubane:
@SleeplessByte I have posted some suggested changes in this commit: https://github.com/tejasbubane/xjavascript/commit/5f96720786a76ab13bd4c78c69522b3dbdcbecb0

* Use `releaseNames` in afterEach (ideal to restore state after each test run)
* Rename `release_names` to `releaseNames` to be consistent with JS naming
* Remove `reshuffle_names` - not being used
* No need of `NAMES_USED_IN_TESTS` anymore!

exercises/robot-name/robot-name.spec.js
SleeplessByte: Yes, but `release_names` was only a function I made in the example so that I could do it. Do we want to force people to implement it?

SleeplessByte:
@tejasbubane I prefer your implementation. The reason I did not do that is just because I did not want to force anything upon anyone :P

BUT if we simply no-op it, it's all good 🗡

exercises/robot-name/robot-name.spec.js
tejasbubane: Good point. We can actually do that. Ruby track [does it](https://github.com/exercism/ruby/blob/master/exercises/robot-name/robot_name_test.rb#L8). Since this is not in the README, we can create a stub file for this exercise and add a blank method called `releaseNames` and also a short comment that this is required for tests.

SleeplessByte: Perfect.

SleeplessByte:
Done @tejasbubane !

PR #598, opened 2019-01-24
[Space-Age] - Change API to take planet as input

tejasbubane:
* Rather than having multiple methods for each planet, just take it as input - makes the solution a lot simpler.

* Also properly adheres to the [canonical tests](https://github.com/exercism/problem-specifications/blob/master/exercises/space-age/canonical-data.json).

* Removes the last extra test to check for reassigning of seconds - which is not in canonical data.

### DIFF
--- a/exercises/space-age/example.js
+++ b/exercises/space-age/example.js
@@ -9,49 +9,9 @@ const EARTH_TO_OTHER_PLANETS = {
   neptune: 164.79132,
 };
 
-export class SpaceAge {
-  constructor(seconds) {
-    this.seconds = seconds;
-  }
+export const age = (planet, seconds) => {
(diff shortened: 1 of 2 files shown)


SleeplessByte:
I'm fine with this but that removes our "prototype based" exercise.

tejasbubane:
> but that removes our "prototype based" exercise.

I did not get what you mean here.

SleeplessByte:
Sorry. Check the mentor notes for this exercise. It will make more sense :D

I'll approve this now

tejasbubane:
> Check the mentor notes for this exercise. It will make more sense :D

Yes. 👍  I had it in mind to change mentor notes while making the change but forgot later.

PR #604, opened 2019-02-08
Add Bash script to facilitate running ESLint

xarxziux:
As per a request on the #toolmakers channel from @SleeplessByte on the Exercism Slack group, I'm submitting a Bash script that hides .eslintignore, runs ESLint on the repo and then restores ESLint.  This is intended to be used by contributors working on issues #480.

This is lightly tested but seems to work well enough as-is.

### DIFF
--- a/bin/lint_all.sh
+++ b/bin/lint_all.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+script_name=$( basename "$0" )
+script_version="0.0.2"
+usage="Usage: ${script_name} [-v|--version|-h|--help]"
+
+# Parse the command line parameters
+while [[ $# -gt 0 ]]
+do
+    param="$1"
+    case "${param}" in
+        -v|--version)
+            echo "${script_name}" version ${script_version}
+            echo "Copyright (C) xarxziux 2018-9"
+            exit 0
+            ;;
+        -h|--help)
+            show_help=1
+            shift
+            ;;
+        *)
+            echo "${script_name}": unknown option "${param}"
+            echo "${usage}"
+            exit 0
+            ;;
+    esac
+done
+
+# Show the help screen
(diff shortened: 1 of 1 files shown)


bin/lint_all.sh
```diff
@@ -0,0 +1,60 @@
...
+usage="Usage: ${script_name} [-v|--version|-h|--help]"
+
+# Parse the command line parameters
+while [[ $# -gt 0 ]]
+do
+    param="$1"
+    case "${param}" in
+        -v|--version)
+            echo "${script_name}" version ${script_version}
+            echo "Copyright (C) xarxziux 2018-9"
```
SleeplessByte: I don't mind giving you attribution, but can't we do a header comment (with your username) with license? This currently is not indicative of the license and discourages modification / extension?

bin/lint_all.sh
```diff
@@ -0,0 +1,60 @@
...
+if [ ! -f .eslintignore ]
+then
+     echo ".eslintignore file not found.  Are you in the root directory?"
+     exit 1
+fi
+
+mv .eslintignore .eslintignore.temp
+npx eslint . > errors.txt 2> /dev/null
+
+# TO-DO - what happens here if Node or ESLint are not installed
```
SleeplessByte: Non zero exit code with error please.

bin/lint_all.sh
xarxziux: This was copy-pasted from another project and doesn't belong here.  I've removed the line.

bin/lint_all.sh
xarxziux: The script captures the exit code from the line, but assumes that any error is being raised by ESLint.  Working out if the error is coming from a failure to run NPX or ESLint is a little trickier.

PR #607, opened 2019-02-09
Fix ESLint errors in Pythagorean Triplet, Spiral Matrix and Sum Of Multiples exercises  

xarxziux:
Fix various minor issues with some example.js file in accordance with issue #480.

Please note that I've cheated here slightly by adding a `eslint-disable-next-line no-use-before-define` comment to the Pythagorean Triplet file.  The solution given has circular references which can't be easily fixed so the alternative was to completely re-write the solution.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,13 +8,10 @@ grains
 list-ops
 nth-prime
 palindrome-products
-pythagorean-triplet
 queen-attack
 rational-numbers
 saddle-points
 secret-handshake
-spiral-matrix
 sublist
-sum-of-multiples
 twelve-days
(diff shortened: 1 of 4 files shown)


.eslintignore
```diff
@@ -8,13 +8,10 @@ grains
...
 nth-prime
 palindrome-products
-pythagorean-triplet
 queen-attack
 rational-numbers
 saddle-points
 secret-handshake
-spiral-matrix
 sublist
-sum-of-multiples
```
SleeplessByte: This PR does not fix sum-of-multiples.

PR #608, opened 2019-02-11
Use exact boolean values in leap tests rather than truthy and falsy

tejasbubane:
I came across a leap solution which returned `"Leap Year"` string in place of true. This works with existing tests which check `truthy`/`falsy`. I think it is better to check for actual values rather than relying on weird coercion.


SleeplessByte:
I agree with you! ❤️

PR #611, opened 2019-02-11
simple-cipher: add test to check that messages longer than the key can be decoded.

paparomeo:
I just come across a solution for the "Simple Cipher" exercise from a student, which could encode messages longer than the key but failed to _decode_ messages longer than the key, and yet, because we haven't a test to cover this, all tests would pass giving an illusion of correctness.

This pull request adds a test to also check that a message longer than the key can be decoded.

Correspondent problem specifications PR: exercism/problem-specifications#1460.

### DIFF
--- a/exercises/simple-cipher/simple-cipher.spec.js
+++ b/exercises/simple-cipher/simple-cipher.spec.js
@@ -116,8 +116,13 @@ describe('Substitution cipher', () => {
     expect(cipher.decode('zabcdefghi')).toEqual('zzzzzzzzzz');
   });
 
-  xtest('can handle messages longer than the key', () => {
+  xtest('can encode messages longer than the key', () => {
     expect(new Cipher('abc').encode('iamapandabear'))
       .toEqual('iboaqcnecbfcr');
+  });
+
+  xtest('can decode messages longer than the key', () => {
+    expect(new Cipher('abc').decode('iboaqcnecbfcr'))
+      .toEqual('iamapandabear');
   });
 });


exercises/simple-cipher/simple-cipher.spec.js
```diff
@@ -116,8 +116,13 @@ describe('Substitution cipher', () => {
...
   });
 
-  xtest('can handle messages longer than the key', () => {
+  xtest('can encode messages longer than the key', () => {
     expect(new Cipher('abc').encode('iamapandabear'))
       .toEqual('iboaqcnecbfcr');
   });
+
+  xtest('can decode messages longer than the key', () => {
+    expect(new Cipher('abc').encode('iboaqcnecbfcr'))
```
SleeplessByte: I think you're trying to decode this.
```suggestion
    expect(new Cipher('abc').decode('iboaqcnecbfcr'))
```

SleeplessByte:
Actually, looking at [the canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/canonical-data.json), you probably took it from there:

```json
{
  "description": "Can handle messages longer than the key",
  "property": "encode",
  "input": {
    "key": "abc",
    "plaintext": "iamapandabear"
  },
  "expected": "iboaqcnecbfcr"
}
```

This expects a plaintext of `"iamapandabear" and expected value of "iboaqcnecbfcr" after encode.

paparomeo:
Silly me. Copy / paste error... 🤦‍♂️. I'll fix as soon as I get home (the invoked method should be `decode` rather than `encode`. Actually you've reminded me that I need to update the test canonical data first / instead. Thanks!

paparomeo:
@SleeplessByte: Fixed! Correspondent problem specifications PR: exercism/problem-specifications#1460.

paparomeo:
exercism/problem-specifications#1460 has been merged @SleeplessByte.

SleeplessByte:
You're the best @paparomeo

junedev:
This was bugging me all the time when mentoring. As I came around to addressing this here I found it was fixed already. Many thanks!

Issue #600, opened 2019-01-24
Checklist: convert export default to named export


SleeplessByte:
Follow up on #436 and the implementation #599 .

- [x] variable-length-quantity
- [x] strain
- [x] palindrome-products
- [x] queen-attack
- [x] sum-of-multiples
- [x] sieve
- [x] hexadecimal

smb26:
hello, can I have this task?

SleeplessByte:
@smb26 sure! Just edit your message and say which one you're fixing :)

smb26:
Thank you!

smb26:
Hey @SleeplessByte, can I do a single pull request with the changes on all the six remaining exercises or should I keep doing separate pull requests?

SleeplessByte:
I prefer separate @smb26 , but you can just keep them all open at the same time.

smb26:
Hello, I have a doubt for the palindrome exercise  - on the example, what is being exported is a generate function that will return two objects: largest and smallest, each with value and factors. 

To change it for a named export, should I split it, ending up with two methods that each will return an object?

sin:
> To change it for a named export, should I split it, ending up with two methods that each will return an object?

I'd say no. It's outside of the scope of converting default export to named export. It would change how the solution works internally, not only how the code is exported. You can see in the [example](https://github.com/exercism/javascript/blob/master/exercises/palindrome-products/example.js) that it uses the same loop to generate both results. These calculations are interlinked and it makes sense to do them in one function. If you think the exercise would be better if split into two separate functions, you can present your arguments, you may be right, but I think it's a separate issue.

smb26:
Honestly, I don't feel like it has something to be split, but since the run-length had a separation between functions, I decided to ask about the best strategy. Thank you @sin

SleeplessByte:
@smb26 Right now it's expecting a single function as default export. When you do this one, I would make it maybe expect `palindromes(opts) => ...`.

```diff
- const Palindromes = (params) => {
+ export const palindromes = (params) => {
  ...
- export default Palindromes;
```

SleeplessByte:
A humongous thanks to @smb26 ! Thank you for making this happen 💟

PR #619, opened 2019-02-18
add named export to sum of multiples

smb26:
see issue #600 

(I'm not 100% sure about the const name)


tejasbubane:
I think `sumOfMultiples` would itself be good enough name for this.

PR #630, opened 2019-03-04
#629 Linked List fix head/tail semantics

### LINKED ISSUE #629: linked-list definition of "head" and "tail"
the notion of "head" and "tail" seem to be backwards in the `example.js`.
I'd expect "back" to correspond to "tail" and "front" to correspond to "head", like [this article](https://hackernoon.com/the-little-guide-of-linked-list-in-javascript-9daf89b63b54).

the example works and tests, but it seems conceptually misleading, imo.

README:

> * `push` (*insert value at back*);

example.js:

```javascript
  push(value) {
    if (this.head) {
      const newHead = new Element(value);
      newHead.next = this.head;
      this.head.prev = newHead;
      this.head = newHead;
    } else {
      this.head = new Element(value);
      this.tail = this.head;
    }
  }
```

monte-hayward:
This PR updates the `example.js` code to be more reasonable. The `head` of the 
list is now also the front, and the `tail` of the list is now also the back. 

This matches the explanation in [this article](https://hackernoon.com/the-little-guide-of-linked-list-in-javascript-9daf89b63b54) and the readme.md of this exercise.

Fixes #629

### DIFF
--- a/exercises/linked-list/example.js
+++ b/exercises/linked-list/example.js
@@ -10,68 +10,68 @@ export class LinkedList {
   }
 
   pop() {
-    if (!this.head) {
+    if (!this.tail) {
       return undefined;
     }
 
-    const { value } = this.head;
-    if (this.head.next) {
-      this.head = this.head.next;
-      this.head.prev = null;
+    const { value } = this.tail;
+    if (this.tail.next) {
+      this.tail = this.tail.next;
+      this.tail.prev = null;
     } else {
-      this.tail = null;
-      this.head = this.tail;
+      this.head = null;
+      this.tail = this.head;
     }
 
     return value;
   }
 
   push(value) {
-    if (this.head) {
-      const newHead = new Element(value);
-      newHead.next = this.head;
-      this.head.prev = newHead;
-      this.head = newHead;
-    } else {
-      this.head = new Element(value);
-      this.tail = this.head;
-    }
-  }
-
-  shift() {
-    if (!this.tail) {
-      return undefined;
-    }
-
-    const { value } = this.tail;
-    if (this.tail.prev) {
-      this.tail = this.tail.prev;
-      this.tail.next = null;
-    } else {
-      this.tail = null;
-      this.head = this.tail;
-    }
-
-    return value;
-  }
-
-  unshift(value) {
     if (this.tail) {
-      const newTail = new Element(value);
-      newTail.prev = this.tail;
-      this.tail.next = newTail;
-      this.tail = newTail;
+      const newtail = new Element(value);
+      newtail.next = this.tail;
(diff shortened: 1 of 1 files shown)


junedev:
@monte-hayward @SleeplessByte I am not 100% sure but I have a feeling this merge request made the example solution incorrect and the same problem exists in the mentoring notes for this exercise.

Let's look at `push` as an example. As explained in the original issue we want to use push to add an element at the end (tail) of the list.
In the old version the implementation of push included the line `this.head.prev = newHead;`.  
This was changed to `this.tail.prev = newtail;`.
But think about that for a moment. We go to the tail, then to the element in front and then replace that one with the new tail. Since our newtail should become the tail I would expect it to be added as `this.tail.next` not `prev`.

To me it seems the tests "accidentally" pass (because push/unshift are exchanged) but the internal representation of the list is incorrect with the current version of the example file.

To illustrate that something is wrong with the list I added a small print method to the example.js file
```
    print() {
        let element = this.head
        while (element) {
            console.log(element.value)
            element = element.next
        }
    }
```
Now I run
```
list.push(1)
list.push(2)
list.push(3)
list.print()
```
If everything would be ok I would expect 1 2 3 as output since I added the values at the end and print them starting from the head. But this is not what happens. I only get 1. 
(You can run it here: https://repl.it/repls/NanoArtisticConditional)

SleeplessByte:
Hi there @junedev,

Thank you for the elaborate answer (including the repl). 

As said on slack: The mentor notes are independent from this solution and should be tackled well... independently. As for the `example.js`. I think the intention was to _reverse_ the `head` and `tail` and therefore your `print` should also _print_ in `reverse` when using `next`. I would expect your print function to print `3` `2` `1`, and it does if I replace line 103 with `let element = this.tail`.

In order to make your print work as you want it too:

```javascript
print() {
  let element = this.head
  while (element) {
    console.log(element.value)
    element = element.prev
  }
}
```

In other words: this solution makes the newly pushed element always the tail. It always points to the previous tail. The head is therefore "stable".  

> `[e,l,e,m,e,n,t,s]=>` (head of the :snake:)
> ```text
>       [1]=>
>    [2, 1]=>
> [3, 2, 1]=>
> ```

junedev:
Thanks for explaining! So the missing piece of information for me was the "stable head" condition. I will take further discussions elsewhere.

SleeplessByte:
@junedev feel free to use #track-javascript, ping me on slack or open an issue.

junedev:
Just in case someone comes by here again: We talked more on Slack and @SleeplessByte agreed that `next` should move you from the head towards the tail and not the other way around.

PR is on the way to fix it.

Issue #421, opened 2018-06-29
Use one, track level, package.json


bdjnk:
I realize this was proposed previously (#235) and closed as `wontfix`, but there is a simple and correct way to have a single install of all the necessary tools. The only necessary `package.json` change would be to the `scripts` section.

Right now we have a `package.json` in each exercise with a `scripts` section as follows:

```json
"scripts": {
  "test": "jest --no-cache ./*",
  "watch": "jest --no-cache --watch ./*",
  "lint": "eslint .",
  "lint-test": "eslint . && jest --no-cache ./* "
},
```

All `package.json` files should be removed from the exercise directories, and a single `package.json` would be added to the track directory. The `scripts` section of it should be changed to the following:

```json
"scripts": {
  "test": "jest --no-cache",
  "watch": "jest --no-cache --watch",
  "lint": "eslint --ignore-pattern '*.spec.js'"
},
```

The usage is changed very slightly.

Running `npm test` or `npm run lint` would test and lint, respectively, all exercises. To test or lint a single particular exercise, simply append `-- <exercise-name>`.

So, for example, if I wanted to test the `simple-cipher` exercise I would run `npm test -- simple-cipher`.

I tested this on MacOS and Linux, and it works correctly. I haven't tested it on Windows, but my understanding is that it should work the same way.

I didn't begin with a pull request, although I'd be happy to do so, because I wanted to get feedback and hopefully buy-in before proceeding.

Thanks for reading. Looking forward to your thoughts.

matthewmorgan:
@bdjnk thanks for reviving this, it would be a welcome change.

If I (dimly) recall, the issue was the behavior of Jest either locally or on Travis.  I seem to remember that it wanted to run _all_ tests in the repository when the user's intent was to run tests for a single exercise.

Any change needs to allow for three scenarios:
1. User tests _one_ exercise only
2. Contributor tests everything locally
3. Travis build runs all tests

If your PR would address this, it's warmly welcomed!

tejasbubane:
More importantly we want the users to receive the `package.json` and `package-lock.json` when they run `exercism fetch`. For this we need to keep these files in each exercise.

matthewmorgan:
Ah yes, @tejasbubane there's the rub.  This works for contributors but not for the end user...I had forgotten about that, I knew there was something blocking this.  

[...]

PR #636, opened 2019-03-17
Chore: eslint per track, loose rules

### LINKED ISSUE #44: Configure ESLint to lint source code at development time

### LINKED ISSUE #502: Lint errors with line endings in windows
Hi everyone,

So I'm new with this project and when I try to run the lint command to solve some errors regarding #480, it shows errors about Expected linebreaks to be 'LF' but found 'CRLF'. I'm running this on windows. Should I change some configuration in the eslint for this?

### LINKED ISSUE #589: Suggestion: prefer eslint syntax over style
With [more](https://github.com/prettier/prettier) and [more](https://github.com/xojs/xo) tools becoming available to automatically format the code base to a certain set of style rules (and [more comprehensive](https://github.com/prettier/prettier-eslint/issues/101) than `eslint --fix`, I'm also more inclined to stop teaching style preference, as indicated by mentor notes, mentor guidance docs and discussions on slack.

## The current state
Currently, the javavscript `package.json` sets a _very restrictive_ code style (`airbnb`) and I *don't* think this:
- helps the student understand the language
- helps the student get fluent in a language
- helps the mentor mentoring as downloading exercises can result in wibbly wobblies all over the place

We don't instruct or enforce these rules (https://github.com/exercism/javascript/issues/44#issuecomment-416760562) strictly, but I seem them in mentoring comments and most IDE's will run these _automagically_ if present.

## Recommendation
I therefore recommend to drop down to [eslint:recommend](eslint:recommended"](https://eslint.org/docs/rules/)) or if we must have a non-company style guide, use [standard](https://github.com/standard/eslint-config-standard) with semicolon rules disabled (people should decide about ASI themselves -- I don't think pushing people who know about ASI helps anyone, and TS doesn't have them either).

If this is something we want, I think doing it sooner rather than later is good and will speed up #480 greatly. The eslintignore list is still quite long and I don't think tháts particularly helpful.

As a final note, I think we should _better_ communicate to the students to run `npm/yarn install` and then `npm/yarn test`. I also suggest running `test` followed by `lint` so that someone is only bugged once the test succeeds, instead of using a `"pretest": "lint"`:

```json
{ 
  "scripts": {
    "test": "test:jest && test:lint",
    "test:jest": "...",
    "test:lint": "..."
   }
}
```

### LINKED ISSUE #620: Checklist: drop eslint for exercises
Per #589,

- [x] move `eslint` config from root `package.json` into root `.eslintrc`
- [x] **remove** all `eslint` dependencies from exercise-level `package.json`
- [x] fix CI check to ignore eslint discrepancy between root and exercise-level `package.json` 
- [x] (optional): add script to add `eslint:recommended` to exercise-level `package.json` on "deploy".

SleeplessByte:
Fixes #620, by adding track-only eslint
Fixes #589, by adopting eslint:recommended for students (track-only)
Fixes #502 (this is handled by git)
Fixes #44 

Workaround for maintainers on #421 by adding a script to copy the package.json to all exercises


bin/sync-config
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env node
+
+/**
+ * This node js executable reads in the top-level package.json, babel.config.js
+ * and the special .eslintrc.track.json and copies them to each exercise.
```
tejasbubane: Can the Makefile do this? Since it already copies `package.json`, `ASSIGNMENT` and `TEST_FILE`.

exercises/accumulate/babel.config.js
```diff
@@ -0,0 +1,13 @@
...
+    {
+      targets: {
+        node: 'current',
+      },
+      useBuiltIns: false,
+    },
+  ],
+];
+
+module.exports = { presets };
```
tejasbubane: Sorry if I am missing something, why do we need to separate babel config from package.json?

SleeplessByte: Because we shouldn't store it in package.json just like we shouldn't store eslint rules in package.json. Either move both out or neither.

bin/sync-config
SleeplessByte: Check the other PR to see where this is going.

SleeplessByte: (so yes, but I rather have ALL functionality in Make or ALL in node/sh, not all over the place. I don't know how to parse json in Make, remove a key and do a comparison)

tejasbubane:
@SleeplessByte I think we should split this into separate PRs for:

1. eslint change at exercise and user levels
1. babel upgrade
1. version keeping while maintaining integerity check

What do you think?

exercises/accumulate/babel.config.js
tejasbubane: https://stedolan.github.io/jq/

bin/sync-config
tejasbubane: https://stedolan.github.io/jq/

SleeplessByte: The point is that we're forcing contributors to learn and install Make when everything else is in JavaScript. We're using npm as package manager and to get the dependencies in but then use another system to run scripts?

SleeplessByte:
@tejasbubane eslint per track without upgrade to new version could work separately; so yes I can do these three if that's what you prefer 👍

tejasbubane:
@SleeplessByte If you are busy (with the track reordering) let me know if I can pick up something from here.

SleeplessByte:
@tejasbubane I have this planned for tomorrow :)

PR #643, opened 2019-03-22
Fix bigint tests

drericrobinson:
For the "Grains" exercise, the BigInt function was not being imported correctly, so the tests wouldn't run.

Also, one of the tests was testing for the .toString() function on BigInt by using coercion, but the test had "===" instead of "==".


SleeplessByte:
The tests ran correctly on travis because it manually copied this over.
@tejasbubane was there a reason for it to be that way?

tejasbubane:
> The tests ran correctly on travis because it manually copied this over.
@SleeplessByte Issue with copying path I guess. I am refactoring Makefile.

PR #644, opened 2019-03-25
Chore: upgrade to babel-7 and move it out of package.json

SleeplessByte:
@tejasbubane this _only_ upgrades babel to version 7 (and updates jest and babel-jest in the same process). 

## Changes

- Upgrade `babel` to version 7+
- Upgrade `jest` to version 24.5+
- Extract babel configuration into separate config file (`babel.config.js`)
- Drop regenerator plugin as LTS node handles it (as recommended by [jest](https://github.com/facebook/jest/blob/e2d80e7144742dd7e34f5cb793d098cceb7816b2/website/blog/2019-01-25-jest-24-refreshing-polished-typescript-friendly.md#other-highlights))
- Drop builtin-extend as it's for babel 6
- Upgrade `babel-jest` to latest to support babel 7

## Fixes

- `queen-attack` and `octal` had a faulty implementation that broke on babel 7

## Additional notes

- the `tools` folder is only used when I'm on windows but shows how we _could_ write scripts like these (devving on a windows machine with make, one without and a linux box). I committed them as we also have tools like `lint_all` that aren't called from make etc and I needed it across machines.


tejasbubane:
@SleeplessByte I have a change to make in Makefile. Do you mind if I push on this same branch?

tejasbubane:
The change is that we need to copy the new babel-config files over while testing.

PR #649, opened 2019-03-26
Promote two-fer to core and sync to 1.2.0

### LINKED ISSUE #639: Curriculum: promote two-fer to core (round 2)
- [ ] sync `two-fer` with the problem specifications, add the `default` argument instead of calling with `''`.
- [ ] promote `two-fer` to core, right after `hello-world`

SleeplessByte:
And make sure we expect an optional argument, instead of passing in null, requiring boolean logic.

Fixes #639

### DIFF
--- a/exercises/two-fer/two-fer.spec.js
+++ b/exercises/two-fer/two-fer.spec.js
@@ -1,4 +1,4 @@
-import { twoFer } from './two-fer.js'
+import { twoFer } from './two-fer'
 
 describe('twoFer()', () => {
   test('no name given', () => {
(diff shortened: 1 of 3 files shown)


exercises/two-fer/two-fer.spec.js
```diff
@@ -1,18 +1,15 @@
-import { twoFer } from './two-fer';
+import { twoFer } from './two-fer.js'
```
tejasbubane: Extension is not required.

SleeplessByte: Oh right! Sorry -- that's automagic :)

tejasbubane:
@SleeplessByte There seems to be a conflict here because of another PR.

SleeplessByte:
I'll fix these conflicts right away

PR #653, opened 2019-03-26
Tweaks to `two-fer` instructions

CristianGiordano:
I've been reviewing some solutions to the exercise and have noticed a lot of users simply writing a solution that checks if the name is Alice.

It appears they are not running the tests or not unskipping the other tests.

### DIFF
--- a/exercises/two-fer/README.md
+++ b/exercises/two-fer/README.md
@@ -6,7 +6,7 @@
 "One for X, one for me."
 ```
 
-When X is a name or "you".
+Where X is a name or "you". For example:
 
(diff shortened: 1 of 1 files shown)


SleeplessByte:
@CristianGiordano This used to be a side-exercise, so most people actually did not get mentored on this.

This change seems okay. You might want to also make it in the `problem-specifications` repo.

CristianGiordano:
Ah yeh, I've just seen this from in the slack channel. Just getting started so am not up to speed with everything. Thanks for the information 👍

PR #654, opened 2019-03-27
Refactor makefile

tejasbubane:
* Break Makefile tasks into smaller chunks for future modification (exercise versioning, etc).

* Fix import of `big-integer` in `grains` example solution file. Not same as #643

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,44 +25,68 @@ BABEL_CONFIG_FILES= $(shell find ./exercises/*/* -maxdepth 1 -name babel.config.
 SOURCE_ESLINT_MD5 ?= "`./bin/md5-hash ./.eslintrc`"
 ESLINTRC_FILES= $(shell find ./exercises/*/* -maxdepth 1 -name .eslintrc)
 
-copy-assignment:
+sync:
 	@cp package.json exercises/$(ASSIGNMENT)
 	@cp babel.config.js exercises/$(ASSIGNMENT)
 	@cp .eslintrc exercises/$(ASSIGNMENT)
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -25,44 +25,68 @@ BABEL_CONFIG_FILES= $(shell find ./exercises/*/* -maxdepth 1 -name babel.config.
...
+test-prepare:
 	@mkdir -p $(OUTDIR)
-	@cp exercises/grains/lib/big-integer.$(FILEEXT) $(OUTDIR)
-	@cp exercises/$(ASSIGNMENT)/$(TSTFILE) $(OUTDIR)
 	@cp exercises/$(ASSIGNMENT)/$(EXAMPLE) $(OUTDIR)/$(subst _,-,$(ASSIGNMENT)).$(FILEEXT)
 	@sed 's/xtest/test/g' exercises/$(ASSIGNMENT)/$(TSTFILE) > tmp_exercises/$(ASSIGNMENT)/$(TSTFILE)
+	@cp package.json $(OUTDIR)
+	@cp babel.config.js $(OUTDIR)
+	@cp .eslintrc $(OUTDIR)
+ifeq ($(ASSIGNMENT),grains)
```
SleeplessByte: It would be more resilient if instead you check if there is a `lib` folder and we always copy over the `lib` folder? That way if we add a new exercise that also needs a library file we don't need to change the makefile.

tejasbubane: Agree, this will be more generic and useful.

tejasbubane: Done

PR #665, opened 2019-04-10
Fix fetch-configlet bash script for windows

SleeplessByte:
Sync with upstream https://github.com/exercism/configlet/pull/156

* This correctly sets the OS and EXTension when on windows, even when in bash
(or a different bourne-shell).

* Additionally this uses unzip if the extension is zip, as tar will correctly say the
zip is not a tar. `unzip` does not allow for piping, so a temporary file is created.

### DIFF
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -10,8 +10,18 @@ case $(uname) in
         echo "linux";;
     (Windows*)
         echo "windows";;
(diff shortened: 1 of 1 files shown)


SleeplessByte:
@tejasbubane given that you're not on Windows and this was approved and merged upstream, I've just went ahead and merged ✔️

PR #676, opened 2019-05-09
Implement D&D Character Exercise

Tyresius92:
Implemented the D&D Character exercise


Tyresius92:
Hmm. Looks like I need to update config.json, and also run Make. I'll switch over to my Linux computer and fix this up. I believe I'll be able to do that this evening.

Tyresius92:
Hey guys. I just ran `make test` and aside from running the tests, the changes it seems to have made is all in the `package.json` files. Is this the expected behavior? 

Additionally, I can take a guess at most of the values in `config.json`, but how do I get the uuid?

tejasbubane:
> Hey guys. I just ran make test and aside from running the tests, the changes it seems to have made is all in the package.json files. Is this the expected behavior?

@Tyresius92 Sorry for the confusion. `make test` copies `package.json` from root of this project to all exercise directories. This is to make sure that all `package.json` files are consistent (with same packages and versions) and we use the same in CI. When you create a new exercise, it is expected you copy the `package.json` from root of this project i.e. [this file](https://github.com/exercism/javascript/blob/master/package.json) and don't make any changes. If changes are required there, please discuss since they will be propagated to all exercises. We have a CI script which ensures all exercise package.json are exactly the same.

> Additionally, I can take a guess at most of the values in config.json, but how do I get the uuid?

Generate a new UUID by any tool of your choice.

SleeplessByte:
> Generate a new UUID by any tool of your choice.

you can use [`configlet uuid`](https://github.com/exercism/configlet#uuid) if you don't want to use something else!

Tyresius92:
Hey guys, just wanted to make sure no further action is required from me on this PR?

exercises/dnd-character/example.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    return diceRolls
+      .sort()
+      .slice(1)
+      .reduce((partialSum, num) => partialSum + num, 0);
+  }
+}
+
+export const ability_modifier = abilityScore => {
```
SleeplessByte: Can you make this `camelCase` as well?

exercises/dnd-character/example.js
```diff
@@ -0,0 +1,36 @@
...
+    this.strength = this.getAbilityScore();
+    this.dexterity = this.getAbilityScore();
+    this.constitution = this.getAbilityScore();
+    this.intelligence = this.getAbilityScore();
+    this.wisdom = this.getAbilityScore();
+    this.charisma = this.getAbilityScore();
+    this.hitpoints = 10 + ability_modifier(this.constitution);
+  }
+
+  getAbilityScore() {
```
SleeplessByte: I think this would be better as _static_ method (class method) on `Character`.

```javascript
export class Character {
  static generateAbility() {
    ...
  }
}
```

exercises/dnd-character/example.js
```diff
@@ -0,0 +1,36 @@
...
+    let diceRolls = [];
+
+    for (var i = 0; i < 4; i++) {
```
SleeplessByte: The example is _not_ the best solution, so it's fine for this not to be perfect, but just in case someone looks, can you make this `let`

exercises/dnd-character/example.js
Tyresius92: I can do that, but then I believe we would need to remove the second suite of tests (lines 82 - 92), or at least the test case on lines 83-86.

SleeplessByte: Change it to:

```javascript
    test('ability within range', () => {
      expect(Character.generateAbility()).toBeLessThanOrEqual(18);
      expect(Character.generateAbility()).toBeGreaterThanOrEqual(3);
    });
```

FYI I picked `generateAbility` here, but feel free to use a _better_ name. I don't think the _`get`_ prefix is right, but `rollAbility` or AbilityScore might be ok.

Tyresius92: Wait, scratch that. I've made the appropriate changes. Changes incoming.

exercises/dnd-character/dnd-character.spec.js
```diff
@@ -0,0 +1,117 @@
...
+  describe('Ability scores calculated properly', () => {
+    test('ability within range', () => {
+      expect(Character.rollAbility()).toBeLessThanOrEqual(18);
+      expect(Character.rollAbility()).toBeGreaterThanOrEqual(3);
+    });
+
+    test('ability scores only calculated once', () => {
+      const Drizzt = new Character();
+
+      expect(Drizzt.strength).toEqual(Drizzt.strength);
```
SleeplessByte: Perhaps we want to test all of them and not just `strength`?

SleeplessByte:
@tejasbubane LGTM, what do you think?

exercises/dnd-character/dnd-character.spec.js
Tyresius92: I thought about that, but canonical-data.json only has the check on strength. I can put in the others as well if you'd like though.

Issue #646, opened 2019-03-25
Makefile vs JS scripts


tejasbubane:
See sample js files in :

**Makefile**:
* Simple elegant bash commands
* Already in place - tiny changes required.
* Needs separate software on windows (like Cygwin etc)

**JS Script**
* Single language everywhere - assuming contributors already have nodeJS installed
* Not as elegant as Makefile - extremely elaborate for simple task like copying a file: [examples](https://github.com/exercism/javascript/tree/d131c88a91a65f2a07d130942fee381f53daf899/bin/tools)

(Please use 👍 to make the change to JS Script and 👎 to stay with Make)

SleeplessByte:
I would also want to add that:

**Makefile:**
- there is _significant_ overhead for new contributors who don't know make and don't use bash (think: using guis a lot, are on windows, or use a different non-bourne style shell) 
- make commands are easy to get wrong cross-platform, even if a windows user has cygwin / bash installed

iHiD:
Sounds good to me.

tejasbubane:
As of now I prefer Makefile for the [simplicity](https://github.com/exercism/javascript/pull/647/files). But my opinion is biased since I don't use windows and makefiles work flawlessly on Unix systems.

Should we investigate any JS libraries that ease out the process and reduce the elaborate handrolled scripts? I found one https://github.com/jakejs/jake but did have a proper look yet.


>  there is _significant_ overhead for new contributors who don't know make and don't use bash 
> (think: using guis a lot, are on windows, or use a different non-bourne style shell)

Makefiles will not be changed by contributors most of the times. Maybe we should add instructions to install Cygwin for windows contributors? 

> make commands are easy to get wrong cross-platform, even if a windows user has cygwin / bash
> installed

I did not know this. Any specific examples?

iHiD:
What's the Makefile actually do here?

SleeplessByte:
> What's the Makefile actually do here?

Copy a few top-level files, create a temporary directory, copy in all the exercise tests, and rename example.js to the correct filename -- then lints and runs tests.

> As of now I prefer Makefile for the simplicity. But my opinion is biased since I don't use windows and makefiles work flawlessly on Unix systems.

I work with both which is why I know the pain. Make is _really_ not straightforward on windows, even with cygwin.

> I did not know this. Any specific examples?

[...]

PR #680, opened 2019-05-20
Use JS scripts instead of Makefile

tejasbubane:
Issue: https://github.com/exercism/javascript/issues/646

* No `make` dependency anymore!

* Uses [shelljs](https://github.com/shelljs/shelljs) for all shell commands. So everything is unix-like without depending on OS commands! `shelljs` gets removed while syncing to exercise dirs and while performing checksum.

* Travis job now uses these JS scripts instead.

* Next step is to streamline process, change all the docs, add contributing doc and delete Makefile.

### DIFF
--- a/scripts/checksum
+++ b/scripts/checksum
@@ -4,14 +4,18 @@ var shell = require('shelljs');
 var helpers = require('./helpers');
 var crypto = require('crypto');
 
+function sha(str) {
+  return crypto.createHash('md5').update(str).digest("hex");
+}
+
 function checksum(filename, assignment, rootChecksum) {
-  if(assignment) {
-    var assignmentConfig = shell.cat('exercises/' + assignment + '/' + filename).toString();
-    var assignmentChecksum = crypto.createHash('md5').update(assignmentConfig).digest("hex");
-    if(assignmentChecksum !== rootChecksum) {
-      shell.echo(filename + ' did not match for ' + assignment);
-      shell.exit(1);
-    }
+  if(!assignment) { return; }
+
+  var assignmentConfig = shell.cat(['exercises', assignment, filename].join('/')).toString();
+  var assignmentChecksum = sha(assignmentConfig);
+  if(assignmentChecksum !== rootChecksum) {
+    shell.echo(filename + ' did not match for ' + assignment);
+    shell.exit(1);
   }
 }
 
@@ -19,7 +23,7 @@ function checkSumAll(filename, rootFileName = filename) {
   shell.echo('\nChecking integrity of ' + filename + '...');
 
   var config = shell.cat(rootFileName).toString();
-  var rootChecksum = crypto.createHash('md5').update(config).digest("hex");
+  var rootChecksum = sha(config);
   var assignment = shell.env['ASSIGNMENT'];
   var assignments = helpers.assignments;
 
--- a/scripts/checksum
+++ b/scripts/checksum
@@ -1,0 +1,41 @@
+#!/usr/bin/env node
+
+var shell = require('shelljs');
+var helpers = require('./helpers');
+var crypto = require('crypto');
+
+function checksum(filename, assignment, rootChecksum) {
+  if(assignment) {
+    var assignmentConfig = shell.cat('exercises/' + assignment + '/' + filename).toString();
+    var assignmentChecksum = crypto.createHash('md5').update(assignmentConfig).digest("hex");
+    if(assignmentChecksum !== rootChecksum) {
+      shell.echo(filename + ' did not match for ' + assignment);
+      shell.exit(1);
+    }
+  }
+}
+
+function checkSumAll(filename, rootFileName = filename) {
+  shell.echo('\nChecking integrity of ' + filename + '...');
+
+  var config = shell.cat(rootFileName).toString();
+  var rootChecksum = crypto.createHash('md5').update(config).digest("hex");
+  var assignment = shell.env['ASSIGNMENT'];
+  var assignments = helpers.assignments;
+
+  if(assignment) {
+    checksum(filename, assignment, rootChecksum);
+  }
+  else {
+    for(i = 0; i < assignments.length; i++) {
+      checksum(filename, assignments[i], rootChecksum);
+    }
+  }
+}
+
+helpers.createExercisePackageJson();
+checkSumAll('package.json', 'exercise-package.json');
+shell.rm('exercise-package.json');
+
+checkSumAll('.eslintrc');
+checkSumAll('babel.config.js');
--- a/scripts/ci
+++ b/scripts/ci
@@ -1,16 +1,33 @@
 #!/usr/bin/env node
 
+/**
+ * Run this script (from root directory): node scripts/ci
+ *
+ * This will run following checks:
+ * 1. Check config in all exercises matches
+ * 2. Run eslint to check code-style
+ * 3. Run tests against sample solutions
+ */
+
 var shell = require('shelljs');
 var helpers = require('./helpers');
 
 var checkResult = shell.exec('scripts/checksum').code;
-if(checkResult != 0) shell.exit(checkResult);
+if(checkResult != 0) { shell.exit(checkResult); }
 
 // Cleanup tmp directory if any exists
 helpers.cleanUp();
 
+/**
+ * PREPARE=true moves all example and test files to single directory - tmp_exercises
+ * This allows running the test command together for all files
+ * which is way faster than running once for each exercise
+ */
 var lintResult = shell.exec('PREPARE=true scripts/lint').code;
-if(lintResult != 0) shell.exit(lintResult);
+if(lintResult != 0) { shell.exit(lintResult); }
 
+/**
+ * CLEANUP=true removes tmp_exercises folder
+ */
 var testResult = shell.exec('CLEANUP=true scripts/test').code;
-if(testResult != 0) shell.exit(testResult);
+if(testResult != 0) { shell.exit(testResult); }
--- a/scripts/ci
+++ b/scripts/ci
@@ -1,0 +1,16 @@
+#!/usr/bin/env node
+
+var shell = require('shelljs');
+var helpers = require('./helpers');
+
+var checkResult = shell.exec('scripts/checksum').code;
+if(checkResult != 0) shell.exit(checkResult);
+
+// Cleanup tmp directory if any exists
+helpers.cleanUp();
+
+var lintResult = shell.exec('PREPARE=true scripts/lint').code;
+if(lintResult != 0) shell.exit(lintResult);
+
+var testResult = shell.exec('CLEANUP=true scripts/test').code;
+if(testResult != 0) shell.exit(testResult);
--- a/scripts/helpers.js
+++ b/scripts/helpers.js
@@ -1,0 +1,75 @@
+// Not intended to be run as a script
+
+var shell = require('shelljs');
+
+var assignments = [];
+var exerciseDirs = shell.ls('-d', 'exercises/*');
+for(i = 0; i < exerciseDirs.length; i++) {
+  assignments.push(exerciseDirs[i].split('/')[1]);
+}
+
+function prepare(assignment) {
+  if(!assignment) {
+    shell.echo('assignment not provided');
+    shell.exit(1);
+  }
+  var exampleFile = 'exercises/' + assignment + '/example.js';
+  var specFile = 'exercises/' + assignment + '/' + assignment + '.spec.js'
+
+  var inDir = 'exercises/' + assignment;
+  var outDir = 'tmp_exercises/' + assignment;
+
+  shell.mkdir('-p', 'tmp_exercises/lib');
+  shell.cp(exampleFile, 'tmp_exercises/' + assignment + '.js')
+  shell.sed('xtest', 'test', specFile)
+    .to('tmp_exercises/' + assignment + '.spec.js');
+
+  var libDir = 'exercises/' + assignment + '/lib/';
+  if(shell.test('-d', libDir)) {
+    shell.cp(libDir + '*.js', 'tmp_exercises/lib');
+  }
+}
+
+function prepareAndRun(command, infoStr, failureStr) {
+  if(shell.env['PREPARE']) {
+    var assignment = shell.env['ASSIGNMENT'];
+
+    if(assignment) {
+      prepare(assignment);
+    }
+    else {
+      for(i = 0; i < assignments.length; i++) {
+        prepare(assignments[i]);
+      }
+    }
+  }
+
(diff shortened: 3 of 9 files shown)


scripts/checksum
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
+
+var shell = require('shelljs');
```
SleeplessByte: Is there a reason we're using `var`?

scripts/checksum
```diff
@@ -0,0 +1,41 @@
...
+var helpers = require('./helpers');
+var crypto = require('crypto');
+
+function checksum(filename, assignment, rootChecksum) {
+  if(assignment) {
```
SleeplessByte: I prefer to exit early:

```javascript
function checksum(filename, assignment, rootChecksum) {
  if(!assignment) {
    return
  }
  
  ...
```

It keeps the indentation low (and therefore lowers cognitive complexity) and shows the intent instantly: If there is no assignment, don't do anything.

scripts/checksum
```diff
@@ -0,0 +1,41 @@
...
+    var assignmentConfig = shell.cat('exercises/' + assignment + '/' + filename).toString();
```
SleeplessByte: Can you use `[path, element].join('/')` or a string template instead? It's quite easy to read this incorrectly, and difficult to see where the boundaries are.

In this case I prefer: `['exercises', assignment, filename].join('/')`

scripts/checksum
```diff
@@ -0,0 +1,41 @@
...
+      shell.exit(1);
+    }
+  }
+}
+
+function checkSumAll(filename, rootFileName = filename) {
+  shell.echo('\nChecking integrity of ' + filename + '...');
+
+  var config = shell.cat(rootFileName).toString();
+  var rootChecksum = crypto.createHash('md5').update(config).digest("hex");
```
SleeplessByte: You might want to extract the two checksumers into their own function so it's easier to replace everywhere.

scripts/ci
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env node
+
+var shell = require('shelljs');
+var helpers = require('./helpers');
+
+var checkResult = shell.exec('scripts/checksum').code;
+if(checkResult != 0) shell.exit(checkResult);
```
SleeplessByte: Add brackets `{` `}`, so we don't run into bugs if we need to add more code.

scripts/ci
```diff
@@ -0,0 +1,16 @@
...
+
+// Cleanup tmp directory if any exists
+helpers.cleanUp();
+
+var lintResult = shell.exec('PREPARE=true scripts/lint').code;
+if(lintResult != 0) shell.exit(lintResult);
```
SleeplessByte: Same as above

scripts/ci
```diff
@@ -0,0 +1,16 @@
...
+
+var testResult = shell.exec('CLEANUP=true scripts/test').code;
+if(testResult != 0) shell.exit(testResult);
```
SleeplessByte: Same as above

scripts/ci
```diff
@@ -0,0 +1,16 @@
...
+var checkResult = shell.exec('scripts/checksum').code;
+if(checkResult != 0) shell.exit(checkResult);
+
+// Cleanup tmp directory if any exists
+helpers.cleanUp();
+
+var lintResult = shell.exec('PREPARE=true scripts/lint').code;
+if(lintResult != 0) shell.exit(lintResult);
+
+var testResult = shell.exec('CLEANUP=true scripts/test').code;
```
SleeplessByte: Can you document `CLEANUP`?

scripts/ci
```diff
@@ -0,0 +1,16 @@
...
+var shell = require('shelljs');
+var helpers = require('./helpers');
+
+var checkResult = shell.exec('scripts/checksum').code;
+if(checkResult != 0) shell.exit(checkResult);
+
+// Cleanup tmp directory if any exists
+helpers.cleanUp();
+
+var lintResult = shell.exec('PREPARE=true scripts/lint').code;
```
SleeplessByte: Can you document `PREPARE`?

scripts/checksum
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
```
SleeplessByte: Could probably use a "here is how you run this and what it does"-style comment

scripts/ci
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env node
```
SleeplessByte: Could probably use a "here is how you run this and what it does"-style comment

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
+// Not intended to be run as a script
+
+var shell = require('shelljs');
+
+var assignments = [];
+var exerciseDirs = shell.ls('-d', 'exercises/*');
```
SleeplessByte: Not a fan of a forced side-effect when loading this script. Can we move this to a function? Memoize it if necessary.

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+for(i = 0; i < exerciseDirs.length; i++) {
+  assignments.push(exerciseDirs[i].split('/')[1]);
+}
+
+function prepare(assignment) {
```
SleeplessByte: Can you document this?

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  shell.sed('xtest', 'test', specFile)
+    .to('tmp_exercises/' + assignment + '.spec.js');
+
+  var libDir = 'exercises/' + assignment + '/lib/';
+  if(shell.test('-d', libDir)) {
+    shell.cp(libDir + '*.js', 'tmp_exercises/lib');
+  }
+}
+
+function prepareAndRun(command, infoStr, failureStr) {
```
SleeplessByte: Can you document this?

scripts/checksum
```diff
@@ -0,0 +1,41 @@
...
+    var assignmentConfig = shell.cat('exercises/' + assignment + '/' + filename).toString();
+    var assignmentChecksum = crypto.createHash('md5').update(assignmentConfig).digest("hex");
+    if(assignmentChecksum !== rootChecksum) {
+      shell.echo(filename + ' did not match for ' + assignment);
+      shell.exit(1);
+    }
+  }
+}
+
+function checkSumAll(filename, rootFileName = filename) {
```
SleeplessByte: Can you document this?

scripts/checksum
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
+
+var shell = require('shelljs');
+var helpers = require('./helpers');
+var crypto = require('crypto');
+
+function checksum(filename, assignment, rootChecksum) {
```
SleeplessByte: Can you document this?

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+for(i = 0; i < exerciseDirs.length; i++) {
+  assignments.push(exerciseDirs[i].split('/')[1]);
+}
+
+function prepare(assignment) {
+  if(!assignment) {
+    shell.echo('assignment not provided');
+    shell.exit(1);
+  }
+  var exampleFile = 'exercises/' + assignment + '/example.js';
```
SleeplessByte: Same as before regarding path joining

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  var specFile = 'exercises/' + assignment + '/' + assignment + '.spec.js'
```
SleeplessByte: Same as before regarding path joining

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+
+  var inDir = 'exercises/' + assignment;
```
SleeplessByte: Same as before regarding path joining

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  var outDir = 'tmp_exercises/' + assignment;
```
SleeplessByte: Same as before regarding path joining

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+
+  shell.mkdir('-p', 'tmp_exercises/lib');
+  shell.cp(exampleFile, 'tmp_exercises/' + assignment + '.js')
```
SleeplessByte: Same as before regarding path joining; in this case I probably would use a string template literal

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  shell.sed('xtest', 'test', specFile)
+    .to('tmp_exercises/' + assignment + '.spec.js');
```
SleeplessByte: Same as before regarding path joining

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+
+  var libDir = 'exercises/' + assignment + '/lib/';
```
SleeplessByte: Same as before regarding path joining

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+      prepare(assignment);
+    }
+    else {
+      for(i = 0; i < assignments.length; i++) {
+        prepare(assignments[i]);
+      }
+    }
+  }
+
+  if(infoStr) shell.echo(infoStr);
```
SleeplessByte: Add brackets

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  var result = shell.exec(command);
+
+  if(shell.env['CLEANUP']) cleanUp();
```
SleeplessByte: Add brackets

scripts/helpers.js
```diff
@@ -0,0 +1,75 @@
...
+
+  if(result.code !== 0) {
+    if(failureStr) shell.echo(failureStr);
```
SleeplessByte: Add brackets

scripts/sync
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env node
```
SleeplessByte: Could probably use a "here is how you run this and what it does"-style comment

scripts/lint
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env node
```
SleeplessByte: Could probably use a "here is how you run this and what it does"-style comment

scripts/test
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env node
```
SleeplessByte: Could probably use a "here is how you run this and what it does"-style comment

scripts/checksum
tejasbubane: These scripts do not go through babel. I wanted to make sure there are no problems with older versions of node (if any). Hence all scripts use ES5 syntax - `function`, `var`, `for` loops etc.

tejasbubane: In near future I will enable test matrix and run tests on multiple node versions.

scripts/helpers.js
tejasbubane: > Not a fan of a forced side-effect when loading this script.

Neither am I. But if we make it a function, we have numerous calls like `assignments()[i]` which are not pleasing to the eye. Also since this is a one-time running script, I think this tradeoff is bearable.

> Memoize it if necessary.

Memoization is a side-effect :)

scripts/checksum
SleeplessByte: I don't think this is a good reason. I would suggest linking up with [node.green](https://node.green/) and target LTS (10.15.3) and above, meaning you can use a _whole lot more_ than ES5. 

The reason for this is that we don't want to set precedent of now being okay with forcefully using old (and even deprecated) modules and usages. Once they're in there, it's another hurldle to get out. The lowest node for this track is LTS anyway, so this shouldn't be a current problem.

tejasbubane: > The lowest node for this track is LTS anyway.

This is not something to feel great about IMO. Most users might have older versions of Node installed (work projects etc. - lot of them might not be aware of `nvm`) and we have to make sure the exercises & tests work for them - hence the test matrix.

The best solution will be to use babel over these scripts as well - which I assume will get best of both worlds - keep our code on shiny latest ES and make sure it runs on old node versions.

tejasbubane:
Fixed all review comments apart from those two (left unresolved). Let me know your thoughts & I can convert to ES syntax and use `babel-cli`.

scripts/checksum
SleeplessByte: Ok. I think running babel on these scripts is a compromise I can definitely live with, especially since it can target the _current user node_, so it will NOT generate `var` if on LTS :).

The "most users are on older versions" argument is what's causing security issues in government software still targeting windows 95 and I think not one we should uphold for _maintaining_ or _contributing_ to this repository. It means that we're okay with users running versions with security holes, issues, deprecated or removed functionality. But that's the last I'll say about it. If our endgoal is as you say: babel + test matrix, I am sufficiently happy ❤️

scripts/helpers.js
SleeplessByte: It means I _must_ have that directory ready prior to executing any dependants (which may be a sore source of bugs). If the issue is the numerous calls I rather have one of two:
- in the _executing_ scripts use: `const assignments = readAssignments()` or something like that, once.
- export default someObject where you've done `Object.defineProperty('assignments'` and it's a getter, so you can have the syntax you want 🗡 

> Memoization is a side-effect :)

 But not forced until you call the function -- so not an import side-effect and keeps the tree shakable.

That said, I'm fine if you leave it. I understand your reasoning. I also don't like assignments()[i], which is why I would load it into a local var first.

scripts/lint
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Run this script (from root directory): node scripts/lint
```
SleeplessByte: Doesn't it allow passing `ASSIGNMENT=` ?

scripts/sync
SleeplessByte: Was this resolved by accident?

scripts/test
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Run this script (from root directory): node scripts/test
```
SleeplessByte: Doesn't it allow passing `ASSIGNMENT=` ?

scripts/checksum
tejasbubane: We never mention leave alone ask users to use old platforms. But we should also make sure our software doesn't break on them. Hopefully this will be good for new-comers who don't get confused and discouraged by unrelated errors.

SleeplessByte: > But we should also make sure our software doesn't break on them.

Don't agree, but I think that's where our opinion differs :)

SleeplessByte:
Let's just get this merged. We can edit later.

PR #684, opened 2019-06-04
Add @types packages

G-Rath:
This adds `@types/jest` & `@types/node` to all exercises; `@types` packages are commonly by tooling to improve intellisense, autocompletion, and other such features.

### DIFF
--- a/exercises/accumulate/package.json
+++ b/exercises/accumulate/package.json
@@ -12,6 +12,8 @@
     "@babel/cli": "^7.2.3",
     "@babel/core": "^7.4.0",
     "@babel/preset-env": "^7.4.2",
+    "@types/jest": "^24.0.13",
+    "@types/node": "^12.0.4",
     "babel-eslint": "^10.0.1",
     "babel-jest": "^24.5.0",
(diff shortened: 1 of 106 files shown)


SleeplessByte:
Thank you @G-Rath !

G-Rath:
Hey, it's just what I do 😉

tejasbubane:
I am not sure if we should add `typescript` packages in `javascript` track 🤔 @SleeplessByte

G-Rath:
@SleeplessByte *cough cough* called it.

@tejasbubane The justification is that some IDEs take advantage of `@types` packages when they're installed:
* VSCode does this in the background whenever it's dealing with JS, meaning you're already downloading `@types/jest` for these projects.
* WebStorm will use them if they're; otherwise you can manually add `@types` packages externally, but you have to do this manually per project.
* Atom I hope would use them, but I have no idea :joy:

SleeplessByte:
> I am not sure if we should add typescript packages in javascript track 🤔 @SleeplessByte

They are dev dependencies which IDE's use to autocomplete. I think they belong in the JS track.

Issue #679, opened 2019-05-20
Add stub files


tejasbubane:
Taken from #678 

Exports required in exercise files can be quite confusing for beginners. We would like to make it easier by adding stub files for `core` exercises with difficulty level `less than or equal to 6`. 

See [config file](https://github.com/exercism/javascript/blob/master/config.json) for core and difficulty levels. `hello-world` exercise has an [example stub file](https://github.com/exercism/javascript/blob/master/exercises/hello-world/hello-world.js).

Note: The filename should be exactly `<exercise-name>.js`. eg. For `two-fer` exercise, there should be a file `exercises/two-fer/two-fer.js` with the following contents:

```js
//
// This is only a SKELETON file for the 'Two Fer' exercise. It's been provided as a
// convenience to get you started writing code faster.
//

export const twoFer = () => {
  //
  // YOUR CODE GOES HERE
  //
};
``` 

`twoFer` here is the function name as required by the [test cases](https://github.com/exercism/javascript/blob/master/exercises/two-fer/two-fer.spec.js).

Feel free to comment below if you need any help.

ankorGH:
@tejasbubane Is someone working on this? If not I will like to work on it.

tejasbubane:
@ankorGH Nobody else is working on this. Go ahead.

SleeplessByte:
Closed via https://github.com/exercism/javascript/commit/189a99767ebe2271eff2b4a784539cec9334dea5

PR #686, opened 2019-06-09
Add stub files for exercises

ankorGH:
Inspired from #679 

Add stub files to all exercises within the difficulty 1 - 6.

cc @tejasbubane

### DIFF
--- a/exercises/accumulate/accumulate.js
+++ b/exercises/accumulate/accumulate.js
@@ -1,0 +1,8 @@
+//
+// This is only a SKELETON file for the 'Accumulate' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+export const accumulate = () => {
+  throw new Error("Remove this statement and implement this function");
+};
--- a/exercises/acronym/acronym.js
+++ b/exercises/acronym/acronym.js
@@ -1,0 +1,8 @@
+//
+// This is only a SKELETON file for the 'Acronym' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+export const parse = () => {
+  throw new Error("Remove this statement and implement this function");
+};
--- a/exercises/all-your-base/all-your-base.js
+++ b/exercises/all-your-base/all-your-base.js
@@ -1,0 +1,8 @@
+//
+// This is only a SKELETON file for the 'All Your Base' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+export const convert = () => {
+  throw new Error("Remove this statement and implement this function");
+};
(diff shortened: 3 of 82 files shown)


SleeplessByte:
This is amazing @ankorGH ! Thank you very much.

Personally, and this is not your fault, I'm not a fan of the repeated `// your code goes here` blocks, which I think is really only necessary for the first (`two-fer`) exercise, and is especially ugly when it's multiple times in a file:

```javascript
class Allergies {
  constructor() {
    //
    // YOUR CODE GOES HERE
    //
  }

  list() {
    //
    // YOUR CODE GOES HERE
    //
  }

  allergicTo() {
    //
    // YOUR CODE GOES HERE
    //
  }
}
```

vs

```javascript
class Allergies {
  constructor() {

  }

  list() {

  }

  allergicTo() {

  }
}
```

Perhaps we'd like to `throw` a not implemented exception instead. (We should check Java and C# stubs, see if they have that or no).

```javascript
export class Allergies {
  list() {
    throw new Error('Not implemented. Implement this method and remove this error.')
  }
 // etc.
}
```

tejasbubane:
This is partly my bad for pasting existing `hello-world` stub in the issue description, modified for `two-fer` without giving much thought. Now that I think of it, I agree with all suggestions made by @SleeplessByte. 👍 

[java](https://github.com/exercism/java/blob/master/exercises/two-fer/src/main/java/Twofer.java) and [csharp](https://github.com/exercism/csharp/blob/master/exercises/two-fer/TwoFer.cs) tracks have errors. And I think we should have them too. Can you make the changes please? @ankorGH

ankorGH:
@SleeplessByte I am currently looking into all the suggestions. Will make changes ASAP.

SleeplessByte:
FYI @ankorGH, since @tejasbubane and I agree, can you change them all (except for `two-fer`) to throw an error?

```javascript
throw new Error('Remove this statement and implement this function')

// e.g.
export class Allergies {
  constructor() {
    throw new Error('Remove this statement and implement this function')
  }
 
  list() {
    throw new Error('Remove this statement and implement this function')
  }

  allergicTo() {
    throw new Error('Remove this statement and implement this function')
  }
}
```

@tejasbubane what do you think of this copy?

ankorGH:
I see [csharp](https://github.com/exercism/csharp/blob/master/exercises/allergies/Allergies.cs) have default  parameter names. Should I add default parameter names or allow the student to name them.

tejasbubane:
I think we should allow the student to name them. What do you think @SleeplessByte ?

SleeplessByte:
> I think we should allow the student to name them. What do you think @SleeplessByte ?

Agreed

ankorGH:
Done with the requested changes.
cc @SleeplessByte @tejasbubane

Issue #628, opened 2019-03-04
Create exercise versioning policy based on Haskell track


tejasbubane:
Exercism tracking issue: https://github.com/exercism/exercism/issues/4786
- haskell: https://github.com/exercism/haskell/pull/717/files#diff-c6795c529069232683d979cd4239c219R2
- ocaml: https://github.com/exercism/ocaml/issues/283
- javascript: https://github.com/exercism/javascript/issues/628

**Note**: This is not the same as https://github.com/exercism/exercism/issues/4266 which removed tracking versions from the user code.

## Proposed implementation for JavaScript

In `package.json` change the `version` number to match the `x.y.z` of [test data versioning](https://github.com/exercism/problem-specifications#test-data-versioning), and add `.q` which we always increment if a change is made without a sync (new test data): `x.y.z.q`.

tejasbubane:
@SleeplessByte For now I suggest not adding extra `.q` and matching with test data versions. This is ofcourse assuming that our test will match test data exactly and we won't be needing any extra tests.

tejasbubane:
Sorry for not being clear with the issue description. Better late than never if anyone else ends up here.

The issue here is not all exercises in this track are in sync with problem specifications. Some (if not most) lag behind and miss some tests or have old version of tests. We need to take this case by case basis and do the following:

* If all tests match, just update version number and we should be good.
* If tests don't match, make the changes required to match them and then update the version number.

I would expect separate PRs with a few exercises - not more than 3-4 in one PR so that it is easy to review.

SleeplessByte:
@tejasbubane yep! I'll make a checklist too.

smb26:
Can I have it please? 😄

SleeplessByte:
@smb26 ❤️ of course. Could you start by opening another issue with at least the ones you're doing but preferably all the exercises as a checkbox (like I did with  https://github.com/exercism/javascript/issues/600) and name it:

```markdown
#Checklist: add exercise versioning
```

Note the steps are listed here: https://github.com/exercism/javascript/issues/628#issuecomment-469569417

SleeplessByte:
@tejasbubane any update on the makefile change to remove the version / replace it with 0.0.0 when doing the integrity check (when running `make test-travis` as well as merging the current version into the copy when running `make`?

sshine:
[...]

PR #687, opened 2019-06-12
Convert scripts to ES6

tejasbubane:
Use [@babel/node](https://babeljs.io/docs/en/babel-node) to run them.

Also fix comments in scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,18 @@
 ---
 language: node_js
 node_js:
-- 'lts/*'
+  - 'lts/*'
 sudo: false
+
 before_install:
-- npm i -g npm@latest
+  - npm i -g npm@latest
+
 install:
-- npm ci
+  - npm ci
+
+before_script:
+  - bin/fetch-configlet
+
 script:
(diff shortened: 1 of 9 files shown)


tejasbubane:
> Any reason you changed from function declaration to "assignment pattern" + arrow function expression? I know it's mostly a stylistic choice, but I think the explicit `function` reads better

Nothing specific. My personal preference of the shiny arrow function over normal `function`. I will change it back to `function`.

>  if you're not re-assigning variables, can you use `const` instead of `let`?

Makes sense. 👍

SleeplessByte:
I don't mind the `const x = () => {}`, but I think it helps our junior contributors if it's _more straightforward_ 🗡  I generally use `function` at the top level, and expression everywhere else.

tejasbubane:
@SleeplessByte Fixed.

SleeplessByte:
I think this means we can now do https://github.com/exercism/javascript/issues/628 ?

tejasbubane:
@SleeplessByte Yup! https://github.com/exercism/javascript/pull/688

PR #690, opened 2019-06-20
Configure travis CI matrix

tejasbubane:
Test on multiple node versions to ensure nothing breaks for end-users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
 ---
 language: node_js
 node_js:
-  - 'lts/*'
+  - '8'
+  - '9'
+  - '10'
+  - '11'
+  - '12'
 sudo: false
 
 before_install:


tejasbubane:
Our code fails on node [v6](https://travis-ci.org/exercism/javascript/jobs/548130402) and [v7](https://travis-ci.org/exercism/javascript/jobs/548130403). Relaxing the min node version to `v8`.

Anyways @SleeplessByte is of the opinion to support only latest LTS.

SleeplessByte:
As long as we **COULD** support multiple versions, I'm all for it 💁; I do think we **SHOULD** only support latest LTS.

tejasbubane:
I think this is fine to be merged now.

There was a mention of LTS in our [installation page](https://exercism.io/tracks/javascript/installation) before I feel. No able to find it now. I think we should add it. cc @SleeplessByte

SleeplessByte:
@tejasbubane yes, let's get that (back) in.

tejasbubane:
> But we might want to consider button v8 in `allow_failures`.

As of now the built succeeds. In future if we use some API unsupported by v8, we will consider moving it.

PR #697, opened 2019-07-09
Change exercise tests to be TDD orientated.

whatcoda:
I have rearranged the tests for this exercise to be more test driven development orientated as we would potentially want the easier test cases to come first so the user doing the exercise can write code to fail fast (eg. negative coins, 0 coins, etc) then build up to the difficult scenarios of using Lilliputian and Lower Elbonia Coins.

This is just a suggestion as I found the exercises as a great way of doing TDD - albeit the tests having already been wrote!

### DIFF
--- a/exercises/change/change.spec.js
+++ b/exercises/change/change.spec.js
@@ -1,7 +1,27 @@
 import { Change } from './change';
 
 describe('Change', () => {
-  test('test change for 1 cent', () => {
+  test('test no coins make 0 change', () => {
+    const change = new Change();
+    const result = change.calculate([1, 5, 10, 21, 25], 0);
+    expect(result).toEqual([]);
+  });
+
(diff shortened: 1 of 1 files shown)


SleeplessByte:
Thanks! Would you also be willing on making this change on the problem-spec repository?

https://github.com/exercism/problem-specifications/

I want to stay close to the canonical when we can, and it seems that this change might benefit a lot of tracks.

whatcoda:
Sure, I've submitted the following PR that reflects the updated ordering: https://github.com/exercism/problem-specifications/pull/1563

SleeplessByte:
Since we're now using `test.toml` and are free to arrange/re-arrange the tests to our order, I'm going to close this in favour of a potential future new PR that just re-orders the tests in our repo only.

PR #698, opened 2019-07-15
simple-cipher: Replace "List" by "Cipher" for the class name

pawamoy:
As imported by the spec.js file, the class name should be `Cipher`, not `List`.


SleeplessByte:
You're absolutely right! Thank you :heart:.

PR #699, opened 2019-07-15
simple-cipher: Update tests according to the spec

pawamoy:
EDIT: at first, this PR just fixed a typo. In the end we updated the test file to match the spec file mentioned in the next comment.

### DIFF
--- a/exercises/simple-cipher/package.json
+++ b/exercises/simple-cipher/package.json
@@ -1,6 +1,6 @@
 {
   "name": "exercism-javascript",
-  "version": "0.0.0",
+  "version": "2.0.0",
   "description": "Exercism exercises in Javascript.",
   "author": "Katrina Owen",
   "private": true,
--- a/exercises/simple-cipher/simple-cipher.spec.js
+++ b/exercises/simple-cipher/simple-cipher.spec.js
@@ -14,7 +14,7 @@ describe('Random key generation', () => {
 describe('Random key cipher', () => {
   const cipher = new Cipher();
 
-  test('has a key made of letters', () => {
+  test('has a key made only of lowercase letters', () => {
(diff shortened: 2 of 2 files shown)


SleeplessByte:
You're absolutely right. It should actually be updated according to https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/canonical-data.json#L81.

Would you be able to do that? (And update the other test-cases as well)? You can then also make the `package.json` version be `2.0.0`, as long as all the tests match ones from that problem-description repo.

pawamoy:
I can do that. I compared the problem specification and the test file, and there are test cases in the test file that are not in the specification file (`generates keys at random` and `has a key that is at least 100 characters long`, and the whole section about throwing errors when the key is incorrect). I guess I do not have to change the specification, only make sure the actual test cases have the same descriptions?

SleeplessByte:
> , and the whole section about throwing errors when the key is incorrect

yes, they were removed!

> I guess I do not have to change the specification, only make sure the actual test cases have the same descriptions?

Correct! Sync OUR tests with THAT canonical data 🗡

pawamoy:
Alright! I'll remove the extra tests that do not appear in the problem spec then!

SleeplessByte:
The generator that I'm still working on will by default copy paste whatever is in there. 

In this case, I actually don't feel that the `ie. jdjdjdjdjd` stuff adds much value. I would _suggest_ you keep the lowercase (for now) and if you want add the `ie. jdjdjdjdjd` as a comment above/inside the test :)

pawamoy:
Done. I think I've implemented all your comments now :slightly_smiling_face: Let me know if there's anything else!

SleeplessByte:
You're the best! @pawamoy

PR #702, opened 2019-07-30
[Housekeeping] Update packages to latest versions

tejasbubane:
This also fixes all `npm audit` issues.

### DIFF
--- a/exercises/acronym/package.json
+++ b/exercises/acronym/package.json
@@ -1,6 +1,5 @@
 {
   "name": "exercism-javascript",
-  "version": "0.0.0",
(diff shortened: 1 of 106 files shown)


exercises/acronym/package.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "exercism-javascript",
-  "version": "0.0.0",
```
tejasbubane: These changes are because the script deletes `version` from copied files [here](https://github.com/exercism/javascript/blob/032fdbd4b7868757a8f1123861e144b433b5d2b7/scripts/helpers.js#L49).

PR #707, opened 2019-08-12
Test that callbacks with static values still fire

yyyc514:
I had a student misunderstand the callback system and **their callbacks would always run** but _only log values_ if the *callback* data had changed. The tests still pass since all the test callbacks mirror the cells data (ie, always changing when the cell changes).

A single test that had static data would have failed.

Can we fix this somehow?

### DIFF
--- a/exercises/react/react.spec.js
+++ b/exercises/react/react.spec.js
@@ -98,6 +98,24 @@ describe('React module', () => {
     expect(callback.values).toEqual([222]);
   });
 
+  xtest('static callbacks fire even if their own value has not changed', () => {
+    const inputCell = new InputCell(1);
+    const output = new ComputeCell(
+      [inputCell],
+      inputs => (inputs[0].value < 3 ? 111 : 222),
+    );
+
+    const callback = new CallbackCell(() => "cell changed");
+    output.addCallback(callback);
+
+    inputCell.setValue(2);
+    expect(callback.values).toEqual([]);
+
+    inputCell.setValue(4);
+    inputCell.setValue(2);
+    inputCell.setValue(4);
+    expect(callback.values).toEqual(["cell changed","cell changed","cell changed"]);
+  });
 
   xtest('callbacks can be added and removed', () => {
     const inputCell = new InputCell(1);


yyyc514:
Open to other suggestions.

What do I need to do to make the linter happy?

exercises/react/react.spec.js
```diff
@@ -97,7 +97,25 @@ describe('React module', () => {
...
   });
+  
+  xtest('static callbacks fire even if their own value has not changed', () => {
+    const inputCell = new InputCell(1);
+    const output = new ComputeCell(
+      [inputCell],
+      inputs => (inputs[0].value < 3 ? 111 : 222),
+    );
+
+    const callback = new CallbackCell(cell => "cell changed");
```
SleeplessByte:
```suggestion
    const callback = new CallbackCell(() => "cell changed");
```

Or if you want to indicate there is an argument passed in:

```suggestion
    const callback = new CallbackCell((_cell) => "cell changed");
```

SleeplessByte: Oh right. We changed this rule back to the default. Try `(ignore)`.

- Run the linter locally! (See https://github.com/exercism/javascript/blob/master/scripts/lint)
- You can pass in an assignment (IIRC `ASSIGNMENT=react`)

SleeplessByte: @yyyc514

yyyc514: Will try that later today I hope.  Thanks.

yyyc514:
I didn't see a way to make it happy so I updated the linter config to allow _ variables.

yyyc514:
```
Checking integrity of .eslintrc...
.eslintrc did not match for react
```

So individual file exist but we're not supposed to change them?

tejasbubane:
> So individual file exist but we're not supposed to change them?

@yyyc514 we have a CI check to ensure files like `package.json`, `package-lock.json` & `.eslintrc` are exactly the same for all exercises. We match them against files in root directory of this repo. You need to change `.eslintrc` from root and then run `npx @babel/node scripts/sync` to sync it to all exercises. Commit these changes.

Sorry for the confusion. I will put together contributors instructions soon.

yyyc514:
Should I really just change it to not show a passed value at all?

SleeplessByte:
If the value passed is not used, it's common not to have it at all

SleeplessByte:
Thank you @yyyc514 . This is great.

yyyc514:
Thanks.  I try.

PR #708, opened 2019-08-13
linked-list: add test 'deletes only the first occurence'

cmccandless:
The following implementation of `LinkedList.delete(value)` would pass existing tests, but is incorrect:
```Javascript
  delete(value) {
    let current = this.head;
    while (current !== null) {
      if (value === current.value) {
        if (current.prev === null) {
          this.head = current.next;
        } else {
          current.prev.setNext(current.next);
        }
        if (current.next === null) {
          this.tail = current.prev;
        } else {
          current.next.setPrev(current.prev);
        }
      }
      current = current.next;
    }
  }
```

### DIFF
--- a/exercises/linked-list/.meta/hints.md
+++ b/exercises/linked-list/.meta/hints.md
@@ -1,0 +1,2 @@
+Your list must also implement the following interface:
+- `delete` (delete the first occurence of a specified value)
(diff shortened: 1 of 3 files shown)


SleeplessByte:
Oof. We really need canonical data for this one.

Can you also add a track specific `HINTS.md` that adds a description for `delete`?

cmccandless:
> Can you also add a track specific HINTS.md that adds a description for delete?

Done.

SleeplessByte:
Thanks @cmccandless , great as always.

PR #710, opened 2019-08-20
[Space Age] Make list consistent, put Earth in proper order

yyyc514:
Sorry, this was driving me nuts.

### DIFF
--- a/exercises/space-age/README.md
+++ b/exercises/space-age/README.md
@@ -2,9 +2,9 @@
 
 Given an age in seconds, calculate how old someone would be on:
 
-   - Earth: orbital period 365.25 Earth days, or 31557600 seconds
    - Mercury: orbital period 0.2408467 Earth years
    - Venus: orbital period 0.61519726 Earth years
+   - Earth: orbital period 1.0 Earth years, 365.25 Earth days, or 31,557,600 seconds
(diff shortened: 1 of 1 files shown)


tejasbubane:
We try to be consistent with [canonical data](https://github.com/exercism/problem-specifications/blob/00dda3a9f5e7aa5a15d5db36535ae9320d840441/exercises/space-age/description.md). Also there are [specs](https://github.com/exercism/javascript/blob/427d409972d8701c7fec858a42cec7b4487e58c9/exercises/space-age/space-age.spec.js) with the same ordering. `Earth` is the base case, so I am not sure if we should change this.

yyyc514:
I can make a PR against canonical if needed. I’m not sure why the _order_ of the specs should affect the writing of the instructions.  That seems entirely irrelevant.

yyyc514:
https://github.com/exercism/problem-specifications/pull/1573

yyyc514:
> Earth is the base case

Just saw this.  **Not sure I agree.**  Seconds aren't unique to earth.  The input is in seconds.  **Given an age in seconds find your age on any planet.**  _Earth is not a base case._  Often students even go so far as to prove this by removing the common "earth_year_in_seconds" constant that so many choose to use.

 I don't think that needs have any bearing on its ordering in the instructions list.

Issue #667, opened 2019-04-12
Checklist: normalise topics


SleeplessByte:
**Note**: the suggestions are generated automagically. If something is _actually_ missing, please open an issue at [problem-specifications](https://github.com/exercism/problem-specifications) to add the missing topics.

```
=> fetching config from ../exercism-javascript/config.json
=> reading topics from ..\problem-specifications\TOPICS.txt
```

<details>
  <summary>List of valid topics</summary>

Source: [#d28d6](https://github.com/exercism/problem-specifications/blob/d28d6a5f40bb19f6a5954b7bda3430f4089aa680/TOPICS.txt)
- `abstract_classes`
- `callbacks`
- `conditionals`
- `events`
- `equality`
- `exception_handling`
- `generics`
- `inheritance`
- `interfaces`
- `loops`
- `operator_overloading`
- `pattern_matching`
- `polymorphism`
- `recursion`
- `threading`
- `type_conversion`
- `variables`
- `arrays`
- `booleans`
- `classes`
- `dates`
- `discriminated_unions`
- `enumerations`
- `floating_point_numbers`
- `graphs`
- `integers`
- `lists`
- `maps`
- `matrices`
- `optional_values`
- `queues`
- `records`
- `sequences`
- `sets`
- `stacks`
- `strings`
- `structs`
- `time`
- `trees`
- `tuples`
- `algorithms`
- `bitwise_operations`
- `concurrency`
- `cryptography`
- `domain_specific_languages`
- `files`
- `filtering`
- `games`
- `globalization`
- `logic`
- `math`
- `memory_management`
- `networking`
- `parallelism`
- `parsing`
- `pattern_recognition`
- `performance`
- `randomness`
- `refactoring`
- `regular_expressions`
- `searching`
- `security`
- `sorting`
- `text_formatting`
- `transforming`
- `behavior_driven_development`
- `functional_programming`
- `object_oriented_programming`
- `reactive_programming`
- `test_driven_development`
</details>

## Exercises with non-standard topics

- [ ] [`hello-world`](https://github.com/exercism/javascript/tree/master/exercises/hello-world)
  Incorrect topics:
  - [ ] `control_flow_conditionals` (perhaps: `conditionals`)
- [ ] [`pangram`](https://github.com/exercism/javascript/tree/master/exercises/pangram)
  Incorrect topics:
  - [ ] `control_flow_conditionals` (perhaps: `conditionals`)
  - [ ] `control_flow_loops` (perhaps: `loops`)
[...]

PR #711, opened 2019-08-21
resolved issue #667

alirezaghey:
Kept "data_structures" since I find it sensible.
Needs to be added to https://github.com/exercism/problem-specifications

### DIFF
--- a/config.json
+++ b/config.json
@@ -12,7 +12,7 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "control_flow_conditionals",
+        "conditionals",
         "optional_values",
         "strings",
         "text_formatting"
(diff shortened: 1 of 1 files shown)


SleeplessByte:
I love this @alirezaghey . Amazing work.

The only thing I would like to change is: can you format the file like before? Each array item on their own line? Apart form that, this is perfect. I personally don't think Matrix is a datastructure "topic" -- could you remove it for now? We'll figure out a _better_ topic and add it to ProblemSpec.

alirezaghey:
Thanks @SleeplessByte. That's encouraging.

Other than in **Matrix**, there or six other occurrences of **data_structures** in the file.
Just to be clear, you are OK with keeping those and want me to only remove it from **Matrix**, right?

SleeplessByte:
Nah, all of them. And let's open a new issue with those 7 that we have to see if there is an alternative!

SleeplessByte:
Thank you for this contribution @alirezaghey

PR #718, opened 2019-08-28
Add testcase

DagmarTimmreck:
The following wrong solution passes all previous tests but not the new one:
```
export class Matrix {
  constructor(string) {
    this._rows = string
      .split('\n')
      .map(row => row.split(' ').map(Number))
    this._columns = this._rows
      .map((_, i) => this._rows.map(row => row[i]))
  }

  get rows() {
    return this._rows
  }

  get columns() {
    return this._columns
  }
}
```


DagmarTimmreck:
See also 
https://github.com/exercism/problem-specifications/pull/1576

SleeplessByte:
Thank you for doing this @DagmarTimmreck :heart:

PR #719, opened 2019-08-31
Update bob/example.js

brendan-c:
Implemented `String.prototype.endsWith()` for brevity


SleeplessByte:
This is fine, but the example code is not meant to be _the best implementation_, only that it _is possible to solve_.

PR #723, opened 2019-09-15
Grains exercise using standard `BigInt` type

paparomeo:
Use upcoming standard [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) in the `grains` exercise, since that is supported since Node.js 10.4.0 and already supported by most major browsers.

I'm not sure what the procedure is to introduce the new `@babel/plugin-syntax-bigint` dependency (does it need to be added to the base `package.json` and _all_ exercises), since it's breaking the build when verifying the checksum.


SleeplessByte:
> I'm not sure what the procedure is to introduce the new @babel/plugin-syntax-bigint dependency (does it need to be added to the base package.json and all exercises), since it's breaking the build when verifying the checksum.

Yes, add it to the base `package.json` and run the "sync" script. It should be listed in CONTRIBUTING.

**Note**: I'm approving this because BigInt is `stage-4`, meaning it ís standard.

paparomeo:
This should be good to go now @SleeplessByte.
*Notable changes since last review*:
- Dropped CI testing for Node.js 8 and 9 since these versions don't support `BigInt` and were failing, but are no longer supported (Node.js 8 stopped being supported on 2020-01-01).
- Not using `BigInt` literals since these are still not supported by Jest (will only be [supported in the upcoming version 25](https://github.com/facebook/jest/pull/8382)).

paparomeo:
Naughty sync script! I _think_ I've fixed "all the things" @SleeplessByte, when you can have another look.

I'll create an for the sync script removing the versions problem tomorrow and might put in PR to fix it if it's straightforward.

tejasbubane:
> Actually, I see that there is an issue with the sync script. The versions should not be removed :(

Ohh boy! I just saw this and realized I deleted exercise versions in https://github.com/exercism/javascript/pull/824 😢

SleeplessByte:
@tejasbubane perhaps we revert #824 so we can merge this? Then fix the script, then re-apply it?

tejasbubane:
@SleeplessByte Done 👍

PR #734, opened 2019-10-02
Update isbn-verifier.spec.js

pyko:
Updating **X is only valid as a check digit** test so if `X` in the middle of ISBN is treated as `10` the test will fail.

Changing check digit to `9` means `checksum('3-598-2X507-9') mod 11 == 0`, so solutions which simply replace `X` with `10` (without checking location) will treat `3-598-2X507-9` as a valid ISBN.

### DIFF
--- a/exercises/isbn-verifier/isbn-verifier.spec.js
+++ b/exercises/isbn-verifier/isbn-verifier.spec.js
@@ -32,7 +32,7 @@ describe('ISBN Verifier Test Suite', () => {
   });
 
   xtest('X is only valid as a check digit', () => {
-    const isbn = new ISBN('3-598-2X507-0');
+    const isbn = new ISBN('3-598-2X507-9');
 
     expect(isbn.isValid()).toEqual(false);
   });


SleeplessByte:
@petertseng just tagging you because I want to be certain this can actually be applied and applied everywhere. In that case its a bugfix and we can lift it to problem-spec.

petertseng:
This PR is the correct thing to do; an implementation that treats X as 10 in all positions would still reject 3-598-2X507-0, therefore 3-598-2X507-0 is not a good test case. Such an implementation would declare 3-598-2X507-9 a valid ISBN, so 3-598-2X507-9 is the correct test case. Please see https://github.com/exercism/problem-specifications/pull/993.

petertseng:
Please see the following:

0 * 1 (0)
7 * 2 (14)
0 * 3 (0)
5 * 4 (20)
X * 5 (50)
2 * 6 (12)
8 * 7 (56)
9 * 8 (72)
5 * 9 (45)
3 * 10 (30)
total: 299 (divisible by 11? false)\

9 * 1 (9)
7 * 2 (14)
0 * 3 (0)
5 * 4 (20)
X * 5 (50)
2 * 6 (12)
8 * 7 (56)
9 * 8 (72)
5 * 9 (45)
3 * 10 (30)
total: 308 (divisible by 11? true)

SleeplessByte:
Sweet. Thank you for that elaboration @petertseng ❤️

Issue #705, opened 2019-08-03
Create stub files for all exercises


kytrinyx:
We have decided to require all file-based tracks to provide stubs for their exercises.

The lack of stub file generates an unnecessary pain point within Exercism, contributing a significant proportion of support requests, making things more complex for our students, and hindering our ability to automatically run test-suites and provide automated analysis of solutions.

We believe that it’s essential to understand error messages, know how to use an IDE, and create files. However, getting this right as you’re just getting used to a language can be a frustrating distraction, as it can often require a lot of knowledge that tends to seep in over time. At the start, it can be challenging to google for all of these details: what file extension to use, what needs to be included, etc. Getting people up to speed with these things are not Exercism’s focus, and we’ve decided that we are better served by removing this source of confusion, letting people get on with actually solving the exercises.

The original discussion for this is at https://github.com/exercism/discussions/issues/238.

Therefore, we’d like this track to provide a stub file for each exercise.

- If this track already provides stub files for all exercises, please close this issue.
- If this track already has an open issue for creating stubs, then my apologies. Please close one as a duplicate.
- Otherwise, please respond to this issue with useful details about what needs to be done to complete this task in this track so that people who are not familiar with the track may easily contribute.

SleeplessByte:
Closing as we already provide this.

tejasbubane:
Opening because of https://github.com/exercism/javascript/issues/721 & https://github.com/exercism/javascript/pull/722

rsuttles58:
I'm going to start knocking out some of these stubs for hacktoberfest.  I may do one in the first pull and if those get approved cleanly then I may bang out a few more.

MaruthiMohanReddyPutha:
Hi team, I would like to contribute to adding stub files to the repo. Could you please provide me the access to create a branch and pull requests?

tejasbubane:
Please check linked PRs to avoid duplication of work.

Roshanjossey:
@tejasbubane, does it make sense to keep a checklist for contributors to better understand which ones are done and which ones are not?

tejasbubane:
Only `diffie-hellman` left.

SleeplessByte:
https://tracks.exercism.io/javascript/master/stubs yasss.

tejasbubane:
Completed and verified by https://github.com/exercism/javascript/pull/722

PR #735, opened 2019-10-02
Add stubs for some exercises

TomPradat:
Part of #705 

Can you confirm that I'm on the right tracks ? I'll add some more stubs

### DIFF
--- a/exercises/change/change.js
+++ b/exercises/change/change.js
@@ -1,0 +1,10 @@
+//
+// This is only a SKELETON file for the 'Change' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+export class Change {
+  calculate(coinArray, target) {
+    throw new Error("Remove this statement and implement this function");
+  }
+}


exercises/change/change.js
```diff
@@ -0,0 +1,10 @@
+//
+// This is only a SKELETON file for the 'Alphametics' exercise. It's been provided as a
```
SleeplessByte:
```suggestion
// This is only a SKELETON file for the 'Change' exercise. It's been provided as a
```

TomPradat: Copy pasting 🙄 , i'll fix this

exercises/zipper/zipper.js
```diff
@@ -0,0 +1,50 @@
+//
+// This is only a SKELETON file for the 'Zipper' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+export default class Zipper {
+  fromTree(tree) {
```
SleeplessByte:
```suggestion
  static fromTree(tree) {
```

exercises/zipper/zipper.js
```diff
@@ -0,0 +1,50 @@
+//
```
SleeplessByte: Looking at https://github.com/TomPradat/javascript/blob/feature/stubs/exercises/zipper/zipper.spec.js, I don't think the methods in this file are correct. I don't see `left` and `right` for example, and things like `insertBefore` are not methods _under test.

TomPradat: I get what you mean, but then there is probably a problem with the README.md for this exercise => https://github.com/exercism/javascript/blob/master/exercises/zipper/README.md

Nowhere is it mentioned that there should have `left` or `right` methods.

So the question is should we update the README.md or the example and test files ?

SleeplessByte: Ahhh. Good question.

The README.md states

> For example given a rose tree (where each node contains a value and a list of child nodes) a zipper might support these operations:

I think we need to bolden the *For example* and we should add a JavaScript Track Hint which methods _are_ added/necessary, OR change the tests to update. I'll look into it <3

TomPradat: Agreed, thanks 👍

exercises/zipper/zipper.js
```diff
@@ -0,0 +1,50 @@
...
+
+  insertAfter() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
+  delete() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
+  next() {
```
tejasbubane: `next` function is duplicated. Please remove one.

tejasbubane:
Ahh, I merged https://github.com/exercism/javascript/pull/753 with zipper stub so it created conflict with this PR. Sorry for the confusion. Can you fix the merge conflicts please? @TomPradat

PR #743, opened 2019-10-03
Update INSTALLATION.md

SleeplessByte:
Adds a more streamlined - beginner friendly, with less magic, description on installation and checking if the installation succeeded.

Per @ihid.

(Reviewers: please check spelling, wording and feel free to suggest better/more content)

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,31 +1,120 @@
-There are a couple of ways to install [NodeJS](http://nodejs.org):
+This track relies on [NodeJS][web-nodejs] throughout to provide a runtime for 
+JavaScript. This means that we assume all execution of JavaScript on your 
+computer will happen using [NodeJS][web-nodejs]. 
 
-* via an [Installer or Binaries](https://nodejs.org/en/download/)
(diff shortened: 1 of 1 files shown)


docs/INSTALLATION.md
```diff
@@ -1,31 +1,120 @@
...
+
+The `hello-world` assignment will teach you how to use the `exercism` cli and
+download an assignment to your computer. 
+
+Each assignment then needs some tools to run the tests:
+
+* [Jest][web-jest]: a test runner based on Jasmine
+* [Babel][web-babel]: to transpile ESNext code to work on your NodeJS version
+* [ESLint][web-eslint]: to perform several static analysis and coding style 
+checks to your JavaScript code.
```
tejasbubane: It might be a little intimidating for beginners of the track with all these tools being used. Do we need to mention these? Can we skip to `npm install`?

SleeplessByte: I only have them because we had them before. So perfectly fine to _remove_ them here 🗡

SleeplessByte: Updated!

PR #752, opened 2019-10-06
Add new exercise grep

### LINKED ISSUE #737: Add new exercise: grep
[problem specification][problem-specification] | [description][readme] |  [canonical data][canonical-data]

[problem-specification]: https://github.com/exercism/problem-specifications/tree/master/exercises/grep
[readme]: https://github.com/exercism/problem-specifications/blob/master/exercises/grep/description.md
[canonical-data]: https://github.com/exercism/problem-specifications/blob/master/exercises/grep/canonical-data.json

TomPradat:
Fixes #737 .

I was wondering about the fact that this exercise is a CLI.

Unless I missed something, it's not possible to test a file that just executes a set of instructions.

I was thinking about exporting a function that takes the pattern, the files and options as arguments like that :

```js
export default (pattern, files, options =  {}) => {}
```
Is the user supposed to use the filesystem or should I provide him the data in the files in some way ?

### DIFF
--- a/config.json
+++ b/config.json
@@ -1435,7 +1435,7 @@
       "slug": "grep",
       "uuid": "78bcbae1-a0f2-460c-ba89-e51844fe9397",
       "core": false,
-      "unlocked_by": null,
+      "unlocked_by": "matrix",
       "difficulty": 4,
       "topics": [
         "files",
--- a/config.json
+++ b/config.json
@@ -1430,6 +1430,18 @@
         "strings"
       ],
       "deprecated": true
+    },
+    {
+      "slug": "grep",
+      "uuid": "78bcbae1-a0f2-460c-ba89-e51844fe9397",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "files",
+        "searching",
+        "text_formatting"
+      ]
     }
   ]
 }
--- a/exercises/grep/grep.js
+++ b/exercises/grep/grep.js
@@ -6,7 +6,7 @@
 //
 // node grep.js args
 //
-// Instead of "grep args".
+// Instead of "./grep.js args".
 // 
 // Read more about shebangs here: https://en.wikipedia.org/wiki/Shebang_(Unix)
 // 
--- a/exercises/grep/grep.js
+++ b/exercises/grep/grep.js
@@ -1,0 +1,4 @@
+//
+// This is only a SKELETON file for the 'Grep' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
--- a/exercises/grep/grep.js
+++ b/exercises/grep/grep.js
@@ -1,4 +1,15 @@
+#!/usr/bin/env node
+
+// The above line is a shebang. On Unix-like operating systems, or environments, this will allow the script to be
+// run by node, and thus turn this JavaScript file into an executable. If you don't have a Unix-like operating
+// system or environment, for example Windows without WSL, you can use 
 //
+// node grep.js args
+//
+// Instead of "grep args".
+// 
+// Read more about shebangs here: https://en.wikipedia.org/wiki/Shebang_(Unix)
+// 
 // This is only a SKELETON file for the 'Grep' exercise. It's been provided as a
 // convenience to get you started writing code faster.
 //
--- a/exercises/grep/grep.spec.js
+++ b/exercises/grep/grep.spec.js
@@ -20,9 +20,9 @@ const spawn_grep = config => {
 
     if (stderr) {
       reject(stderr);
+    } else {
+      resolve(stdout);
     }
-
-    resolve(stdout);
   });
 };
 
@@ -116,11 +116,9 @@ describe("grep exercise", () => {
           files: ["midsummer-night.txt"]
         })
       ).resolves.toBe(
-        formatStringTemplate(
-          `3:Nor how it may concern my modesty,
-            5:But I beseech your grace that I may know
-            6:The worst that may befall me in this case,`
-        )
+        formatStringTemplate(`3:Nor how it may concern my modesty,
+        5:But I beseech your grace that I may know
+        6:The worst that may befall me in this case,`)
       );
     });
 
--- a/exercises/grep/grep.spec.js
+++ b/exercises/grep/grep.spec.js
@@ -1,233 +1,371 @@
-const { spawnSync } = require('child_process')
-const path = require('path')
+const { spawnSync } = require("child_process");
+const path = require("path");
 
-const BASE_DIR = './tmp_exercises';
+const BASE_DIR = path.resolve(__dirname);
 
-const resolveDataFile = file => path.resolve(BASE_DIR, 'data', file);
+const resolveDataFile = file => path.resolve(BASE_DIR, "data", file);
 
 const spawn_grep = config => {
   const args = [
-    path.resolve(BASE_DIR, './grep.js'),
+    path.resolve(BASE_DIR, "grep.js"),
     ...config.flags,
     config.pattern,
-    ...config.files.map(file => path.resolve(BASE_DIR, 'data', file))
-  ]
+    ...config.files.map(file => path.resolve(BASE_DIR, "data", file))
+  ];
 
   return new Promise((resolve, reject) => {
-    const child = spawnSync('node', args, { stdio: 'pipe' })
-    const stderr = child.stderr.toString().trim()
-    const stdout = child.stdout.toString().trim()
+    const child = spawnSync("node", args, { stdio: "pipe" });
+    const stderr = child.stderr.toString().trim();
+    const stdout = child.stdout.toString().trim();
 
     if (stderr) {
-      reject(stderr)
+      reject(stderr);
     }
 
-    resolve(stdout)
+    resolve(stdout);
   });
-}
+};
 
-describe('grep exercise', () => {
-  describe('Test grepping a single file', () => {
-    it('One file, one match, no flags', () => {
-      return expect(spawn_grep({
-        pattern: "Agamemnon",
-        flags: [],
-        files: ["iliad.txt"]
-      })).resolves.toBe("Of Atreus, Agamemnon, King of men.")
-    })
+const formatStringTemplate = stringTemplate =>
+  stringTemplate
+    .split("\n")
+    .map(sentence => sentence.trim())
+    .join("\n");
 
-    it('One file, one match, print line numbers flag', () => {
-      return expect(spawn_grep({
-        pattern: "Forbidden",
-        flags: ["-n"],
-        files: ["paradise-lost.txt"]
-      })).resolves.toBe("2:Of that Forbidden Tree, whose mortal tast")
-    })
+describe("grep exercise", () => {
+  describe("Test grepping a single file", () => {
+    it("One file, one match, no flags", () => {
+      return expect(
+        spawn_grep({
+          pattern: "Agamemnon",
+          flags: [],
+          files: ["iliad.txt"]
+        })
+      ).resolves.toBe("Of Atreus, Agamemnon, King of men.");
+    });
 
-    it('One file, one match, case-insensitive flag', () => {
-      return expect(spawn_grep({
-        pattern: "FORBIDDEN",
-        flags: ["-i"],
-        files: ["paradise-lost.txt"]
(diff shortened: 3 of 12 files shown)


SleeplessByte:
Great question. From the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/grep/canonical-data.json):

```
    " The language track implementing this exercise       ",
    " should ensure that when the tests run, three files  ",
    " are created with the following contents. The file   ",
    " names and their contents are listed below:          ",
```

I think we should _actually create these files_, instead of having them be generated by the tests. So, add `iliad.txt`, `midsummer-night.txt` and `paradise-lost.txt` as files to the exercise directory `grep` (these will be downloaded automagically when using the `exercism` cli!)

For your second question, let's see if @tejasbubane agrees with this, I actually think it would be great if you make this script _executable_. That means, NO EXPORTS. I think we should do `spawn` or `spawnSync` and execute `node` on the file, with actual arguments, and use `stdout` for parsing. This exercise a the _perfect_ opportunity to learn about:

- `process.args` / `process.argv0`
- `process.stdout` / `process.stderr`
- `path` 
- `fs` 

## Example:

```javascript
describe("tests something", () => {
  it("works", () => {
    const output = spawnSync('node', ['grep.js', '"Agamemnon"', 'iliad.txt'], { stdio: 'pipe' }).stdout.toString().trim()
    expect(output).toBe("Of Atreus, Agamemnon, King of men.")
  })
})
```

## Stub

Add the hashbang at the top so people can actually execute it on UNIX/POSIX environments:

```javascript
#!/usr/bin/env node
```

We can even debate adding a "cross env" wrapper `grep.sh` and `grep.bat`, but that is to be seen.

TomPradat:
Alright, I didn't know about `spawSync`, I agree this is a great opportunity to learn more about nodejs, I'll go on this 👍

TomPradat:
The MVP is working, i'll finish this ASAP :)

TomPradat:
So i got a problem because the test script only copies the `example.js` and the spec file, which causes the script to fail because the .txt files are not copied

SleeplessByte:
> So i got a problem because the test script only copies the example.js and the spec file, which causes the script to fail because the .txt files are not copied

Yes! Let's add them. 

If you look at this code: https://github.com/exercism/javascript/blob/master/scripts/helpers.js#L74-L77

```js
  const libDir = ['exercises', assignment, 'lib'].join('/');
  if(shell.test('-d', libDir)) {
    shell.cp(libDir + '/*.js', 'tmp_exercises/lib');
  }
```

... you can see the `lib` dir is being copied, as long as it ends on .js. Let's add a `data` folder, where we copy anything:

```js
  const dataDir = ['exercises', assignment, 'data'].join('/');
  if(shell.test('-d', dataDir)) {
    shell.cp([dataDir, '*'].join('/'), 'tmp_exercises/data');
  }
```

TomPradat:
> 
> 
> > So i got a problem because the test script only copies the example.js and the spec file, which causes the script to fail because the .txt files are not copied
> 
> Yes! Let's add them.
> 
> If you look at this code: https://github.com/exercism/javascript/blob/master/scripts/helpers.js#L74-L77
> 
> ```js
>   const libDir = ['exercises', assignment, 'lib'].join('/');
>   if(shell.test('-d', libDir)) {
>     shell.cp(libDir + '/*.js', 'tmp_exercises/lib');
>   }
> ```
> 
> ... you can see the `lib` dir is being copied, as long as it ends on .js. Let's add a `data` folder, where we copy anything:
> 
> ```js
>   const dataDir = ['exercises', assignment, 'data'].join('/');
>   if(shell.test('-d', dataDir)) {
>     shell.cp([dataDir, '*'].join('/'), 'tmp_exercises/data');
>   }
> ```

Nice, should I add it myself ?

SleeplessByte:
Yes go ahead!

TomPradat:
Should I add stuff in the stub file ? Or should we let the user do the exercise from scratch ?

Also should I write a README.md ?

SleeplessByte:
See CONTRIBUTING for the readme generation. You don't need to do this yourself.

Let's add the stub file, but only have a comment with 'this is a stub blablabla'

TomPradat:
If everything is ok, i can rebase my last 3 commits and we can merge this :)

SleeplessByte:
You don't need to rebase. I'll squash-merge

TomPradat:
Hi, do you have time to merge it ?

config.json
```diff
@@ -1430,6 +1430,18 @@
         "strings"
       ],
       "deprecated": true
+    },
+    {
+      "slug": "grep",
+      "uuid": "78bcbae1-a0f2-460c-ba89-e51844fe9397",
+      "core": false,
+      "unlocked_by": null,
```
SleeplessByte: Let's pick an exercise. Just pick an exercise with `"core": true`, where you think it would be nice to have this as a side _after_! It can be any of the core exercises.

config.json
```diff
@@ -1430,6 +1430,18 @@
...
+      "difficulty": 4,
+      "topics": [
```
SleeplessByte: Can you confirm that all three are valid topics from the TOPICS.txt list?

exercises/grep/grep.js
```diff
@@ -0,0 +1,4 @@
+//
```
SleeplessByte:
```suggestion
#!/usr/bin/env node

// The above line is a shebang. On Unix-like operating systems, or environments, this will allow the script to be
// run by node, and thus turn this JavaScript file into an executable. If you don't have a Unix-like operating
// system or environment, for example Windows without WSL, you can use 
//
// node grep.js args
//
// Instead of "grep args".
// 
// Read more about shebangs here: https://en.wikipedia.org/wiki/Shebang_(Unix)
// 
```

exercises/grep/grep.spec.js
```diff
@@ -0,0 +1,233 @@
+const { spawnSync } = require('child_process')
+const path = require('path')
+
+const BASE_DIR = './tmp_exercises';
```
SleeplessByte:
```suggestion
const BASE_DIR = path.join('.', 'tmp_exercises');
```

exercises/grep/grep.spec.js
```diff
@@ -0,0 +1,233 @@
...
+
+const resolveDataFile = file => path.resolve(BASE_DIR, 'data', file);
+
+const spawn_grep = config => {
+  const args = [
+    path.resolve(BASE_DIR, './grep.js'),
```
SleeplessByte:
```suggestion
    path.resolve(BASE_DIR, 'grep.js'),
```

exercises/grep/grep.spec.js
```diff
@@ -0,0 +1,233 @@
...
... do entreat your grace to pardon me.\n${resolveDataFile('midsummer-night.txt')}:I know not by what power I am made bold,\n${resolveDataFile('midsummer-night.txt')}:Nor how it may concern my modesty,\n${resolveDataFile('midsummer-night.txt')}:In such a presence here to plead my thoughts;\n${resolveDataFile('midsummer-night.txt')}:But I beseech your grace that I may know\n${resolveDataFile('midsummer-night.txt')}:The worst that may befall me in this case,\n${resolveDataFile('midsummer-night.txt')}:If I refuse to wed Demetrius.\n${resolveDataFile('paradise-lost.txt')}:Of Mans First Disobedience, and the Fruit\n${resolveDataFile('paradise-lost.txt')}:Of that Forbidden Tree, whose mortal tast\n${resolveDataFile('paradise-lost.txt')}:Brought Death into the World, and all our woe,\n${resolveDataFile('paradise-lost.txt')}:With loss of Eden, till one greater Man\n${resolveDataFile('paradise-lost.txt')}:Restore us, and regain the blissful Seat,\n${resolveDataFile('paradise-lost.txt')}:Sing Heav'nly Muse, that on the secret top\n${resolveDataFile('paradise-lost.txt')}:Of Oreb, or of Sinai, didst inspire\n${resolveDataFile('paradise-lost.txt')}:That Shepherd, who first taught the chosen Seed`)
```
SleeplessByte: Can you add the newlines in the string directly so it's not this long? String template literals allow for newlines!

exercises/grep/grep.spec.js
SleeplessByte: This actually doesn't allow you to run this _locally_, as locally the `tmp_exercises` directory won't exist. This should probably something along the lines of:

```javascript
const BASE_DIR = path.resolve(__dirname)
```

... this way it doesn't matter _where_ the `cwd()` is, because `__dirname` will be the `path.dirname` of the _current file_!

config.json
TomPradat: Well i got the topics from the python track so i guess there must be valid => https://github.com/exercism/python/blob/master/config.json

config.json
TomPradat: Alright, I'll try that, but I don't know all the exercises, so i'll choose given the difficulty of the core exercises

exercises/grep/grep.spec.js
TomPradat: The problem is with the tabs, i used a function to remove the tabs. If you though about something better, let me know 👍

exercises/grep/grep.spec.js
TomPradat: Indeed, this is much cleaner, thanks :)

TomPradat:
This should be good now I think

exercises/grep/grep.js
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
+
+// The above line is a shebang. On Unix-like operating systems, or environments, this will allow the script to be
+// run by node, and thus turn this JavaScript file into an executable. If you don't have a Unix-like operating
+// system or environment, for example Windows without WSL, you can use 
+//
+// node grep.js args
+//
+// Instead of "grep args".
```
Calamari: Shouldn't that be more like `"./grep.js args"`?

The file is still named with suffix `.js` and if it is not added to the `$PATH` I don't think just writing the name will execute it (at least not in my shell).

exercises/grep/grep.spec.js
```diff
@@ -0,0 +1,371 @@
...
+      return expect(
+        spawn_grep({
+          pattern: "may",
+          flags: ["-n"],
+          files: ["midsummer-night.txt"]
+        })
+      ).resolves.toBe(
+        formatStringTemplate(
+          `3:Nor how it may concern my modesty,
+            5:But I beseech your grace that I may know
```
Calamari: Indentation looks inconsistent here to the `formatStringTemplates` used below

Calamari:
I like this exercise, I think it would make a great addition to explore JS more.

exercises/grep/grep.js
TomPradat: Indeed

exercises/grep/grep.spec.js
TomPradat: Weird, I used prettier to format the file, I'll check it :+1:

Calamari: Ah, it might be, that you started the backtick in a new row then, the other ones, you put it into the same line as `formatStringTemplate`, which might be the reason for prettier to handle it that way.

Then disregard my comment :)

exercises/grep/grep.spec.js
```diff
@@ -0,0 +1,371 @@
...
+    ...config.files.map(file => path.resolve(BASE_DIR, "data", file))
+  ];
+
+  return new Promise((resolve, reject) => {
+    const child = spawnSync("node", args, { stdio: "pipe" });
+    const stderr = child.stderr.toString().trim();
+    const stdout = child.stdout.toString().trim();
+
+    if (stderr) {
+      reject(stderr);
```
Calamari: Shouldn't here be a return as well?
Else it would try also to run `resolve(stdout)` below, right?

TomPradat: I've done a few tests, it seems to work fine. I added an else statement though because it's less error prone indeed

Calamari: Sounds like a good idea. Promises cannot be changed, but I was not quite sure about the implementation. I think I also had one once where it would throw if you try to call resolve or reject a second time.

exercises/grep/grep.js
SleeplessByte: The POSIX variant is `grep.js args` if you're in the right directory, but this comment was initially meant to mimic `actual-grep args`

PR #753, opened 2019-10-07
Add stubs for some exercises #735

hyuko21:
I've added stubs for **zipper**, **wordy** and **word-search** listed at  #705

### DIFF
--- a/exercises/word-search/word-search.js
+++ b/exercises/word-search/word-search.js
@@ -13,4 +13,4 @@ class WordSearch {
   }
 }
 
-module.exports = WordSearch;
+export default WordSearch;
--- a/exercises/word-search/word-search.js
+++ b/exercises/word-search/word-search.js
@@ -3,42 +3,6 @@
 // convenience to get you started writing code faster.
 //
 
-function searchHorizontally() {
-  throw new Error("Remove this statement and implement this function");
-}
-
-function flipCoordinates() {
-  throw new Error("Remove this statement and implement this function");
-}
-
-function flipGrid() {
-  throw new Error("Remove this statement and implement this function");
-}
-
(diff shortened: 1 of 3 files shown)


exercises/word-search/word-search.js
```diff
@@ -0,0 +1,52 @@
...
+  constructor() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
+  find() {
+    throw new Error("Remove this statement and implement this function");
+  }
+}
+
+module.exports = WordSearch;
```
SleeplessByte: If you look at the examples and other stubs, you see that we use `es6` `export`, not `CommonJS` `module.exports`.

exercises/word-search/word-search.js
```diff
@@ -0,0 +1,52 @@
+//
+// This is only a SKELETON file for the 'Word Search' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+function searchHorizontally() {
```
SleeplessByte: All these top-level functions should _not_ be stubbed, as an implementation for this exercises does not necessarily need them. The `class` and `find` are correct.

tejasbubane: You can check corresponding [spec file](https://github.com/exercism/javascript/blob/4a0f58d67f27fda8d433f934aaeef386f3f1033e/exercises/word-search/word-search.spec.js) for the exercise and stub **only** the functions/classes used there.

hyuko21: Thanks for your replies. Sorry for those mistakes, I didn't realize it at first.

I've had made those requested changes.

SleeplessByte:
Great fix @hyuko21 , thanks!!

Issue #741, opened 2019-10-03
What was it like to learn JavaScript?


ErikSchierboom:
We’ve recently [started a project](https://exercism.io/blog/redesigning-tracks-in-partnership-with-chicago-university-and-sloan-foundation) to find the best way to design our tracks, in order to optimize the learning experience of students.

As a first step, we’ll be [examining the ways in which languages are unique and the ways in which they are similar](https://github.com/exercism/exercism/issues/5059). For this, we’d really like to use the knowledge of **everyone** involved in the Exercism community (students, mentors, maintainers) to answer the following questions:

1. How was your experience learning JavaScript? What was helpful while learning JavaScript? What did you struggle with? How did you tackle problems?
1. In what ways did JavaScript differ from other languages you knew at the time? What was hard to learn? What did you have to unlearn? What syntax did you have to remap? What concepts carried over nicely?

Could you spare 5 minutes to help us by answering these questions? It would greatly help us improve the experience students have learning JavaScript :)

_Note: this issue is **not** meant as a discussion, just as a place for people to post their own, personal experiences._

_Want to keep your thoughts private but still help? Feel free to email me at erik@exercism.io_

Thank you!

tstirrat15:
I started off being a little scared and reluctant. Seeing things like [wat](https://www.destroyallsoftware.com/talks/wat) made it seem like JS was at best weird and at worst kinda broken?

The first project where I first really tried to have a go at it was an Angular2/Typescript project, where everything was OOP and typed. I muddled through, but the OOP patterns dictated by Angular2's design didn't make a whole lot of sense to me.

I started properly getting into it with my current job, where I work on a frontend written in ES6 and React. I really felt like I found a groove here, and a lot of it was getting more comfortable with the functional programming techniques that Javascript makes really easy.

I love how easy it is to write a curried function with arrow notation. I love how easy it is to express functions of functions.

Some problems I ran into included the concepts of equality, and how that's down to object references, and needing to keep track of what's considered an object and what's considered a primitive, and how that affects notions of equality.

I can't speak to the experience of using classes extensively - I tend to prefer a functional style, so I haven't had to deal with things that people complain about in that sphere like the weirdnesses of prototypical inheritance.

seeksort:
[...]

PR #754, opened 2019-10-08
add scale-generator exercise

### LINKED ISSUE #751: Add new exercise: scale-generator
[problem specification][problem-specification] | [description][readme] |  [canonical data][canonical-data]

[problem-specification]: https://github.com/exercism/problem-specifications/tree/master/exercises/scale-generator
[readme]: https://github.com/exercism/problem-specifications/blob/master/exercises/scale-generator/description.md
[canonical-data]: https://github.com/exercism/problem-specifications/blob/master/exercises/scale-generator/canonical-data.json

seeksort:
Adding scale-generator exercise. Closes #751. 

One thing I'm not sure about is updating `config.json`. I just based it on the same entry in other tracks, not sure if there's other stuff I should consider.

### DIFF
--- a/config.json
+++ b/config.json
@@ -994,10 +994,11 @@
       "slug": "scale-generator",
       "uuid": "b9c586e8-998b-4f5d-ab98-a08be29a9f19",
       "core": false,
-      "unlocked_by": null,
+      "unlocked_by": "pangram",
       "difficulty": 3,
       "topics": [
-        "pattern_matching",
+        "loops",
+        "pattern_recognition",
         "strings",
         "arrays"
       ]
--- a/config.json
+++ b/config.json
@@ -991,6 +991,18 @@
       ]
     },
     {
+      "slug": "scale-generator",
+      "uuid": "b9c586e8-998b-4f5d-ab98-a08be29a9f19",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 3,
+      "topics": [
+        "pattern_matching",
+        "strings",
+        "arrays"
+      ]
+    },
+    {
       "slug": "connect",
       "uuid": "2fa2c262-77ae-409b-bfd8-1d643faae772",
       "core": false,
--- a/exercises/scale-generator/.eslintrc
+++ b/exercises/scale-generator/.eslintrc
@@ -1,0 +1,26 @@
+{
+  "root": true,
+  "parser": "babel-eslint",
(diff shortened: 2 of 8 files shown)


SleeplessByte:
Sweet. We'll review this. You did great with the config.

- Is this a new UUID?
- We need to decide _which_ exercise in the **core** unlocks this exercise
- Are those `topics` correct with this exercise?

seeksort:
Yup, I used a uuidgen. And the topics are covered by the exercise; I can add more if you think there are other relevant topics!

How do we determine the unlock exercise? Is it intended to be the same for all tracks or does it depend on the track? The [Python version](https://github.com/exercism/python/blob/master/config.json#L1140) uses `kindergarden-garten` as the unlock exercise, but when I tried to do that I got an error about using a non-core exercise to unlock another non-core exercise.

SleeplessByte:
Good question!

Each track has its _own_ core. You can see it [here](https://tracks.exercism.io/javascript/master) for JavaScript, in bold. Each side exercise is unlocked by a core exercise. Since tracks are not the same across languages, neither are the unlocks. Usually I do the following:

- Is it related to a core exercise, probably there (re: ciphers)
- Otherwise, is there a core exercise without anything? Re-assign things so each core exercise has some (not the case now)
- Otherwise, use binomial on the difficulty. So in this case, it should be tied to a core exercise with 2-4 difficulty, or can be added to an earlier one if there's not a lot of 3s yet.

If we look at the [Python track](https://tracks.exercism.io/python/master), we can see that `kindergarten` is mid of the second-third of the core exercises, which I think is late. I think `unlocked_by`: `pangram` is a good candidate!

seeksort:
Thanks for the guidance @SleeplessByte ! Looking at pangram, I agree it's a good candidate for unlocking this one, since it looks like there are similar matching and arrays elements. I'll push up an update for `config.json`.

SleeplessByte:
Let's do this!

SleeplessByte:
@seeksort if you have 5 more minutes, would you be so kind and see if you can help us out by responding to [this issue](https://github.com/exercism/javascript/issues/741)?

Awesome work on this. Let's see how this propagates :)

SleeplessByte:
image

Woot! 🎉

seeksort:
Yay! And it deployed so quickly!

Sure, I'll check it out. Thanks again for your help!

PR #755, opened 2019-10-08
Added stub file for Change

### LINKED ISSUE #705: Create stub files for all exercises

nimrodcoder:
Please check. Also is there any procedure on how to make these files. Will make more if procedure is explained.

### DIFF
--- a/exercises/change/change.js
+++ b/exercises/change/change.js
@@ -1,0 +1,14 @@
+//
+// This is only a SKELETON file for the 'Allergies' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+export class Candidate {
+  constructor() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
+  addCoin() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
+}


exercises/change/change.js
```diff
@@ -0,0 +1,14 @@
+//
+// This is only a SKELETON file for the 'Allergies' exercise. It's been provided as a
```
SleeplessByte:
```suggestion
// This is only a SKELETON file for the 'Change' exercise. It's been provided as a
```

exercises/change/change.js
```diff
@@ -0,0 +1,14 @@
...
+// convenience to get you started writing code faster.
+//
+export class Candidate {
```
SleeplessByte: According to [the `.spec` file](https://github.com/exercism/javascript/blob/952c01149428f1f1d8a82f65a37b8c02c5c95e42/exercises/change/change.spec.js), it's actually `Change` that is exported, not `Candidate`.

exercises/change/change.js
```diff
@@ -0,0 +1,14 @@
...
+  constructor() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
+  addCoin() {
```
SleeplessByte: The function expected is actually called `calculate` and receives _two_ arguments.

SleeplessByte:
This does NOT fix #705, which is an issue for _all_ stubs.

SleeplessByte:
There is also #735, which was first.

PR #757, opened 2019-10-09
Stub file for Two Fer exercise

mluisamc:


exercises/two-fer/two-fer.js
```diff
@@ -0,0 +1,10 @@
+//
+// This is only a SKELETON file for the 'Two fer' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+export class TwoFer {
+  return() {
```
SleeplessByte: The [spec](https://github.com/mluisamc/javascript/blob/stub-files/exercises/two-fer/two-fer.spec.js) does not expect a `class`, nor is this valid syntax.

mluisamc: Ok, modified!

exercises/two-fer/two-fer.js
```diff
@@ -0,0 +1,8 @@
+//
+// This is only a SKELETON file for the 'Two fer' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+export const twofer = () => {
```
SleeplessByte:
```suggestion
export const twoFer = () => {
```

The casing is incorrect. 

@tejasbubane maybe we want to use `export function twoFer` instead? Given that this is the first exercise and all?? What do you think?

exercises/two-fer/two-fer.js
```diff
@@ -0,0 +1,8 @@
...
+  throw new Error("Remove this statement and implement this function");
+};
```
SleeplessByte: Can you add a newline here?

exercises/two-fer/two-fer.js
tejasbubane: I always wonder why `export const xyz` is valid but `export class xyz` is not 🤷‍♂

exercises/two-fer/two-fer.js
tejasbubane: I prefer `twoFer` given the JS convention of camelCase names & exercise name being `two-fer`.

exercises/two-fer/two-fer.js
SleeplessByte: It is valid, I was talking about the `return() {`

exercises/two-fer/two-fer.js
SleeplessByte: I actually meant `export const twoFer` vs `export function twoFer`. What are your thoughts on that @tejasbubane ?

tejasbubane: I think `const` will be fine.

SleeplessByte:
Thank you!!

SleeplessByte:
@mluisamc if you have 5 more minutes, would you be so kind and see if you can help us out by responding to [this issue](https://github.com/exercism/javascript/issues/741)?

PR #758, opened 2019-10-09
Add stub for the 'circular-buffer' exercise

diego-caceres:
Part of #705 - Add stub for the 'circular-buffer' exercise.

The spec for this one was a bit different from the others, I was not sure what to implement as Skeleton for the Errors.

### DIFF
--- a/exercises/circular-buffer/circular-buffer.js
+++ b/exercises/circular-buffer/circular-buffer.js
@@ -4,6 +4,10 @@
 //
 
 class CircularBuffer {
+  constructor() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
   write() {
     throw new Error("Remove this statement and implement this function");
   }
@@ -21,10 +25,7 @@ class CircularBuffer {
   }
 }
 
-export default () => {
-  // Complete this with the code for the exercise.
-  return new CircularBuffer();
-}
+export default CircularBuffer;
 
 export class BufferFullError extends Error {
   constructor() {
--- a/exercises/circular-buffer/circular-buffer.js
+++ b/exercises/circular-buffer/circular-buffer.js
@@ -15,6 +15,10 @@ class CircularBuffer {
   forceWrite() {
     throw new Error("Remove this statement and implement this function");
   }
+
+  clear() {
+    throw new Error("Remove this statement and implement this function");
+  }
 }
 
 export default () => {
--- a/exercises/circular-buffer/circular-buffer.js
+++ b/exercises/circular-buffer/circular-buffer.js
@@ -22,6 +22,14 @@ export default () => {
   return new CircularBuffer();
 }
 
-export const BufferFullError = new Error("Error: Buffer is full");
+export class BufferFullError extends Error {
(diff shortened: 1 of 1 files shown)


exercises/circular-buffer/circular-buffer.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export default () => {
+  // Complete this with the code for the exercise.
+  return new CircularBuffer();
+}
+
+export const BufferFullError = new Error("Error: Buffer is full");
+
+export const BufferEmptyError = new Error("Error: Buffer is empty");
```
tejasbubane: A stubbed error like this would be better as this does not give away the message implementation:

https://github.com/exercism/javascript/blob/d1fcd01c54c213b5a6116de0d9e296028e70b0f1/exercises/robot-simulator/robot-simulator.js#L6-L10

diego-caceres: Ah great, I hadn't found that example. Thanks!

exercises/circular-buffer/circular-buffer.js
```diff
@@ -0,0 +1,35 @@
...
+    throw new Error("Remove this statement and implement this function");
+  }
+
+  read() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
+  forceWrite() {
+    throw new Error("Remove this statement and implement this function");
+  }
```
tejasbubane: There's one more function required - `clear` - which clears the buffer.

exercises/circular-buffer/circular-buffer.js
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export default () => {
+  // Complete this with the code for the exercise.
+  return new CircularBuffer();
+}
```
tejasbubane: We can directly export the class itself.

```js
export default CircularBuffer;
```

diego-caceres: I had thought about this, but if I do, when running the tests I see this error:
` TypeError: Class constructor CircularBuffer cannot be invoked without 'new'`

      3 | describe('CircularBuffer', () => {
      4 |   test('reading an empty buffer throws a BufferEmptyError', () => {
    > 5 |     const buffer = circularBuffer(1);
        |                    ^
      6 |     expect(() => buffer.read()).toThrow(BufferEmptyError);
      7 |   });
      8 | 

      at Object.test (exercises/circular-buffer/circular-buffer.spec.js:5:20)

The way it is used from the spec file made me think that it would need to export a function. What do you think? 
If this error is acceptable, I can change it to default export the class.

tejasbubane: Test files should use `new CircularBuffer`.

Keep `export default CircularBuffer` in stub file & `const buffer = new CircularBuffer` in test file.

diego-caceres: Oh, my bad, I didn't change it, but I hadn't updated it and it was modified a couple of days ago. I'll update it in a few minutes.

PR #759, opened 2019-10-09
Add stub for the 'connect' exercise

diego-caceres:
Part of #705 - Add stub for the 'connect' exercise.

### DIFF
--- a/exercises/connect/connect.js
+++ b/exercises/connect/connect.js
@@ -1,0 +1,14 @@
+//
+// This is only a SKELETON file for the 'Connect' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+export class Board {
+  winner() {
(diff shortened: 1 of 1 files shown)


exercises/connect/connect.js
```diff
@@ -0,0 +1,18 @@
...
+export class Board {
+  constructor() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
+  winner() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
+  score() {
```
tejasbubane: `score` is an internal function not required by [specs](https://github.com/exercism/javascript/blob/d1fcd01c54c213b5a6116de0d9e296028e70b0f1/exercises/connect/connect.spec.js). Users are free to have their own implementation with different internal functions.

PR #772, opened 2019-10-19
Added stub for "Queen Attack" exercise

abhnvgupta:
This PR is part of #705 and adds a stub file for the "Queen Attack" exercise.

### DIFF
--- a/exercises/queen-attack/queen-attack.js
+++ b/exercises/queen-attack/queen-attack.js
@@ -1,0 +1,20 @@
+//
+// This is only a SKELETON file for the 'Queen Attack' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+class QueenAttack {
+    constructor(params = { white: [], black: []}) {
+        this.white = params.white;
+        this.black = params.black;
+        throw new Error("Remove this statement and implement this function");
+    }
+
+    toString() {
+        throw new Error("Remove this statement and implement this function");
+    }
+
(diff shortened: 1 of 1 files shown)


exercises/queen-attack/queen-attack.js
```diff
@@ -0,0 +1,20 @@
+//
+// This is only a SKELETON file for the 'Queen Attack' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+export class QueenAttack {
+    constructor(params = { white: [], black: []}) {
+        this.white = params.white;
+        this.black = params.black;
```
tejasbubane: This is part of the implementation & we try to keep stub files free from this - so that users can choose different ways to implement. For example in this case, I may use something other than arrays and have `get` functions for white & black.

Can you please remove this and just have an empty constructor:

```js
constructor() {
  throw new Error("Remove this statement and implement this function");
}
```

abhnvgupta: Alright. I was initially going to keep the constructor empty, but, in the test cases, the constructor expects an object with arrays for white and black, so I thought it would be nice to let the users know that.

SleeplessByte: We understand the reasoning 😄 👍, it's just that with leaving it empty, students can choose to forward the object to a helper function, and are not _led_ by the initial naming etc.

PR #774, opened 2019-10-23
add hangman exercise

### LINKED ISSUE #773: Add new exercise: Hangman

seeksort:
Closes #773 

The problem specification mentions using functional reactive programming to solve this problem, and recommends installing an FRP library. This didn't sound like a good idea given that all exercises use the same dependencies. 

Additionally, the [Python version](https://github.com/exercism/python/tree/master/exercises/hangman) does not use FRP. The only other exercises I saw that have implemented this exercise are C#, F#, and OCaml. The first two had their own version of observables in their common libraries, while OCaml is a functional language.

I went ahead and implemented the exercise without FRP, but I'm interested if anyone has any ideas on implementing FRP without a library. I know that observables are "streams" that emit messages when activities occur, but that's the extent of my FRP knowledge.


SleeplessByte:
> I went ahead and implemented the exercise without FRP, but I'm interested if anyone has any ideas on implementing FRP without a library. I know that observables are "streams" that emit messages when activities occur, but that's the extent of my FRP knowledge.

I think the `react` exercise uses RP too! I think [this](https://stackoverflow.com/a/34496621) answer on stackoverflow is pretty decent.

As for providing a library, which we don't to for react, we can go the `lib` folder approach. This is _build into_ our `scripts`:

https://github.com/exercism/javascript/blob/master/scripts/helpers.js#L77-L80

I [worked on](https://github.com/SleeplessByte/unobservable) a somewhat spec compliant observable thing in ~350 bytes, which we could include? All I would need to do (or you, if you find it interesting) is rolling the output up into a single file.

seeksort:
That's a good SO link, thanks! I've also been doing some reading into FRP and going through the react exercise, which has been helpful. I'm going to try to take a crack at using your observable code in the exercise. I'll let you know if I have any questions!

exercises/hangman/example.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export class Hangman {
+  constructor(word) {
+    this.remaining_guesses = 9
+    this.status = GAME_STATUS.ONGOING
+    this.word = word
+    this.guesses = []
+    this.maskedWord = ''
+    for (var i = 0; i < this.word.length; i++) {
```
SleeplessByte: Instead, perhaps use `'_'.repeat(this.word.length)`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat

exercises/hangman/example.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  getStatus() {
+    return this.status
+  }
+
+  getRemainingGuesses() {
+    return this.remaining_guesses
+  }
+}
```
SleeplessByte: Can you add an empty newline at the end? I think some of the CI will otherwise complain.

exercises/hangman/hangman.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  getStatus() {
+    throw new Error("Remove this statement and implement this function");
+  }
+
+  getRemainingGuesses() {
+    throw new Error("Remove this statement and implement this function");
+  }
+}
```
SleeplessByte: Same here

exercises/hangman/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+[Functional Reactive Programming][frp] is a way to write interactive
+programs. It differs from the usual perspective in that instead of
+saying "when the button is pressed increment the counter", you write
+"the value of the counter is the sum of the number of times the button
+is pressed."
+
+Implement the basic logic behind hangman using functional reactive
+programming.  You'll need to install an FRP library for this, this will
+be described in the language/track specific files of the exercise.
```
SleeplessByte: This means we should add a HINT file. We can just say that _no_ library is necessary.

seeksort: I've reworded this sentence to indicate that no library is necessary, and added the word "optionally" with regard to FRP. Let me know what you think.

seeksort:
@SleeplessByte Unfortunately, I've had trouble finding enough spare time to try to work through an FRP implementation with the library you posted. I saw that you left comments, therefore I went ahead and responded to those on the implementation as it is now. Definitely feel free to modify if you have an idea for FRP, but I didn't want to leave this PR hanging indefinitely.

junedev:
I am closing this because it is stale. If someone else picks up the exercise, they can still refer to this.

PR #775, opened 2019-10-23
fix lint file; lint command was broken

seeksort:
I noticed that the lint command `npx @babel/node scripts/test` wasn't working. I copied a fix from the test script, as helpers.js relies on the same variables for both lint and test.


SleeplessByte:
Ah yes. Thank you!

Issue #776, opened 2019-10-30
[Flatten Array] Impl Class does not make sense


gabriel376:
**Current Implementation**
```Javascript
// flatten-array.spec.js
import { Flattener } from './flatten-array.js';

describe('FlattenArray', () => {
  const flattener = new Flattener();
  test('flattens a nested list', () => {
    expect(flattener.flatten([[]])).toEqual([]);
  });
}

// flatten-array.js
export class Flattener {
  flatten(array) {
    throw new Error("Remove this statement and implement this function");
  }
}
```

**Proposed Implementations**
1) Use constructor argument
```Javascript
// flatten-array.spec.js
import { Flattener } from './flatten-array.js';

describe('FlattenArray', () => {
  test('flattens a nested list', () => {
    const flattener = new Flattener([[]]);
    expect(flattener.flatten()).toEqual([]);
  });
}

// flatten-array.js
export class Flattener {
  constructor(array) {
    this.array = array;
  }
  flatten() {
    throw new Error("Remove this statement and implement this function");
  }
}
```

2) Use static method
```Javascript
// flatten-array.spec.js
import { Flattener } from './flatten-array.js';

describe('FlattenArray', () => {
  test('flattens a nested list', () => {
    expect(Flattener.flatten([[]])).toEqual([]);
  });
}

// flatten-array.js
export class Flattener {
  static flatten(array) {
    throw new Error("Remove this statement and implement this function");
  }
}
```

3) Use plain function
```Javascript
// flatten-array.spec.js
import { flatten } from './flatten-array.js';

describe('FlattenArray', () => {
  test('flattens a nested list', () => {
    expect(flatten([[]])).toEqual([]);
  });
}

// flatten-array.js
export function flatten() {
  throw new Error("Remove this statement and implement this function");
}
``` 

I would go with 3)

SleeplessByte:
I think it currently makes sense, but I also _prefer_ your version three 😄.

@tejasbubane what do you think?

gabriel376:
Regardless of what we decide, we should keep some consistency across exercises regarding naming, structure, etc
See [sieve](https://github.com/exercism/javascript/blob/master/exercises/sieve/sieve.js), for instance, that uses plain function.
Otherwise, if the exercise is about classes, makes sense to work with classes.

tejasbubane:
> Regardless of what we decide, we should keep some consistency across exercises regarding naming, structure, etc
[...]

PR #779, opened 2019-11-01
use plain function for flatten-array exercise (#776)

gabriel376:
Use plain function instead of class on flatten-array exercise (closes https://github.com/exercism/javascript/issues/776)


gabriel376:
thank you!

PR #782, opened 2019-11-11
Add test to detect invalid rna sequence.

### LINKED ISSUE #781: Add extra test about invalid codon to exercise Protein translation
Many students use a regex to extract the codons from the rna. This approach is correct but many of them use `rna.match(/.{3}/g)` instead of `rna.match(/.{1,3}/g)`, what doesn't detect an incorrect sequence of RNA. If translate the codon e.g. "AUGOO", it will not throw an error with {3}, but there is an invalid codon "OO" it wont be matched. With {1,3} "OO" will be matched and error will be thrown. Kudos to @futur8 for catch that here: https://exercism.io/solutions/868dfebd87324127b311fa2215ad3eca/comments/162449.

WebCu:
Fix #781

### DIFF
--- a/exercises/protein-translation/protein-translation.spec.js
+++ b/exercises/protein-translation/protein-translation.spec.js
@@ -56,4 +56,8 @@ describe('ProteinTranslation', () => {
   xtest('Invalid codon throws error', () => {
     expect(() => translate('LOL')).toThrow(new Error('Invalid codon'));
   });
+
+  xtest('Invalid codon throws error', () => {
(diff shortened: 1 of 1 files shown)


tejasbubane:
Did you manage to port this [upstream](https://github.com/exercism/problem-specifications)? Should we wait for it to go upstream or merge it here? @SleeplessByte

SleeplessByte:
Let's just merge it here.

SleeplessByte:
Thank you @WebCu .


@tejasbubane I'll write this down somewhere just in case upstream "opens up" again.

PR #787, opened 2019-11-26
Updated 'Raindrops' tests

ovidiu141:
Synced tests with canonical data.

### DIFF
--- a/exercises/raindrops/raindrops.spec.js
+++ b/exercises/raindrops/raindrops.spec.js
@@ -9,41 +9,44 @@ describe('Raindrops', () => {
 
   xtest('the sound for 7 is Plong', () => expect(convert(7)).toEqual('Plong'));
 
-  xtest('the sound for 6 is Pling', () => expect(convert(6)).toEqual('Pling'));
+  xtest('the sound for 6 is Pling as it has a factor 3', () =>
+    expect(convert(6)).toEqual('Pling'));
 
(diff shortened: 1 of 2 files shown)


exercises/raindrops/raindrops.spec.js
```diff
@@ -1,35 +1,49 @@
...
+  xtest('the sound for 3 is Pling', () => expect(convert(3)).toEqual('Pling'));
 
-  xtest('converts 5', () => expect(convert(5)).toEqual('Plang'));
+  xtest('the sound for 5 is Plang', () => expect(convert(5)).toEqual('Plang'));
 
-  xtest('converts 7', () => expect(convert(7)).toEqual('Plong'));
+  xtest('the sound for 7 is Plong', () => expect(convert(7)).toEqual('Plong'));
 
-  xtest('converts 6', () => expect(convert(6)).toEqual('Pling'));
+  xtest('the sound for 6 is Pling', () => expect(convert(6)).toEqual('Pling'));
```
tejasbubane: We would like to stick to problem-specifications as much as possible which includes exact descriptions like in this case: `the sound for 6 is Pling as it has a factor 3`.

A lot of thought goes into the problem-specifications and we would like to use that in the tracks. Please change all test-descriptions to be same as problem-specifications. Also keep the same order - `8` before `9`.

PR #788, opened 2019-11-27
Update 'Clock' tests

ovidiu141:
Synced tests with canonical data (added two missing tests, regrouped tests).

### DIFF
--- a/exercises/clock/clock.spec.js
+++ b/exercises/clock/clock.spec.js
@@ -153,67 +153,67 @@ describe('Clock', () => {
 
   describe('Compare two clocks for equality', () => {
     xtest('clocks with same time', () => {
-      expect(new Clock(15, 37).equals(new Clock(15, 37))).toBeTruthy();
+      expect(new Clock(15, 37).equals(new Clock(15, 37))).toBe(true);
     });
 
     xtest('clocks a minute apart', () => {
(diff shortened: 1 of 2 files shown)


exercises/clock/clock.spec.js
```diff
@@ -70,140 +70,150 @@ describe('Clock', () => {
...
-      });
+    xtest('clocks with negative hours and minutes that wrap', () => {
+      expect(new Clock(18, 7).equals(new Clock(-54, -11513))).toBeTruthy();
+    });
 
-      xtest('clocks with negative hours and minutes that wrap', () => {
-        expect(new Clock(18, 7).equals(new Clock(-54, -11513))).toBeTruthy();
-      });
+    xtest('full clock and zeroed clock', () => {
+      expect(new Clock(24, 0).equals(new Clock(0, 0))).toBeTruthy();
```
tejasbubane: `toBeTruthy()` uses `!!` to check for `truthy` values. Which means things like non-empty strings, arrays would pass the test. We would like to check for the value to be exactly equal to `true`. Can you please change **all** `toBeTruthy()` to `toBe(true)` & `toBeFalsy()` to `toBe(false)`?

Ref: https://stackoverflow.com/a/32767435/3507206

PR #791, opened 2019-12-04
sync word-count tests with canonical data version 1.4.0

ovidiu141:
Updated tests, removed class structure and updated the example.

@tejasbubane I removed a test about handling properties of JavaScript Object’s prototype. Depending on the focus of the language track, having this test might be a good idea (however this test is not preset in the canonical data). What do you think about this, should we keep it?


SleeplessByte:
It was an implementation test so we don't need it :)

PR #812, opened 2019-12-28
Sync 'Wordy' tests with canonical-data 1.5.0

ovidiu141:


exercises/wordy/example.js
```diff
@@ -1,47 +1,31 @@
...
-    });
-    return result;
+  if (!/[+/*-]+/.test(question)) {
+    throw new Error('Unknown operation');
   }
-}
 
-export { Wordy as WordProblem, ArgumentError };
+  try {
+    return eval(question);
```
SleeplessByte: Whilst the example file is never meant to be an optimal solution, I am not sure if we should be having `eval` in here -- for the simple reason that people might copy and use it and it will start discussions not meant for this exercise.

ovidiu141:
> Can you merge in master, and run the CI locally. If it passes, please ping here that it passes.
> You may then self-squash-merge this in, if you wish.

CI passes locally...

PR #831, opened 2020-01-19
Keep version when updating package json

### LINKED ISSUE #825: Sync script deletes exercise versions
Our CI scripts compare [`package.json` file in root](https://github.com/exercism/javascript/blob/master/package.json) with the one in each exercise. But `version` in package.json is for book-keeping to keep track of [canonical versions](https://github.com/exercism/problem-specifications/blob/63bb34cde0d4e262e3f34b5dc2c1f4c6014e527e/exercises/hello-world/canonical-data.json#L3) & hence is different for each exercise. The CI script deletes field from JSON before comparing.

When we change/update any package, we sync the root file with all exercises using [sync script](https://github.com/exercism/javascript/blob/master/scripts/sync). This uses the same helper function above which deletes the field which results in deletion of all versions after sync. Issue was first noticed in https://github.com/exercism/javascript/pull/723 and effect can be seen here: https://github.com/exercism/javascript/pull/824/files#diff-29adfdb706266f78376d86b88aed872aL3

paparomeo:
Keeps the version of the assignment when updating `package.json` files with the sync script.

Fixes a couple of exercises that were slightly inconsistent due to manual fixes.

Fixes #825.

### DIFF
--- a/exercises/scale-generator/package.json
+++ b/exercises/scale-generator/package.json
@@ -1,7 +1,7 @@
 {
   "name": "exercism-javascript",
-  "description": "Exercism exercises in Javascript.",
   "version": "2.0.0",
+  "description": "Exercism exercises in Javascript.",
   "author": "Katrina Owen",
   "private": true,
   "repository": {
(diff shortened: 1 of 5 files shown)


tejasbubane:
@paparomeo Sorry for the long time. I somehow missed notification for this PR.

PR #837, opened 2020-03-06
Update fetch-configlet to fix Travis build failures

JesseSingleton:
## Issue
Travis builds are failing because of fetch-configlet failures, as
tracked in exercism/configlet#173. This
patch is a copy of exercism/cpp#344, and exercism/haskell#898 which
fixed the cpp and haskell track

## Fix
This makes the test case-insensitive rather than simply fixing the
case, by switching to perl. This is taken from
exercism/sml@a721a5c.
 master (#898)

### DIFF
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -37,8 +37,11 @@ case $(uname -m) in
 esac)
 
 
-VERSION="$(curl --silent --head $LATEST | awk -v FS=/ '/Location:/{print $NF}' | tr -d '\r')"
-URL=https://github.com/exercism/configlet/releases/download/$VERSION/configlet-$OS-${ARCH}.$EXT
+VERSION="$(curl --head --silent $LATEST | perl -nE 's/\R//g; /^Location:/i && print [split/\//]->[-1]')"
+URL=https://github.com/exercism/configlet/releases/download/$VERSION/configlet-$OS-${ARCH}.tgz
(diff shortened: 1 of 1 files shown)


bin/fetch-configlet
```diff
@@ -37,8 +37,11 @@ case $(uname -m) in
 esac)
 
 
-VERSION="$(curl --silent --head $LATEST | awk -v FS=/ '/Location:/{print $NF}' | tr -d '\r')"
-URL=https://github.com/exercism/configlet/releases/download/$VERSION/configlet-$OS-${ARCH}.$EXT
+VERSION="$(curl --head --silent $LATEST | perl -nE 's/\R//g; /^Location:/i && print [split/\//]->[-1]')"
+URL=https://github.com/exercism/configlet/releases/download/$VERSION/configlet-$OS-${ARCH}.tgz
```
SleeplessByte:
```suggestion
URL=https://github.com/exercism/configlet/releases/download/$VERSION/configlet-$OS-${ARCH}.$EXT
```

This is important because we need this to work on Windows too (see the rest of the code).

JesseSingleton:
@SleeplessByte Thanks for the feedback, good catch. I've updated my PR, keen to get the build working again!

JesseSingleton:
Closing as #841 fixes the issue with a simpler change.

PR #839, opened 2020-03-08
improve zipper tests

### LINKED ISSUE #838: Tests for Zipper are far from sufficient
Reference: https://exercism.io/mentor/solutions/962ebabf19e848e5bceed0533f1898ad#discussion-post-605355

We should add a test similar to:

```js
test('traversal returns a new Zipper object', () => {
    expect(zipper.value()).toEqual(1);
    let left = zipper.left();
    expect(zipper.value()).toEqual(1);
    expect(left.value()).toEqual(2);
  })
```

This seems to be a pretty key thing to not test for.

I don't actually think the students **exact** implementation is super critical here.  What they did was implement a single object with a "cursor" that kept track of where it was in the tree and acted accordingly.  So that calling `left` or `right` would merely update the cursor position...

Which 100% works since the current tests never require that a new object be generated.

joshgoebel:
Resolves #838

### DIFF
--- a/exercises/zipper/zipper.spec.js
+++ b/exercises/zipper/zipper.spec.js
@@ -27,6 +27,36 @@ describe('Zipper', () => {
     zipper = Zipper.fromTree(t1);
   });
 
+  test('left returns a new Zipper', () => {
+    let left = zipper.left();
+    expect(left).not.toBe(zipper)
+  })
+
+  test('right returns a new Zipper', () => {
+    let right = zipper.right();
+    expect(right).not.toBe(zipper)
+  })
+
+  test('setValue returns a new Zipper', () => {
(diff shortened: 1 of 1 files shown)


yyyc514:
@SleeplessByte Like this?

SleeplessByte:
Yeah, this looks fine. Maybe also add the README change through a hint!

yyyc514:
I'm not sure exactly what you want it to say, but you're welcome to edit it and add whatever.

SleeplessByte:
It's in https://github.com/exercism/javascript/issues/838#issuecomment-596106538 . Basically explain the JS implementation (tree) yadayada

SleeplessByte (approved):

PR #841, opened 2020-03-08
Fix fetch-configlet script to get 'location' header

ovidiu141:
## Issue
Travis builds are failing because of fetch-configlet failures, as
tracked in exercism/configlet#173. This
patch is a copy of exercism/cpp#344, and exercism/haskell#898 which
fixed the cpp and haskell track

### DIFF
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -37,7 +37,7 @@ case $(uname -m) in
 esac)
 
 
-VERSION="$(curl --silent --head $LATEST | awk -v FS=/ '/Location:/{print $NF}' | tr -d '\r')"
+VERSION="$(curl --silent --head $LATEST | awk -v FS=/ '/location:/{print $NF}' | tr -d '\r')"
 URL=https://github.com/exercism/configlet/releases/download/$VERSION/configlet-$OS-${ARCH}.$EXT
 
 case $EXT in


SleeplessByte:
@JesseSingleton please see this. It's a far simpler change it seems.

@ovidiu141 can we just make that grep case insensitive? Header names are case insensitive and that way we don't need to change it again if it changes again!

JesseSingleton:
Happy to use this as it's simpler, I'll close my PR in lieu of this change.

SleeplessByte:
@JesseSingleton I am still looking to make this case insensitive. If you or @ovidiu141 have input: highly appreciated ❤️

JesseSingleton:
@SleeplessByte My PR used `perl` and was taken from another exercism repo, I'm not familiar with `awk` but something along these  [lines](https://stackoverflow.com/questions/44931041/case-insensitive-string-matching-in-awk) perhaps? Unless @ovidiu141 had any input.

ovidiu141:
> @SleeplessByte My PR used `perl` and was taken from another exercism repo, I'm not familiar with `awk` but something along these [lines](https://stackoverflow.com/questions/44931041/case-insensitive-string-matching-in-awk) perhaps? Unless @ovidiu141 had any input.

@JesseSingleton I'm also not very familiar with `awk`, but I think that't the way to go -> using `toLower`. You can make the change yourself if you want..

tejasbubane:
Wow! Did Github just change the HTTP header? `Location` is pretty standard right?

ovidiu141:
> Wow! Did Github just change the HTTP header? `Location` is pretty standard right?

Well, I've just discovered this, inside the [HTTP 2 specs](https://tools.ietf.org/html/rfc7540):

Just as in HTTP/1.x, header field names are strings of ASCII characters that are compared in a case-insensitive fashion. However, header field names MUST be converted to lowercase prior to their encoding in HTTP/2. A request or response containing uppercase header field names MUST be treated as malformed

PR #851, opened 2020-04-29
Update space age - example.js

JesseSingleton:
Across the site `Number` is generally recommended over `parse....` in mentor notes. It's mentioned in this exercise and in others, so I'm updating this example to use `Number`. If there's a specific reason to use it in this case please let me know. 

Perhaps there are more examples where this is the case.


SleeplessByte:
Thank you!

PR #852, opened 2020-04-30
Disable npm audit in all exercises

### LINKED ISSUE #848: Add notification about npm output
`npm` outputs very scary looking "vulnerability" messaging, multiple times, all the way down the dependency tree, when there is a CVE for such dependency. About 100% of the time, these warnings are **not** applicable to the use-case of this track and can safely be ignored. 

Add a message about this in the instructions to avert issues like https://github.com/exercism/exercism/issues/5212.

tejasbubane:
Closes #848

https://docs.npmjs.com/misc/config#audit

Add `.npmrc` to switch audit check off after `npm install`.

Sadly this config is not allowed in `package.json` - so had to create new `.npmrc` file.

### DIFF
--- a/scripts/helpers.js
+++ b/scripts/helpers.js
@@ -4,11 +4,9 @@
  */
 
 const shell = require('shelljs');
-const fs = require('fs');
 
 const exerciseDirs = shell.ls('-d', 'exercises/*');
 
-const config = JSON.parse(fs.readFileSync('config.json'))['exercises'];
 const assignments = exerciseDirs.map(dir => dir.split('/')[1])
       .filter(exercise => !exercise.deprecated);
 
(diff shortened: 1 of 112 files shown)


scripts/helpers.js
```diff
@@ -4,11 +4,9 @@
  */
 
 const shell = require('shelljs');
-const fs = require('fs');
 
 const exerciseDirs = shell.ls('-d', 'exercises/*');
 
-const config = JSON.parse(fs.readFileSync('config.json'))['exercises'];
```
tejasbubane, line 11: eslint told me these variables were unused.

SleeplessByte: I think we used it and then you made the scripts better ;)

PR #861, opened 2020-05-26
Fix spec for complex-numbers exercise

burennto:
Specs were failing because `abs`, `conj`, `exp` are defined as functions in the current exercise template, not getters.

### DIFF
--- a/exercises/complex-numbers/complex-numbers.js
+++ b/exercises/complex-numbers/complex-numbers.js
@@ -32,15 +32,15 @@ export class ComplexNumber {
     throw new Error("Remove this statement and implement this function");
   }
 
-  abs() {
+  get abs() {
     throw new Error("Remove this statement and implement this function");
   }
 
-  conj() {
+  get conj() {
     throw new Error("Remove this statement and implement this function");
   }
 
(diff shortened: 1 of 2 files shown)


tejasbubane:
Hey @burennto thanks for spotting this and sorry for my disastrously late review. The reason our specs are failing are because they run the spec file which you changed against [example implementation](https://github.com/exercism/javascript/blob/bed3424fd266b83904e5d59889984e5303c6678f/exercises/complex-numbers/example.js).

To fix this, you need to revert your changes from the spec file and instead change the exercise template which we serve students when they download the exercise. That template is [here](https://github.com/exercism/javascript/blob/bed3424fd266b83904e5d59889984e5303c6678f/exercises/complex-numbers/complex-numbers.js#L35-L45).

SleeplessByte:
@burennto are you still interested in working on this?

burennto:
Hi @tejasbubane and @SleeplessByte , as per instructions I reverted the spec changes and pushed a new commit that updated the exercise template.

SleeplessByte (approved):
Thanks! Looks good.

PR #862, opened 2020-06-02
add exercise dominoes

chauchakching:
:)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1454,6 +1454,16 @@
         "filtering",
         "games"
       ]
+    },
+    {
+      "slug": "dominoes",
+      "uuid": "3d155146-f3a9-43c6-a662-2a0679949a12",
+      "core": false,
+      "unlocked_by": "bob",
+      "difficulty": 6,
+      "topics": [
+        "graph_theory"
+      ]
     }
   ]
 }
--- a/exercises/dominoes/dominoes.js
+++ b/exercises/dominoes/dominoes.js
@@ -1,0 +1,8 @@
+//
+// This is only a SKELETON file for the 'Dominoes' exercise. It's been provided as a
+// convenience to get you started writing code faster.
+//
+
+export const chain = () => {
+  throw new Error("Remove this statement and implement this function");
+};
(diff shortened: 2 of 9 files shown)


exercises/dominoes/dominoes.js
```diff
@@ -0,0 +1,8 @@
+//
+// This is only a SKELETON file for the 'Yacht' exercise. It's been provided as a
```
tejasbubane, line 2: `Dominoes` exercise

exercises/dominoes/dominoes.spec.js
```diff
@@ -0,0 +1,129 @@
+// import { chain } from './dominoes'
```
tejasbubane, line 1: remove comment

exercises/dominoes/dominoes.spec.js
```diff
@@ -0,0 +1,129 @@
...
+    it('Should have a chain', () => {
+      expect(result).not.toBe(null)
+    })
+  } else {
+    it('Should not have a chain', () => {
+      expect(result).toBe(null)
+    })
+  }
+
+  if (!expected) return
```
tejasbubane, line 18: This is not required, `expect` will raise error and rest of this function will not be executed.

chauchakching: But the rest of the tests are irrelevant if `result` is expected to be `null`

config.json
```diff
@@ -1454,6 +1454,16 @@
         "filtering",
         "games"
       ]
+    },
+    {
```
SleeplessByte, line 1458: @tejasbubane do we still use the script to auto sort this file? If not, this is fine :)

exercises/dominoes/dominoes.spec.js
SleeplessByte: Instead, create two helper functions:
```
runTestsExpectingChain
runTestsExpectingNull
```

or similar.

It will remove the rather cryptic boolean value, and make it clear that the tests for "a chain" are different than "no chain".

config.json
tejasbubane: I don't think we have anything to sort it ATM - I think this is fine.

PR #874, opened 2020-10-16
Add GitHub Actions

### LINKED ISSUE #871: Moving from Travis to GitHub Actions
Hello 🙂 

Over the last few months we've been transferring all our CI from Travis to [GitHub Actions (GHA)](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions). We've found that GHA are easier to work with, more reliable, and much much faster. 

Based on our success with GHA and increasing intermittent failures on Travis, we have now decided to try and remove Travis from Exercism's org altogether and [shift everything to GHA](https://github.com/exercism/exercism/issues/5396). This issue acts as a call to action if your track is still using Travis. 

For most CI checks this should be a transposing from Travis' syntax to [GHA syntax](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions), and hopefully quite straightforward (see [this PR for an example](https://github.com/exercism/csharp/pull/1338/files)).  However, if you do encounter any issues doing this, please ask on Slack where lots of us now have experience with GHA, or post a comment here and I'll tag relevant people. This would also make a good Hacktoberfest issue for anyone interested in making their first contribution 🙂 

If you've already switched this track to GHA, please feel free to close this issue and ignore it.

Thanks!

SleeplessByte:
Fixes #871 

Please review @exercism/github-actions 

---

The following **does not** run `path/to/script` through babel node. It's because `@babel/node` doesn't have `"bin"` set-up correctly (imo) in `package.json`. 

```bash
npx @babel/node path/to/script
```

The currently correct way to achieve _what was intended_:

```bash
npx -p @babel/core -p @babel/node babel-node path/to/script
```

The _downside_ of that is that _even_ when `@babel/node` and/or `@babel/core` are in `package.json`, they're _still_ re-installed. 

This has been solved by not relying on `npx` to install the packages, and instead use the ones from `package.json`.


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,37 @@
...
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Install project dependencies
+      run: | 
+        npm ci
+        bin/fetch-configlet
```
SaschaMann:
```suggestion


    - name: Fetch configlet
      uses: exercism/github-actions/configlet-ci@master
```

This potentially saves you work on updating these

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: Run exercism/configlet Linter
+      run: bin/configlet lint .
```
SaschaMann:
```suggestion
      run: configlet lint .
```

SaschaMann (approved):

.github/workflows/node.js.yml
SleeplessByte: Good idea. Only downside is: local thing is then no-longer necessarily the same as running it on Github Actions.

SaschaMann: Upside of that: the GHA one is always the latest release whereas the local one may be outdated (if it ever changes, currently it doesn't)

SleeplessByte: (I more meant that the local _fetch script_ might be outdated, but alas).

SleeplessByte:
@tejasbubane when you have a sec, please merge; working on optimisations on this

cmccandless (changes requested):
Make sure to change the status badge at the top of the README as well:
https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge

SleeplessByte:
Good idea!

iHiD (approved):

SleeplessByte:
@tejasbubane I know you're quite occupied; let's just get this merged to get rid of Travis and we can make other edits via PRs later.

PR #883, opened 2020-10-20
Speed up lint

SleeplessByte:


iHiD (approved):

ErikSchierboom (approved):
This seems like a sensible change. Only running one shell command instead of N commands.

SleeplessByte:
Yeah, basically the gain here is from node not having to boot (at least 250ms * n exercises)

Issue #901, opened 2020-12-24
two-bucket: fix implemenation


SleeplessByte:
Don't work on this issue until #895 is merged, or use that PR as a base branch.

When `two-bucket` was synced, it became evident that some of the tests cases are not passing, at all. Fix the implementation so these test cases pass and replace `test.skip` with `xtest`: https://github.com/exercism/javascript/blob/774fb12c4d1f492af9658ece4eec515bf92552e7/exercises/two-bucket/two-bucket.spec.js

slaymance:
Hey @SleeplessByte I noticed the tests throwing for this exercise and submitted a fix in [PR 905](https://github.com/exercism/javascript/pull/905). We'll still need another PR to update the example solution since it's failing a couple of the new tests, and running into an infinite loop on a couple of them. With this PR, though, we can at least get the tests working for students.

PR #905, opened 2020-12-28
Two Bucket: fix errant tests

slaymance:
This PR addresses part of the problem tracked in [Issue #901](https://github.com/exercism/javascript/issues/901). A couple of test cases were added that were expected to throw yet were invoked immediately in the spec file. This PR wraps those invocations in an anonymous function.

The example solution provided does not pass the test cases as it stands and will need to be addressed in a separate PR. This is an immediate fix so the tests are working for students.

### DIFF
--- a/exercises/two-bucket/two-bucket.spec.js
+++ b/exercises/two-bucket/two-bucket.spec.js
@@ -59,7 +59,7 @@ describe('TwoBucket', () => {
   });
 
   describe('Measure using bucket one of size 2 and bucket two of size 3', () => {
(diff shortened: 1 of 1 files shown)


SleeplessByte (changes requested):
Can you revert the test.skip changes? They are there on purpose to keep the ci passing!

SleeplessByte (approved):

SleeplessByte:
Thanks for this @slaymance . Small oversight on my part 😅

PR #907, opened 2020-12-30
Adding square-root exercise

### LINKED ISSUE #906: Add new exercise: square-root
[problem specification ](https://github.com/exercism/problem-specifications/tree/master/exercises/square-root)| [description](https://github.com/exercism/problem-specifications/blob/master/exercises/square-root/description.md) | [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/square-root/canonical-data.json)

evelynstender:
Closes #906

### DIFF
--- a/config.json
+++ b/config.json
@@ -231,9 +231,9 @@
       "slug": "square-root",
       "uuid": "31c7725e-def0-4dc9-88a0-e19afb9bfb06",
       "core": false,
-      "unlocked_by": "space-age",
-      "difficulty": 1,
-      "topics": ["math"]
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": ["bitwise_operations", "algorithms", "loops", "math"]
     },
     {
       "slug": "reverse-string",
--- a/config.json
+++ b/config.json
@@ -228,6 +228,14 @@
       "topics": ["booleans", "integers", "logic"]
     },
     {
+      "slug": "square-root",
+      "uuid": "31c7725e-def0-4dc9-88a0-e19afb9bfb06",
+      "core": false,
+      "unlocked_by": "space-age",
+      "difficulty": 1,
+      "topics": ["math"]
+    },
+    {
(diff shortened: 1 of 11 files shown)


exercises/square-root/example.js
```diff
@@ -0,0 +1,3 @@
+export const squareRoot = (radicand = 0) => {
+  return Math.sqrt(radicand);
```
SleeplessByte, line 2: I think the point is to _not_ use a built-in function 😅

SleeplessByte (commented):
I think we should clarify we don't want to use `Math.sqrt`. We might check other languages that have implemented this exercise to confirm that other tracks also disallow using built-in functions.

If this is the case:

- add `.meta/hints.md` and disallow `Math.sqrt`
- `bin/configlet generate . --only square-root` again to update the README.md
- update the example solution

evelynstender:
Oh yeah! That makes sense! I didn't realise we couldn't use `Math.sqrt()`. I didn't see it anywhere haha

I will update! Thanks!

config.json
```diff
@@ -227,6 +227,14 @@
       "difficulty": 1,
       "topics": ["booleans", "integers", "logic"]
     },
+    {
+      "slug": "square-root",
+      "uuid": "31c7725e-def0-4dc9-88a0-e19afb9bfb06",
+      "core": false,
+      "unlocked_by": null,
```
evelynstender, line 234: I'm not sure which exercise should unlock this (if any)

evelynstender:
@SleeplessByte I updated the example, added the `hint.md` and updated the readme file. I'm not sure which exercise should unlock this (if any), so I left it as `null`

SleeplessByte (approved):
Perfect!

SleeplessByte:
@evelynstender in v2, _only_ core exercises can unlock anything, and since this is a reimplementation and mathy exercise, it will never become core.

PR #932, opened 2021-01-13
New Exercise: Promises

slaymance:
I thought it would be useful to add an exercise which focuses on `Promise`s since they're such a core part of JS development. My idea was to ask learners to implement a function that behaves like NodeJS's `util.promisify`, as well as the `Promise` static helper methods in native JS:

- `Promise.all`
- `Promise.allSettled`
- `Promise.race`
- `Promise.any`

I've included some tests for each function as a starting point and an example implementation of all the functions. I'd love some feedback and additional contributions. I think this would make a great language-specific exercise!


SleeplessByte:
@slaymance before reviewing, can you make the CI's pass?

SleeplessByte:
Thanks <3

SleeplessByte (approved):
Alright. This is looking like a great addition to the _practice_ exercises. I have no things you should change. I like it.

In order to merge (sorry!) two more things:

1. [x] move the `promises` folder into `practice`
2. [x] then rebase (watch out: `package.json` has a key `practice` too, now)
3. [ ] run `npm i` in the root
4. [ ] run  `ASSIGNMENT=practice/promises npx babel-node scripts/sync` in the root

As far as I'm concerned, it's mergable right after.

SleeplessByte:
@slaymance I performed the tasks, added introduction, added .meta/config files and added you as the Author. 

Thanks for contributing :)

Issue #896, opened 2020-12-17
forth: implement two new tests


SleeplessByte:
Don't work on this issue until #895 is merged, or use that PR as a base branch.

The exercise _forth_ has two new test cases that are not implemented. Your mission, should you accept it, is to change the `example.js` implementation so that these two tests would pass. The current `example.js`  implementation has lazy evaluation but no special handling for "rebinding" words. 

## Instructions

Flip the following test from `false` to `true`:

```toml
# can use different words with the same name
"ac12aaaf-26c6-4a10-8b3c-1c958fa2914c" = false
```

Add the following test to the `.spec.js`

```javascript
xtest('can use different words with the same name', () => {
  forth.evaluate(': foo 5 ;');
  forth.evaluate(': bar foo ;');
  forth.evaluate(': foo 6 ;');
  forth.evaluate('bar foo');
  expect(forth.stack).toEqual([5, 6]);
});
```

Flip the following test from `false` to `true`:

```toml
# can define word that uses word with the same name
"53f82ef0-2750-4ccb-ac04-5d8c1aefabb1" = false
```

Add the following test to the `.spec.js`

```javascript
xtest('can define word that uses word with the same name', () => {
  forth.evaluate(': foo 10 ;');
  forth.evaluate(': foo foo 1 + ;');
  forth.evaluate('foo');
  expect(forth.stack).toEqual([11]);
});
```

Make sure the PR CI passes:

```shell
npx babel-node scripts/pr exercises/forth

# Or alternatively
ASSIGNMENT=forth npx babel-node scripts/test
```

slaymance:
Addressed in [PR 934](https://github.com/exercism/javascript/pull/934).

SleeplessByte:
Closed by #934

PR #934, opened 2021-01-14
(896) forth: implement two new tests

slaymance:
This PR addresses the issue captured in [Issue 896](https://github.com/exercism/javascript/issues/896).

Since the example implementation relies on lazy evaluation when a new command is defined, the two new tests were failing because they expected the command to be evaluated immediately at the time that it's defined. To solve this, I simply try to evaluate the newly defined command but revert to lazy evaluation if it fails.


SleeplessByte (approved):
Nice. Thank you for the mental gymnastics implementing this.

Issue #897, opened 2020-12-18
alphametics: optimise implementation for last test


SleeplessByte:
Don't work on this issue until #895 is merged, or use that PR as a base branch.

The last test currently takes > 30s to complete. When fixed, replace `test.skip` with `xtest`.

slaymance:
Addressed in [PR 935](https://github.com/exercism/javascript/pull/935).

PR #935, opened 2021-01-15
alphametics: optimise implementation for last test

slaymance:
This PR addresses [Issue 897](https://github.com/exercism/javascript/issues/897).

I refactored the example solution using [my solution](https://github.com/slaymance/exercism/blob/master/javascript/alphametics/alphametics.js) to alphametics as a guideline. The example implementation struggled a bit with the last test because we would replace each letter with a number every time we wanted to test a permutation. I refactored it so that we sum up the places of each letter and then multiply that sum by a number for each permutation, which is much faster than replacing each letter individually. For example:
```javascript
// If we have the string 'MOM':
{
  'M': 10 ** 2 + 10 ** 0, // === 101
  'O': 10 ** 1 // === 10
}
// and if we want to test M = 9 and O = 2:
{
  'M': 909, // === 101 * 9
  'O': 20 // 10 * 2
}
```
For letters the appear in the total (right side of the equals sign), then their count is subtracted. At the end, we add up all the counts of the letters and check to see if the sum is equal to 0. If so, then that permutation of numbers is our solution.

### DIFF
--- a/exercises/alphametics/alphametics.spec.js
+++ b/exercises/alphametics/alphametics.spec.js
@@ -91,7 +91,7 @@ describe('Solve the alphametics puzzle', () => {
     expect(solve(puzzle)).toEqual(expected);
   });
 
-  test.skip('puzzle with ten letters and 199 addends', () => {
+  xtest('puzzle with ten letters and 199 addends', () => {
     const puzzle =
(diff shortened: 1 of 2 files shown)


SleeplessByte (approved):
Much better solution. Thank you @slaymance

PR #980, opened 2021-02-09
[v3] Move Practice Exercise example files to .meta directory

ErikSchierboom:
The example implementation files for Practice Exercises should be located in the `.meta` directory. Besides being consistent with Concept Exercises, it also has the added benefit that no extra configuration is needed to prevent the example file from being downloaded to the student (as files in the `.meta` directory are never downloaded).

See [the spec](https://github.com/exercism/docs/blob/main/anatomy/tracks/practice-exercises.md#file-example-implementation).

This PR attempts to detect these example files and move them to the exercise's `.meta` directory.

Note that this change will likely break your CI, so feel free to either merge this and fix the build later, or push a fix to this PR.  

## Tracking

https://github.com/exercism/v3-launch/issues/18


SleeplessByte:
@tejasbubane if this is part of your work, we can ignore. If you haven't done it yet, you can merge this to your branch (but should still become `proof.ci.js`)

SleeplessByte:
We'll be doing this ourselves because we want to rename these files too.

PR #1010, opened 2021-02-26
Fix stub for circular-buffer to match test spec

ramitmittal:
The test spec file `circular-buffer.spec.js` tests the `forceWrite` method of `CircularBuffer` class.
However, the skeleton file `circular-buffer.js` provides an empty method with the name of `overwrite`.
I have renamed this method for the sake of consistency.


SleeplessByte (approved):

SleeplessByte:
Thank you! This was an oversight :)

Issue #1017, opened 2021-02-28
[V3] Fix exercise names, fix documents, fix all the things


SleeplessByte:
I know what needs to be done, so... there's that. This issue is just so I can write things down between times I can work on this, and so other people could pick it up should something happen to me (nothing planned!, just in case).

I will also reference this issue from PRs related.

- [x] Lucian's Lucious Lasagna (`basics`): ~#1018~ 
- [x] Annalyn's Infiltration (`booleans`): ~#1020~
- [x] Freelancer Rates (`numbers`): ~#1021~
- [x] Poetry Club Door Policy (`strings`): ~#1043~ ~https://github.com/exercism/javascript-test-runner/pull/36~
- [x] Elyses Enchantments (`arrays`) : ~#1071~
- [ ] Elyses Analytic Enchantments  (`array-analysis`) 
- [x] Elyses Destructured Enchantments (`array-destructuring`)
- [x] Elyses Looping Enchantments (`array-loops`) #1097
- [ ] Elyses Transformative Enchantments (`array-transformations`)
- [x] Translation Service (`promises`) ~#1071~ (needs another issue)

## IF YOU ARE WORKING ON THIS

Please let me know in this issue so I don't work on it. Anything listed above is being worked on by _someone_. I tick them off once there is a PR.

angelikatyborska:
I'm going to work this weekend on "fixing everything" about Elyses Destructured Enchantments

junedev:
Just for reference, `errors` also needs to be fixed as @angelikatyborska noted. In particular it is missing the exercise entry in the global config.json file.

junedev:
I created more descriptive issues for the remaining items on the list above, closing this one.
#1325 
#1326 

That way it is easier for people to see that there are already open issues for these exercises and to add comments there instead of opening a new issue.

PR #1018, opened 2021-02-28
[V3] Lucian's Lucious Lasagna: fix everything around this exercise

SleeplessByte:
# Reviewers, please spell/grammar check

Related #1017 

## Changes

- Rename from concept based exercise (`basics`) to story based exercise `lasagna`
- Add full name to `config.json`
- Add blurb for `basics` to `config.json`
- Add a big welcome block to the stub
- Add design document
- Sync concept documents with exercise documents
- Change concept link names to be descriptive
- Remove duplicate concept links


exercises/concept/lasagna/.meta/exemplar.js
```diff
@@ -0,0 +1,74 @@
...
+// IDEs display.
+//
+// You don't need to write these yourself; it is not expected in idiomatic
+// JavaScript, but some companies and style-guides do enforce them.
+//
+// 💡 You're allowed to completely clear a stub before you get started. Often
+// we recommend using the stub, because they are already set-up correctly to
+// work with the tests, which you can find in ./lasagna.spec.js
+//
+// Good luck preparing some lasagna!
```
neenjaw, line 31: ❤️

neenjaw (approved):
Changes look awesome.  Really like the intro blurbs and the `> 💡 quotes`

exercises/concept/lasagna/.meta/exemplar.js
SleeplessByte: I thought this to be a much better idea for the **first** exercise than putting this in the description. Often students look in the code FIRST, and that's even more apparent as the code is on the left side of the screen. I also don't like this showing up in the description/instructions because it makes no sense if you _don't see the stub_.

neenjaw: Agreed

concepts/basics/about.md
```diff
@@ -2,11 +2,21 @@
 
 JavaScript is a dynamic, prototype based language. It has a few [primitives][mdn-primitive], and everything else is considered an object.
 
-While it is most well-known as the scripting language for Web pages, many non-browser environments also use it, such as Node.js. The language is actively being developed; and because of it's multi-paradigm property, allows for many styles of programming.
+While it is most well-known as the scripting language for Web pages, many non-browser environments also use it, such as Node.js. The language is actively being developed; and because of it's multi-paradigm property, allows for many styles of programming. JavaScript is a dynamic language, supporting object-oriented, imperative, and declarative (e.g. functional programming) styles.
```
ErikSchierboom, line 5: Nice introduction!

concepts/basics/about.md
```diff
@@ -20,7 +30,9 @@ MY_MUTABLE_VALUE_CONSTANT;
 // => { food: "pear" }
 ```
 
-As a rule, we don't mutate values that look like `const SCREAMING_SNAKE_CASE`. When we _must_ to enforce this, use [`Object.freeze(value)`][mdn-object-freeze].
+## Constant value (immutability)
+
+As a rule, on Exercism, and many other organisations and project style guides, don't mutate values that look like `const SCREAMING_SNAKE_CASE`. Technically the values _can_ be changed, but for clarity and expectation managment on Exercism this is discouraged. When this _must_ be enforced, use [`Object.freeze(value)`][mdn-object-freeze].
```
ErikSchierboom, line 35:
```suggestion
As a rule, on Exercism, and many other organisations and project style guides, don't mutate values that look like `const SCREAMING_SNAKE_CASE`. Technically the values _can_ be changed, but for clarity and expectation management on Exercism this is discouraged. When this _must_ be enforced, use [`Object.freeze(value)`][mdn-object-freeze].
```

concepts/basics/about.md
```diff
@@ -34,6 +46,21 @@ MY_MUTABLE_VALUE_CONSTANT;
...
+```javascript
+function add(num1, num2) {
+  return num1 + num2;
+}
+
+add(1, 3);
+// => 4
+```
+
+> 💡 In JavaScript there are _many_ different ways to declare a function. These other ways look different than using the `function` keyword. The track tries to gradually introduce them, but if you already know about them, feel free to use any of them. In most cases, using one of the other isn't better or worse.
```
ErikSchierboom, line 62: Love this suggestion!

concepts/basics/introduction.md
```diff
@@ -22,7 +26,11 @@ MY_FIRST_CONSTANT = 20;
 // => TypeError: Assignment to constant variable.
 ```
 
-In JavaScript, units of functionality are encapsulated in _functions_, usually grouping functions together in the same file if they belong together. These functions can take arguments, and can return a value using the `return` keyword. Methods are invoked using `()` syntax.
+> 💡 In a later Concept exercise the difference between _constant_ assignment / binding and _constant_ value is explored and explained.
```
ErikSchierboom, line 29:
```suggestion
> 💡 In a later Concept Exercise the difference between _constant_ assignment / binding and _constant_ value is explored and explained.
```

exercises/concept/lasagna/lasagna.js
```diff
@@ -0,0 +1,69 @@
+// @ts-check
```
ErikSchierboom, line 1: While I really like everything that's being said in these comments, my only concern would be that people would have to scroll down a long way before seeing any actual code when working on this in the browser. But that might not be an issue 🤷

ErikSchierboom (approved):
LGTM!

ErikSchierboom (commented):
While reviewing https://github.com/exercism/javascript/pull/1021/files#,  I checked the `hints.md` file of this exercise. It looks like the link to the arithmetic operators no longer refers to the right section, which is probably due to a change in the URL structure as MDN.

Old url: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators
New url: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#arithmetic_operators

It might be worth updating the URL for the hints.

exercises/concept/lasagna/lasagna.js
SleeplessByte: Eh. I'm not worried about that. I rather have it in their face the first 3 exercises, than adding it without context and it "just working".

SleeplessByte:
@ErikSchierboom I think MDN didn't get the notion that https://www.w3.org/Provider/Style/URI.html

exercises/concept/lasagna/lasagna.js
ErikSchierboom: I can see how that particularly makes sense for JS given the large number of people relatively new to programming.

SleeplessByte: Yeah. Trying to find the fine line between those two things.

Issue #989, opened 2021-02-21
[V3] Implement new Concept Exercise: Comparison and Conditionals


junedev:
## Getting started
**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md)
- [What are concepts and how they are structured?](https://github.com/exercism/docs/blob/main/anatomy/tracks/concepts.md)
- [What are concept exercises and how they are structured?](https://github.com/exercism/docs/blob/main/anatomy/tracks/concept-exercises.md)

Please also watch the following video:
- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

See the documentation _above_ (general documentation), as well as [How to implement a Concept Exercise in JavaScript](https://github.com/exercism/javascript/blob/main/reference/implementing-a-concept-exercise.md).

## Goal
The goal of this exercise is to teach the student how to compare values in Javascript and then use that in conditional statements.

## Concepts
The following concept pages need to be created. You can combine the two introduction.md files from the concepts and use that as introduction.md for the concept exercise. No need to create different content at this point.
* `comparison`
* `conditionals`

## Learning Objectives
In the concepts the student should learn about the following operators and statements and then practice them in the concept exercise.

### Comparison
* `<`, `>`, `>=`, `<=` for numbers and strings
* strict equality (`===`)
* not equal (`!==`)

### Conditionals
* `if(){}`
* `if(){} else{}`
* `if(){} else if(){} else{}`

## Out of Scope
* Loose equality `==` and truthy/falsy (will be taught once the student learned about type coercion)
* Comparing values of different types
* Shallow/deep comparison of objects
* ternary operator `x ? y : z`
* `switch`, `for`

## About.md of the Concept
If there are no pitfalls, advanced usages etc. that you would like to point out, the about.md file can be a copy of the introduction.md of the concept.

For the `conditionals` concept here some ideas what to mention in the about.md file.
* multiple `else if`
* if statements without a block to execute just one thing (e.g., `if(...) return ...`)

## Prerequisites
* `basics`
* `booleans`
* `numbers`
* `strings`

[...]

PR #1037, opened 2021-03-13
Add concept exercise conditionals

junedev:
Addresses https://github.com/exercism/javascript/issues/989


SleeplessByte:
Exciting stuff!

Once you undraft this and fix the conflicts and still run into prettier, you can add a comment `/format` as the only content of that comment to trigger prettier here:

```
/format
```

The same goes for _integrity_ issues. If the precheck fails and complains that the files are not in sync, you can use `/sync` as the only content of a comment to have a GitHub Action sync all the exercises:

```
/sync
```

SleeplessByte:
Okaaaaay. It wasn't supposed to trigger it, but 🤷🏽 . Luckily everything was already up to date, so nothing happened 🌠

junedev:
Open questions:
1. **Scope**: Currently the ternary operator is out of scope for this exercise. It think it is better to teach it somewhat later because otherwise students might overuse it and create code that is hard to read but makes them feel clever. Should we keep it like it is or include the ternary operator?
2. **Concept naming**: Currently the concept for the if statement is called `conditionals`. Since there will also be `conditionals-switch` and `conditionals-ternary` later on, `conditionals-if` would be a more consistent name. Should I stick with `conditionals` nevertheless?
3. **About.md content**: I included some parts (e.g. on how to write the condition) that are more "general good practice" than JavaScript specifics. I am of course open to removing some or all of those if they are unwanted. Just let me know.
4. **Enforcing if/else if/else**: In these simple functions we are dealing with in the tasks you could always also use early returns instead of practicing how to write else if and else what is the goal of the exercise. I have dealt with that by tweaking the exercise such that not using early returns avoids some duplication and by explicitly mentioning what should be used in case of task 3. Additionally I understand our analysis tooling can also help here. Any other thoughts on how to deal with this issue?

SleeplessByte:
I moved it so you could decide :)

SleeplessByte:
> Scope: Currently the ternary operator is out of scope for this exercise. It think it is better to teach it somewhat later because otherwise students might overuse it and create code that is hard to read but makes them feel clever. Should we keep it like it is or include the ternary operator?

I would even suggest that it's okay if we don't _explicitly_ teach the ternary operator. Its use is often not necessarily more idiomatic, but a stylistic choice. I feel very (positively) strongly about using it, but I know that various style guides discourage or even ban it altogether. Does that mean we should skip it? No, but it does mean its not a worthy concept to put our time in _right now_ ;).

I think it's a good idea to leave it out of this exercise and revisit it some time in the future when we've covered the other bases. 

> Concept naming: Currently the concept for the if statement is called conditionals. Since there will also be conditionals-switch and conditionals-ternary later on, conditionals-if would be a more consistent name. Should I stick with conditionals nevertheless?

Eh. Yes. I agree with you, but we _have_ made the choice that the "basics of x" / "x 101" / "introduction to x" is named "x", and not "x-something". I think this is a great exercise to teach people how to do the majority of conditionals, and thus I think conditionals is aptly named. In fact, any exercise that potentially expects a ternary or switch can probably be solved by someone who has only done _this_ exercise. My vote would be `conditionals` for cross-track consistency.

> About.md content: I included some parts (e.g. on how to write the condition) that are more "general good practice" than JavaScript specifics. I am of course open to removing some or all of those if they are unwanted. Just let me know.

I like this. Let's keep it.

> Enforcing if/else if/else: In these simple functions we are dealing with in the tasks you could always also use early returns instead of practicing how to write else if and else what is the goal of the exercise. I have dealt with that by tweaking the exercise such that not using early returns avoids some duplication and by explicitly mentioning what should be used in case of task 3. Additionally I understand our analysis tooling can also help here. Any other thoughts on how to deal with this issue?

I often would write early returns. We can *definitely* have the Analyzer check for lack of `else` statements and then print out something along the lines of:

```text
Nice. That's an _early return_. For the purpose of the Concept that this exercise aims to teach, try solving this using an `else` statement.
```

If you'd like something like this (you can suggest the copy), can you add it to `design.md` at the bottom? See `lasagna` if you want to know what I mean. I can then point you / aid you / do this for you in the Analyzer. Let me know.

concepts/comparison/about.md
```diff
@@ -1,44 +1,87 @@
...
+'a' < 'above';
 // => true
-// A number is equal to a bigint if they represent the same value.
 
-1.0 == 1n;
+'a' === 'A';
+// => false
+```
+
+You need to be careful when you compare two variables that contain numeric values but are of type string. Due to the dictionary order the result will not be the same as comparing values of type number.
```
SleeplessByte, line 46: Great example, and common pitfall imo.

concepts/comparison/about.md
```diff
@@ -1,44 +1,87 @@
...
+10 < 2;
+// => false
+
+'10' < '2';
+// => true (because "1" comes before "2")
+```
+
+## Strict Equality
+
+You might wonder about the three equal signs for checking equality in JavaScript. `===` represents the check for _strict equality_ which means that no type conversion is performed and values of different types are always unequal.
```
SleeplessByte, line 58:
```suggestion
You might wonder about the three equal signs for checking equality in JavaScript.
`===` represents the check for _strict equality_ which means that no type conversion is performed and values of different types are always unequal.
```

concepts/comparison/about.md
```diff
@@ -1,44 +1,87 @@
...
+1 === 1n;
+// => false
+// The value on the left has type number, the value on the right has type bigint.
+```
+
+Using `===` and `!==` is the recommended way of checking equality in JavaScript.
+
+## Avoiding Implicit Type Conversion
+
+TThere is also `==` and `!=` which represents checking for _loose equality_. You should avoid it because it will apply implicit type conversion before performing the comparison. The outcomes in these cases are hard to predict and sometimes not what you would expect. You can read more about it [here][mdn-loose-equals].
```
SleeplessByte, line 74:
```suggestion
There is also `==` and `!=` which represents checking for _loose equality_.
You should avoid it because it will apply implicit type conversion before performing the comparison.
The outcomes in these cases are hard to predict and sometimes not what you would expect. 
You can read more about it [here][mdn-loose-equals].
```

Typo in the first word.

concepts/comparison/about.md
```diff
@@ -1,44 +1,87 @@
...
+
+```javascript
+0 == false;
 // => true
-// A number is equal to a bigint if they represent the same value.
 ```
 
-There are different outcomes when comparing numbers with different types.
-In general, only two operands of the type `number` can ever be _strictly equal_ (`===`), and the following can be used for _loose equality_ (`==`):
+In theory you can also compare values of different types (e.g., `"1" < 2`). Then the values will be implicitly converted to determine whether the result is true or false. Just as checking for loose equality, this is also not recommended for the same reason as mentioned above.
```
SleeplessByte, line 81:
```suggestion
In theory you can also compare values of different types (e.g., `"1" < 2`).
Then the values will be implicitly converted to determine whether the result is true or false.
Just as checking for loose equality, this is also not recommended for the same reason as mentioned above.
```

concepts/comparison/introduction.md
```diff
@@ -1 +1,69 @@
...
+
+1 === 1.0;
+// => true
+// All numbers are floating-points, so this is different syntax for
+// the exact same value.
+```
+
+## Comparing Strings
+
+In JavaScript the comparison operators above can also be used to compare strings. In that case a dictionary (lexicographical) order is applied. You can find a list of the exact order of all the characters [here][utf-16-list].
```
SleeplessByte, line 33:
```suggestion
In JavaScript the comparison operators above can also be used to compare strings.
In that case a dictionary (lexicographical) order is applied.
You can find a list of the exact order of all the characters [here][utf-16-list].
```

concepts/comparison/introduction.md
```diff
@@ -1 +1,69 @@
...
+'a' < 'above';
+// => true
+
+'a' === 'A';
+// => false
+```
+
+## Strict Equality
+
+You might wonder about the three equal signs for checking equality in JavaScript. `===` represents the check for _strict equality_ which means that no type conversion is performed and values of different types are always unequal.
```
SleeplessByte, line 48:
```suggestion
You might wonder about the three equal signs for checking equality in JavaScript.
`===` represents the check for _strict equality_ which means that no type conversion is performed and values of different types are always unequal.
```

concepts/comparison/links.json
```diff
@@ -1 +1,14 @@
-[]
+[
+  {
+    "url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#relational_operators",
+    "description": "Links to the official documentation for all relational operators"
```
SleeplessByte, line 4:
```suggestion
    "description": "MDN: Relational operators"
```

This makes these in-line with the other links. I'm also hesitant calling it official without denoting the authority (even tho I _feel_ the same way).

concepts/conditionals/about.md
```diff
@@ -1,3 +1,134 @@
...
+if (condition) {
+  // code that is executed if "condition" is true
+} else {
+  // code that is executed otherwise
+}
+```
+
+## Nested If-Statements
+
+To nest another condition into the `else` statement you can use `else if`. Note that there is no `elseif` keyword in JavaScript. It is literally just `else` followed by another `if` statement.
```
SleeplessByte, line 25:
```suggestion
To nest another condition into the `else` statement you can use `else if`.
Note that there is no `elseif` keyword in JavaScript.
Instead write `else` followed by another `if` statement.
```

I try to avoid `just`, in this context.

concepts/conditionals/about.md
```diff
@@ -1,3 +1,134 @@
...
+  // code that is executed if "condition1" is true
+} else if (condition2) {
+  // code that is executed if "condition2" is true
+  // but "condition1" was false
+} else {
+  // code that is executed otherwise
+}
+```
+
+Theoretically you can nest as many additional conditions as you want. In practice you would use a [`switch` statement](/tracks/javascript/concepts/conditionals-switch) instead in these cases.
```
SleeplessByte, line 38:
```suggestion
Theoretically you can nest as many additional conditions as you want.
In practice you would use a [`switch` statement](/tracks/javascript/concepts/conditionals-switch) instead in these cases.
```

concepts/conditionals/about.md
```diff
@@ -1,3 +1,134 @@
...
+if (condition) doSomething();
+
+// or
+
+if (condition)
+  doSomething();
+```
+<!-- prettier-ignore-end -->
+
+This is sometimes used when checking for an error condition for example. In general it is not recommended because it is easy to forgot to add the brackets back in when adding a second statement that should depend on the same condition.
```
SleeplessByte, line 104:
```suggestion
This is sometimes used when checking for an error condition for example.
In general it is not recommended because it is easy to forget to add the brackets back in when adding a second statement that should depend on the same condition.
```

Small typo

concepts/conditionals/about.md
```diff
@@ -1,3 +1,134 @@
...
+
+When writing functions, it is a common pattern to omit the `else` block and use an early `return` in the `if` block instead.
```
SleeplessByte, line 106:
```suggestion
When writing functions, it is a common pattern to omit the `else` block and use an early `return` in the `if` block instead.
In many cases this reduces _nesting_ and makes the code more readable and easier to follow.
```

concepts/conditionals/about.md
```diff
@@ -1,3 +1,134 @@
...
+  if (num === 0) {
+    return 'You passed 0, please provide another number.';
+  }
+
+  return 'Thanks for passing such a nice number.';
+}
+```
+
+[concept-type-coercion]: /tracks/javascript/concepts/type-coercion
+[concept-null-undefined]: /tracks/javascript/concepts/null-undefined
```
SleeplessByte, line 132: This doesn't seem to be used?
```suggestion
```

concepts/conditionals/about.md
```diff
@@ -1,3 +1,134 @@
...
+[mdn-nan]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN
```
SleeplessByte, line 133: This doesn't seem to be used?

```suggestion
```

concepts/conditionals/links.json
```diff
@@ -1 +1,6 @@
-[]
+[
+  {
+    "url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/if...else",
+    "description": "Official documentation for the if-else statement"
```
SleeplessByte, line 4:
```suggestion
    "description": "MDN: The if-else statement"
```

config.json
```diff
@@ -131,6 +131,13 @@
         "concepts": ["type-conversion"],
         "prerequisites": ["arrays", "strings", "numbers"],
         "status": "wip"
+      },
+      {
+        "slug": "vehicle-purchase",
```
SleeplessByte, line 136: Missing a `"title"`

exercises/concept/vehicle-purchase/.docs/instructions.md
```diff
@@ -0,0 +1,49 @@
...
+needsLicense('car');
+// => true
+
+needsLicense('bike');
+// => false
+```
+
+## 2. Choose between two potential vehicles to buy
+
+You evaluate your options of available vehicles. You manage to narrow it down to two options but you need help making the final decision.
```
SleeplessByte, line 23:
```suggestion
You evaluate your options of available vehicles.
You manage to narrow it down to two options but you need help making the final decision.
```

exercises/concept/vehicle-purchase/.docs/instructions.md
```diff
@@ -0,0 +1,49 @@
+# Instructions
+
+In this exercise you are going to write some code to help you prepare to buy a vehicle.
+
+You have three tasks, one to determine if you will need to get a licence, one to help you choose between two vehicles and one to estimate the acceptable price for a used vehicle.
+
+## 1. Determine if you will need a drivers licence
+
+Some kinds of vehicles require a drivers license to operate them. Assume only the kinds `'car'` and `'truck'` require a license, everything else can be operated without a license.
```
SleeplessByte, line 9:
```suggestion
Some kinds of vehicles require a drivers license to operate them.
Assume only the kinds `'car'` and `'truck'` require a license, everything else can be operated without a license.
```

exercises/concept/vehicle-purchase/.docs/introduction.md
```diff
@@ -0,0 +1,77 @@
...
+'a' < 'above';
+// => true
+
+'a' === 'A';
+// => false
+```
+
+## Conditionals
+
+A common way to conditionally execute logic in JavaScript is the if-statement. It consists of the `if` keyword, a condition wrapped in round brackets and a code block wrapped in curly brackets. The code block will only be executed if the condition evaluates to `true`.
```
SleeplessByte, line 46:
```suggestion
A common way to conditionally execute logic in JavaScript is the if-statement.
It consists of the `if` keyword, a condition wrapped in round brackets and a code block wrapped in curly brackets.
The code block will only be executed if the condition evaluates to `true`.
```

concepts/comparison/about.md
kotp, line 46:
```suggestion
You need to be careful when you compare two variables that appear to contain numeric values but are of type string. Due to the dictionary order the result will not be the same as comparing values of type number.
```

kotp (commented):
Some little suggestion(s).

concepts/comparison/about.md
ErikSchierboom: Indeed, excellent content! Once we support the new info blocks (see https://github.com/exercism/docs/pull/84), this might be a perfect candidate.

config.json
ErikSchierboom: More precisely: missing a `"name"` (see https://github.com/exercism/docs/blob/main/anatomy/tracks/config-json.md#concept-exercises) :)

ErikSchierboom (approved):

concepts/comparison/about.md
```diff
@@ -1,44 +1,87 @@
...
 // => true
-// Remember, all numbers are floating-points, so this is different syntax for
+// All numbers are floating-points, so this is different syntax for
 // the exact same value.
+```
 
-1 === 1n;
+## Comparing Strings
+
+In JavaScript the comparison operators above can also be used to compare strings. In that case a dictionary (lexicographical) order is applied. You can find a list of the exact order of all the characters [here][utf-16-list].
```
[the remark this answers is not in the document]
iHiD: I do! :)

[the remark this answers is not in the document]
junedev: sorry, forgot about this :see_no_evil:

concepts/comparison/about.md
```diff
@@ -1,44 +1,87 @@
...
 
-| A      | B         | `==`                  |
-| ------ | --------- | --------------------- |
-| Number | Undefined | `false`               |
-| Number | Null      | `false`               |
-| Number | Number    | `A === B`             |
-| Number | String    | `A === ToNumber(B)`   |
-| Number | Boolean   | `A === ToNumber(B)`   |
-| Number | Object    | `A == ToPrimitive(B)` |
+What should you do instead? You will learn later how to do [explicit type conversion][concept-type-conversion]. With that you can then ensure values have the correct type before performing the comparison. Then your code will be easier to understand and less error prone.
```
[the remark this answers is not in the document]
junedev: Good point!

concepts/comparison/about.md
```diff
@@ -1,44 +1,87 @@
...
+
+You need to be careful when you compare two variables that contain numeric values but are of type string. Due to the dictionary order the result will not be the same as comparing values of type number.
+
+```javascript
+10 < 2;
+// => false
+
+'10' < '2';
+// => true (because "1" comes before "2")
+```
```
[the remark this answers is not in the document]
junedev: done

config.json
junedev: fixed

concepts/comparison/about.md
junedev: fixed

exercises/concept/vehicle-purchase/vehicle-purchase.spec.js
```diff
@@ -0,0 +1,57 @@
...
+  describe('needsLicense', () => {
+    const testCases = [
+      ['car', true],
+      ['truck', true],
+      ['bike', false],
+      ['stroller', false],
+      ['e-scooter', false],
+    ];
+
+    testCases.forEach(([kind, expected]) => {
```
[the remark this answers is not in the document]
junedev: fixed

junedev:
@SleeplessByte
I added the note regarding the Analyzer feedback to the design.md file.

> Feel free to apply which you like, and then self squash merge.

I fixed all the things mentioned in the review but I don't have rights to merge myself.

SleeplessByte:
🎆🔥 🚒 🧯

PR #1051, opened 2021-03-20
Add command to cd into exercise directory

zorbash:
The first step of the requirements assumes the user is familiar with the command-line and the exercism cli. 
Let's guide them so that they won't get stuck early on.

```
cd $(exercism workspace)/javascript/*
```

The above will `cd` into the directory of the first exercise.

### DIFF
--- a/config/exercise-readme-insert.md
+++ b/config/exercise-readme-insert.md
@@ -9,7 +9,11 @@ dependencies:
 
 ## Requirements
 
-Please `cd` into exercise directory before running all below commands.
+`cd` into exercise directory:
+
+```bash
+$ cd $(exercism workspace)/javascript/*
+```
 
 Install assignment dependencies:
 


SleeplessByte (changes requested):
I like the idea.

- `$(exercism workspace)` won't work cross-env
- `*` will pick the first, which is also probably not what you want?

zorbash:
> * `$(exercism workspace)` won't work cross-env

By cross-env you mean that it won't run on Powershell / Windows?

> * `*` will pick the first, which is also probably not what you want?

It's fair to assume that someone reading the requirements has only downloaded a single exercise and it's probably their first time trying exercism. Is there a more sensible directory to cd them into?

SleeplessByte:
> By cross-env you mean that it won't run on Powershell / Windows?

Yes, or any shell that's not bourne-based / lacks sub-shells with that syntax.

> It's fair to assume that someone reading the requirements has only downloaded a single exercise and it's probably their first time trying exercism. Is there a more sensible directory to cd them into?

But then I don't think it should be inside _every_ exercise. That said, I don't know if it will work, but you can try to replace `*` with `{{ .Spec.Slug }}` and run the readme generate command (see: https://github.com/exercism/javascript/blob/main/CONTRIBUTING.md#improving-the-readmemd). I don't know if it will work; I think it will not replace the interpolation inside an insertion file, but might be worth the try.

tejasbubane:
Related https://github.com/exercism/cli/issues/944

zorbash:
Thanks @tejasbubane for pointing out there's an existing discussion about this.

SleeplessByte:
Given that we'll redirect most new users to the website (in v3), and do a _lot_ of handholding if they download the CLI (in v3) and the open discussion to solve this elsewhere first, I'm going to close this for now.

Issue #990, opened 2021-02-22
[V3] Implement new Concept Exercise: For Loops


junedev:
## Getting started
**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md)
- [What are concepts and how they are structured?](https://github.com/exercism/docs/blob/main/anatomy/tracks/concepts.md)
- [What are concept exercises and how they are structured?](https://github.com/exercism/docs/blob/main/anatomy/tracks/concept-exercises.md)

Please also watch the following video:
- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

See the documentation _above_ (general documentation), as well as [How to implement a Concept Exercise in JavaScript](https://github.com/exercism/javascript/blob/main/reference/implementing-a-concept-exercise.md).

## Goal
The goal of this exercise is to introduce for loops and the increment and decrement operator.

## Concepts
The following concept pages need to be created. You can combine the two introduction.md files from the concepts and use that as introduction.md for the concept exercise. No need to create different content at this point.
* `for-loops`
* `increment-decrement`

## Learning Objectives
In the concepts the student should learn about the following statements and operators and then practice them in the concept exercise.

* What does a for loop do/ when is it used
* Syntax `for(...){...}`
* What is the loop counter, initialization, condition and iteration
* How to iterate over an array with a simple for loop (more advanced ways of iterating over arrays will be introduced later)
* What is the increment/decrement operator `i++`/`i--`

## Out of Scope
* `while`

Also see "About" section below for other things that should not be part of the concept introduction and the exercise.

## About.md of the Concept
Here some ideas of what could be mentioned.

### For Loops
* Labels
* `break`/`continue`
* explain that initialization, condition and iteration [are all optional](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for#optional_for_expressions)

### Increment/Decrement Operators
* Talk about the return value of `i++`/`i--`
* `++i`/`--i` (explain the difference to `i++`/`i--`)

## Prerequisites
* `arrays`
* `comparison`
* `conditionals`

[...]

PR #1059, opened 2021-03-28
Add Concept Exercise: For Loops

junedev:
Addresses https://github.com/exercism/javascript/issues/990 

### Notes
* Since we do not have a concept about asynchronous code yet and it is hard to predict how we will cover this topic exactly, I am currently linking to the MDN docs instead although there are not a super good introduction.
* I used some prettier exceptions in the about.md for for loops because otherwise the main point of the examples was very hard to see, hope that is ok.


concepts/for-loops/about.md
```diff
@@ -0,0 +1,124 @@
...
+const list = ['a', 'b', 'c'];
+for (let i = 0; i < list.length; i++) {
+  // code that should be executed for each item in the array
+}
+
+// i is not available here
+```
+
+Variables that are defined in the header with `let` are scoped to the for statement.
+In the example above you would see `ReferenceError: i is not defined` if you would try to access the counter outside of the for loop.
```
neenjaw, line 29: Since this is more related to variable scope with `let`, is there a concept for that which could be referenced as well?

neenjaw (approved):
Looks great @junedev, a nice straight-forward exercise! I left one question about variable scope, I wasn't sure if that had been implemented yet or if it was on the priority list.

LGTM

neenjaw:
@SleeplessByte can you apply the`reputation/contributed_code/major` label to this? I think the labels are locked on this repo

concepts/for-loops/about.md
```diff
@@ -0,0 +1,124 @@
...
+
+All three parts of the header are optional.
+This is rarely utilized in practice.
+You can read more about this in the [MDN documentation][mdn-optional-header-parts]
```
SleeplessByte, line 33:
```suggestion
You can read more about this in the [MDN documentation][mdn-optional-header-parts].
```

concepts/for-loops/about.md
```diff
@@ -0,0 +1,124 @@
...
+  for (let j = 0; j < coords[i].length; j++) {
+    // do something with coords[i][j]
+  }
+}
+```
+
+## Break, Continue and Labels
+
+Inside a loop body you can use the `break` keyword to stop the execution of the loop entirely.
+In contrast to this, the keyword `continue` only stops the execution of the current iteration and continues with the next one.
```
SleeplessByte, line 57:
```suggestion
In contrast, the keyword `continue` only stops the execution of the current iteration and continues with the next one.
```

`this` is so overloaded in JavaScript :P

SleeplessByte (approved):
Looking great!

concepts/for-loops/about.md
junedev: You are right, this is not really related to the for loop and should be part of the concept were we deep dive into variables. I removed it here.

junedev:
`/sync`

junedev:
I can't get the sync check to pass. I tried `npx babel-node scripts/sync` locally, no change. I tried `/sync` here, did not do anything. I tried an empty commit to rerun the check, still not working. What's wrong?
cc @SleeplessByte

junedev:
Besides the failing CI this is ready to be merged, I addressed the review comments.

SleeplessByte:
> I can't get the sync check to pass. I tried `npx babel-node scripts/sync` locally, no change. I tried `/sync` here, did not do anything. I tried an empty commit to rerun the check, still not working. What's wrong?
> cc @SleeplessByte

The `pull_request` pr check (I think) runs on a _merged detached head_ (which is great). So what has happened here is that the base branch `main` was actually updated and after merge, this exercise would not be in sync. 

1. `git checkout main`
2. `git fetch upstream`
3. `git rebase upstream/main`
4. `git checkout <your branch>`
5. `git rebase main`
6. `git push --force`

Now syncing does work.

(I've done this for you).

PR #1125, opened 2021-05-05
Remove node_modules and add it to gitignore

tejasbubane:
`node_modules` got merged in https://github.com/exercism/javascript/pull/1118


SleeplessByte:
Ugh, I totally missed that. also, why???

PR #1130, opened 2021-05-09
Add instructions to not use `Date` for `clock` exercise

tejasbubane:
Came across a solution where student used `Date` class to implement `clock` exercise.

### DIFF
--- a/exercises/practice/clock/.docs/instructions.append.md
+++ b/exercises/practice/clock/.docs/instructions.append.md
@@ -1,0 +1,3 @@
+# Instructions append
+
+Using the built-in [Date class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) and its methods is not allowed.


exercises/practice/clock/.docs/instructions.append.md
```diff
@@ -0,0 +1,4 @@
+# Instructions append
+
+Using the built-in [Date class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) and its methods
+is not allowed.
```
SleeplessByte, line 4:
```suggestion
Using the built-in [Date class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) and its methods is not allowed.
```

SleeplessByte: I think the new format now is "one sentence per line".

SleeplessByte (approved):

PR #1157, opened 2021-05-29
Add prerequisites

TomPradat:


config.json
```diff
@@ -606,7 +606,7 @@
         "name": "Complex Numbers",
         "uuid": "ea9a9a3e-ae6a-470d-8bb4-2afead507f24",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["classes", "numbers"],
```
TomPradat, line 609: Should we introduce a "math" concept ? That would teach the basics of the "Math" object

SleeplessByte: If you feel that `math` is a concept in JavaScript. If it's just knowing that `Math` exist then maybe not.

@junedev what do you think?

junedev: I would vote in favor of covering the Math object with some concept e.g. because there are pitfalls like incorrect results for some of the functions (not 100% sure these still exist though).

That said, the concept name could end up something like "built-in-objects" that includes other things. Not sure yet. So I think putting in "math" for now and replacing in later if needed is fine.

SleeplessByte: Let's do that. @TomPradat can you make the change?

SleeplessByte (approved):

PR #1159, opened 2021-05-30
Add prerequisite for difference-of-squares exercise

TomPradat:


tejasbubane:
@TomPradat Can you clarify this a little bit? Because the exercise [can be solved without arrays](https://github.com/exercism/javascript/blob/3be193160a49a70c2ac2e69aa7c482b21d7ba09e/exercises/practice/difference-of-squares/.meta/proof.ci.js) so I would not consider `array-transformations` as a "prerequisite".

TomPradat:
That's true @tejasbubane , I chose a set of concepts that can be used to solve the exercise.

According to https://github.com/exercism/javascript/issues/960 if alternative concepts also allow the student to solve the exercise, we can specify the concepts of either approaches.

That said It seems that currently more exercises teach `array-transformations` than `for-loops`. So I don't mind changing for `for-loops`. What do you think ?

SleeplessByte:
Loops sounds good. Even when we use transformations, we're using loops, but not the other way around, I guess.

SleeplessByte (approved):

SleeplessByte (approved):

Issue #993, opened 2021-02-23
 [V3] Implement new Concept Exercise: Objects


junedev:
## Getting started
**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md)
- [What are concepts and how they are structured?](https://github.com/exercism/docs/blob/main/anatomy/tracks/concepts.md)
- [What are concept exercises and how they are structured?](https://github.com/exercism/docs/blob/main/anatomy/tracks/concept-exercises.md)

Please also watch the following video:
- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

See the documentation _above_ (general documentation), as well as [How to implement a Concept Exercise in JavaScript](https://github.com/exercism/javascript/blob/main/reference/implementing-a-concept-exercise.md).

## Goal
The goal of this exercise is to give a simple introduction to objects as maps/dictionaries.

## Concepts
The following concept page needs to be created. You can use the same content for the introduction.md of the concept and the concept exercise. No need to create different content at this point.
* `objects`

## Learning Objectives
In the concepts/introduction the student should learn about the following things and then practice them in the concept exercise.

* What are objects - for now describe them as maps/dictionaries that hold key-value pairs, more advanced things will be taught later
* Objects are the core building blocks in JavaScript
* How to create an object literal with `{}` (either empty or with some initial values)
* What is allowed as keys
* What is allowed as values (e.g., numbers, strings etc or other objects, arrays or even functions)
* How to add a key-value pair to an object, how to change the value of an existing key
* How to retrieve the value (show `obj["key"]` and `obj.key` notation)
* How to remove an entry
* How to check a key exists in the object with `hasOwnProperty`
* How to iterate through the keys with `for ... in`

## Out of Scope
* Prototypes and classes
* all the built-in objects
* `this`
* Object destructering
* null and undefined (will be introduced in the next concept exercise, including when they show up in the context of objects, it is a bit hard to tiptoe around this in this concept/exercise but the student can't learn everything at once)
* `new Object()` / `Object.create`

## About.md of the Concept
Here some ideas of what could be mentioned.

[...]

Issue #951, opened 2021-01-28
[JavaScript] Write stub issues for new concept exercises


SleeplessByte:
Here is a list of concepts modified from the C# one. I marked a few that I know exist already, but we should add design document to ALL our current exercises too. 

**How can you help?**: 

1. For many of these we don't have exercises or exercise ideas yet, for example `chars` + `codepoints` doesn't exist yet. These need a `implement new exercise` issue with all the relevant information (see current open issues). 
2. Some of these concepts do have exercise ideas (and/or are being worked on), for example `errors`. **Ignore those**. 
3. Some of these already have exercises, but are not mentioned in the exercise. @ErikSchierboom can you explain how you assign the concepts and how it gets turned into that automated reference readme document? Once Erik does, we need to update our files (design doc, config.json) so we can do this too.
4. We're missing quite a few things from [this document](https://github.com/exercism/v3/tree/master/languages/javascript/reference). Things missing should be added here and go through the above three "steps".

## Concepts

- [x] arrays (`Array.prototype`)
- [ ] assignment
- [x] basics
- [ ] bit-manipulation
- [x] booleans (`boolean`)
- [x] casting (JavaScript term: type-conversion)
- [ ] chars (JavaScript term: chars and codepoints)
- [ ] classes (JavaScript term alternative: prototypes)
- [x] closures
- [ ] codepoints
- [ ] conditionals-ternary
- [ ] constants (`const` for bindings, `Object.freeze` for immutability)
- [ ] constructors
- [ ] datetimes (`Date.prototype`)
- [ ] default-value (in: parameter/argument, destructuring)
- [ ] dictionaries (`Map.prototype`, `WeakMap.prototype`, `object`)
- [ ] encapsulation (JavaScript term: closures)
- [ ] equality
- [ ] exception-filtering (JavaScript term: type-checking)
- [x] exceptions
- [ ] fields (JavaScript term: properties)
- [ ] floating-point-numbers (JavaScript term: numbers)
- [ ] for-loops
- [ ] foreach-loops 
- [ ] functions
- [ ] if-statements
- [ ] implicit-casting
- [ ] imports (JavaScript term: module imports)
- [ ] inheritance (JavaScript term: prototype inheritance)
- [ ] interpolation
- [ ] math-operators
- [ ] method-overloading
- [ ] mutation
- [ ] named-arguments
- [ ] modules
- [ ] module-exports
- [ ] module-imports
- [ ] module-dynamic-imports
- [ ] null-coalescing
[...]

PR #1160, opened 2021-05-30
Add concept exercise objects

junedev:
Closes https://github.com/exercism/javascript/issues/993

This one was harder than the others ones. I feel it could still need some more work (left some notes in design.md) but I want to get it out of the door. Hope it is ok for now.


exercises/concept/high-score-board/.meta/exemplar.js
```diff
@@ -0,0 +1,76 @@
+// @ts-check
+
+/**
+ * Creates a new score board with an initial entry.
+ *
+ * @returns {Object.<string, number>} new score board
```
junedev, line 6: @SleeplessByte I am not up to date with this JSDoc ts-check business. :point_up_2:  This is what I found in the JSDoc documentation how to describe the object. Is that ok or should there be something else (more typescripty) here?

SleeplessByte: @junedev you want to represent a shape that has keys of string and values of number?

```typescript
@returns {Record<string, number>}
```

which is essentially the same as:

```typescript
@returns {{ [key: string]: number }}
```

junedev: > you want to represent a shape that has keys of string and values of number?

Yes. Inserted your suggestion. Thanks!

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
+# About
+
+## Explanation
+
+Besides primitive data types like `number` and `string`, there is another important data type in JavaScript called `object`.
+Objects are collections of key-value pairs.
+In other languages they are referred to as maps or dictionaries and often times values need to have the same data type.
```
SaschaMann, line 7:
```suggestion
In other languages they are referred to as maps or dictionaries and often times all values need to have the same data type.
```

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+In JavaScript, only the type of the key is restricted, it has to be a string.
+The values inside one object can have different types.
+They can be primitive types like numbers but also arrays, other objects or even functions.
+
+## Creating an Object
+
+You create an object (literal) with curly brackets.
```
SaschaMann, line 14: Does the student know what `(literal)` refers to at this point in the track?

If not, perhaps adding a brief footnote or explanation would prevent some web searching.

concepts/objects/about.md
SaschaMann: My initial guess would have been that they're more similar to structs or data classes in other languages, which can have varying value types. Perhaps you could briefly mention how those two relate?

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+  'key/with/slashes': '...',
+  'key-with-dashes': '...',
+  'key with spaces': '...',
+  '#&()[]{}èä樹keyWithSpecialChars': '...',
+};
+```
+
+Often times you want to combine existing variables into an object.
+This would often lead to key-value pairs like `name: name`.
+JavaScript has a short-hand notation that allows to just write `name` instead.
```
SaschaMann, line 55:
```suggestion
JavaScript has a short-hand notation that allows just writing `name` instead.
```

`allow` is followed by a gerund, I think.

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+    city: 'Batticaloa',
+  },
+  fruits: ['melon', 'papaya'],
+  addNumbers: function (a, b) {
+    return a + b;
+  },
+};
+```
+
+The trailing comma after the last entry is optional in JavaScript.
```
SaschaMann, line 36: It may be a good idea to mention why one would want to add a trailing comma (better diffs etc).

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+const key = 'greeting';
+obj[key];
+// => hello world
+```
+
+Using the dot notation as a short-hand has the same restriction as omitting the quotation marks.
+It only works if the key follows the identifier naming rules.
+
+If you try to retrieve a key that does not exist in the object, JavaScript returns `undefined`.
+See the [Null and Undefined Concept][concept-null-undefined] for more details on this.
```
SaschaMann, line 114:
```suggestion
See [Null and Undefined][concept-null-undefined] for more details on this.
```

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+// name Ali
+// age 65
+```
+
+It might seem like `for...in` always visits the keys in the order in which they appear or were added to the object (insertion order).
+Nevertheless you should not rely on this because there are special cases where the order will differ from the insertion order.
+Always write your code as if the order would be arbitrary.
+If you need a reliable order, use a [`Map`][jsinfo-map] or a nested [array][concept-arrays] instead of an object.
+
+The for-in loop has a pitfall in case the object has [inherited keys][concept-inheritance].
```
SaschaMann, line 207:
```suggestion
The `for…in` loop has a pitfall in case the object has [inherited keys][concept-inheritance].
```

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+Those will also be visited by the for-in loop.
```
SaschaMann, line 208:
```suggestion
Those will also be visited by the `for…in` loop.
```

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+obj.hasOwnProperty('age');
+// => false
+```
+
+`hasOwnProperty` returns `false` for [inherited keys][concept-inheritance]. This is usually the desired behavior.
+If you want include inherited keys in the existence check you can use the [`in` operator][mdn-in-operator] instead.
+
+## Looping Through an Object
+
+There is a special `for...in` loop to iterate over all keys of an object.
```
SaschaMann, line 187: Are objects in JS iterables? Would `for…of` work too?

concepts/objects/introduction.md
```diff
@@ -0,0 +1,129 @@
+# Introduction
```
SaschaMann, line 1: Imo the headers from the about should be in this doc, too. Makes it easier to navigate since it's quite long.

concepts/objects/about.md
SaschaMann: Not sure which one is the common name, but I think consistency might be nicer. Same applies to the introduction.

exercises/concept/high-score-board/.docs/hints.md
```diff
@@ -0,0 +1,38 @@
+# Hints
+
+## 1. Create a new high score board
+
+- You create a new object with curly brackets.
```
SaschaMann, line 5:
```suggestion
- Create a new object with curly brackets.
```

consistency with the other hints

exercises/concept/high-score-board/.docs/instructions.md
```diff
@@ -0,0 +1,112 @@
...
+The function returns the same score board object that was passed in after adding the new player.
+
+```javascript
+addPlayer({ 'Dave Thomas': 0 }, 'José Valim', 486373);
+// => {'Dave Thomas': 0, 'José Valim': 486373}
+```
+
+## 3. Remove players from a score board
+
+If a player did not visit the arcade hall for one year, they are removed from the high score board.
```
SaschaMann, line 37: Very minor/subjective:

This somewhat sounds like the task would be about implementing logic that detects old scores and removes them. It's not misleading since the rest of the task makes it clear what it's about but perhaps you could avoid this by changing it to something like "If a player was found to be cheating, they are manually removed from the high score board."

exercises/concept/high-score-board/.meta/config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "blurb": "FIXME",
+  "authors": ["junedev"],
+  "contributors": [],
+  "files": {
+    "solution": ["high-score-board.js"],
+    "test": ["high-score-board.spec.js"],
+    "exemplar": [".meta/exemplar.js"]
+  },
+  "forked_from": ["julia/high-score-board"]
```
SaschaMann, line 10:
```suggestion
  "forked_from": ["julia/high-score"]
```

exercises/concept/high-score-board/.meta/design.md
```diff
@@ -0,0 +1,94 @@
+# Design
+
+## Learning objectives
+
+- What are objects - for now describe them as maps/dictionaries that hold key-value pairs, more advanced things will be taught later
+- Objects are the core building blocks in JavaScript
```
SaschaMann, line 6: To me their importance didn't come across from the exercise docs. Perhaps ask others for input on this, though, before adding more info that might just end up being redundant.

exercises/concept/high-score-board/.meta/design.md
```diff
@@ -0,0 +1,94 @@
...
+- How to create an object literal with `{}` (either empty or with some initial values)
+- What is allowed as keys
+- What is allowed as values (e.g., numbers, strings etc or other objects, arrays or even functions)
```
SaschaMann, line 9:
```suggestion
- What keys are allowed
- What values are allowed (e.g., numbers, strings etc or other objects, arrays or even functions)
```

exercises/concept/high-score-board/.meta/design.md
```diff
@@ -0,0 +1,94 @@
...
+- How to add a key-value pair to an object, how to change the value of an existing key
+- How to retrieve the value (show `obj["key"]` and `obj.key` notation)
+- How to retrieve nested values
+- How to remove an entry
+- How to check a key exists in the object with `hasOwnProperty`
+- How to iterate through the keys with `for ... in`
```
SaschaMann, line 15:
```suggestion
- How to iterate through the keys with `for...in`
```

config.json
```diff
@@ -168,6 +168,14 @@
         "prerequisites": ["arrays", "callbacks", "errors", "recursion"],
         "status": "beta"
       },
+      {
+        "slug": "high-score-board",
+        "name": "High Score Board",
+        "uuid": "431d1f13-d280-4808-bbc6-a72e628c15c2",
+        "concepts": ["objects"],
+        "prerequisites": ["for-loops"],
```
SaschaMann, line 176: `strings` and `numbers` might be pre-reqs too but I don't know enough about the JS progression. Just pointing it out for someone else's review.

exercises/concept/high-score-board/.meta/design.md
```diff
@@ -0,0 +1,94 @@
...
+
+1. `createScoreBoard`
+
+   - `essential`: Make sure no class, map etc. was created, there should be just an object.
+   - `actionable`: If the student created an empty object first and then added the value, give feedback to include the entry in the object literal directly.
+   - `actionable`: Check that the object was returned directly, no intermediate assignment to a variable necessary.
+
+2. `addPlayer`
+
+   - ???
```
SaschaMann, line 51:
```suggestion
   - TODO/FIXME
```

Again, not sure what the conventions are in the JS repo, but `???` isn't easily findable via grepping, I'm guessing.

exercises/concept/high-score-board/.meta/design.md
```diff
@@ -0,0 +1,94 @@
...
+
+3. `removePlayer`
+
+   - `essential`: Make sure `delete` was used and not setting to undefined or null.
```
SaschaMann, line 55:
```suggestion
   - `essential`: Make sure `delete` was used and not set to undefined or null.
```

exercises/concept/high-score-board/.meta/design.md
```diff
@@ -0,0 +1,94 @@
...
+   - Same feedback as in `updateScore` applies.
+   - Using `updateScore` in the solution should be treated as equally correct as the exemplar solution.
+
+6. `normalizeScore`
+
+   - `actionable`: No intermediate variables necessary.
+
+## Notes
+
+The exercise is inspired by the [High Score Board Exercise in the Swift track][swift-high-score].
```
SaschaMann, line 75: `forked_from` points to the Julia implementation, and `source` is empty. Perhaps add Elixir to either of those?

I'm not the best person to ask about this tho, @iHiD will likely have a different stance on it.

exercises/concept/high-score-board/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Exercism
```
SaschaMann, line 3:
```suggestion
Copyright (c) 2021 Exercism
```

SaschaMann (commented):
Looks great overall, I left some minor comments (from the PoV of someone who barely knows JS). I did not look at the tests/exemplar files.

concepts/objects/about.md
iHiD: Also, on this line "often times" isn't familiar sounding English to me. Just "often" without the "times" is clearer 🙂

SaschaMann: I think it's a US phrase mostly

exercises/concept/high-score-board/LICENSE
SleeplessByte: This needs to be fixed in the ROOT, and then it propagates via a sync. Probably a separate PR :)

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+  'key with spaces': '...',
+  '#&()[]{}èä樹keyWithSpecialChars': '...',
+};
+```
+
+Often times you want to combine existing variables into an object.
+This would often lead to key-value pairs like `name: name`.
+JavaScript has a short-hand notation that allows to just write `name` instead.
+There is also a short-hand notation for defining a function in an object.
+There you can omit the colon and the `function` keyword.
```
SleeplessByte, line 57: Might not be the place to mention this, but the short-hand notation is **not** equivalent to the key-value notation. The difference is small, but there:

```javascript
const appleTree = {
  kind: 'apple',
  getParent: function() {
    return super.toString();
  }
};
// Uncaught SyntaxError: 'super' keyword unexpected here

const pearTree = {
  kind: 'pear',
  getParent() {
    return super.toString();
  }
};
// this works

// You can then declare things on its prototype
Object.setPrototypeOf(pearTree, { toString() { return "BigTree" } })
pearTree.getParent()
// => "BigTree"
```

It's a weird edgy case, but it matters as soon as you're going to type to _copy_ that method you've declared or if you're trying to write mixins.

SleeplessByte: There's another _difference_ between these:

```javascript
const obj = {
  calcSum: function calc(a, b) {
    return a + b;
  },
};

obj.calcSum.name
// => "calc"

const obj2 = {
  calcSum: function (a, b) {
    return a + b;
  },
};

obj2.calcSum.name
// => "calcSum"

const obj3 = {
  calcSum(a, b) {
    return a + b;
  },
};

obj3.calcSum.name
// => "calcSum"
```

It's pretty insignificant, but 🤷🏽

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+
+If you want to create an object literal with a key name that is stored in a variable or derived via an expression, you need to wrap the variable or expression in square brackets.
+
+```javascript
+const key = 'greeting';
+
+const obj = {
+  [key]: 'hello world',
+};
+```
```
SleeplessByte, line 89: Might be worth showing the result

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Retrieving a Value
+
+There are two ways to retrieve the value for a given key, dot notation and bracket notation.
```
SleeplessByte, line 93: We should definitely NOT mention this, but there are more than two ways.

```javascript
const obj = { greeting: 'hello world' }

const descriptor = Object.getOwnPropertyDescriptor(obj , 'greeting')
// => {value: "hello world", writable: true, enumerable: true, configurable: true}

descriptor.value
// => "hello world"
```

And naturally by way of destructuring.

There's even `Object.values`:

```javascript
const obj = { foo: 'bar', greeting: 'hello world' }

const values = Object.values(obj)
const index = Object.getOwnPropertyNames(obj).indexOf('greeting')

values[index]
// => "hello world"
```

By placing these in this comment on this PR, someone, someday might find them and be sad or happy.

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+    city: 'Batticaloa',
+  },
+};
+
+obj.address.city;
+// => 'Batticaloa'
+
+obj['address']['city'];
+// => 'Batticaloa'
+```
```
SleeplessByte, line 131: Perhaps worth mentioning that you _could_ mix and match?

```javascript
obj.address['city']
```

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+
+obj.greeting = 'Hi there!';
+obj['greeting'] = 'Welcome.';
+
+obj.newKey1 = 'new value 1';
+obj['new key 2'] = 'new value 2';
+
+const key = 'new key 3';
+obj[key] = 'new value 3';
+```
```
SleeplessByte, line 149: If you add the property using `Object.create` or like this:

```javascript
const obj = {}
Object.defineProperty(obj, 'greeting', { value: 'hello world', configurable: true })
// => {greeting: "hello world"}
```

You can actually then do:

```javascript
Object.defineProperty(obj, 'greeting', { value: 'goodbye world' })
// => {greeting: "goodbye world"}
```

`defineProperty` etc is probably something we want somewhere but not here, but I don't know where so that's why I put it here so I hope it gives you inspiration.

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+const obj = {
+  key1: 'value1',
+  key2: 'value2',
+};
+
+delete obj.key1;
+delete obj['key2'];
+```
+
+Note that although `undefined` is returned for non-existing keys, setting a key to `undefined` does not remove the entry.
```
SleeplessByte, line 165: Very good advice!

Might be worth showing `hasOwnProperty` to validate something is gone, given that you show it below?

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+
+obj.hasOwnProperty('greeting');
+// => true
+
+obj.hasOwnProperty('age');
+// => false
+```
+
+`hasOwnProperty` returns `false` for [inherited keys][concept-inheritance]. This is usually the desired behavior.
+If you want include inherited keys in the existence check you can use the [`in` operator][mdn-in-operator] instead.
```
SleeplessByte, line 183: You might want to mention `getOwnPropertyNames` which are all the properties `hasOwnProperty` is true for.

concepts/objects/about.md
SleeplessByte: `for...in` and `for...of` are not equivalent, and the concept of enumerable is not the same as iterable, in JavaScript (@junedev all this might be worth mentioning, somewhere).

JavaScript objects have _enumerable properties_. They are enumerable by default, and can be made non enumerable using `Object.create` or `Object.defineProperty` setting `{ enumerable: false }`.

They are _only_ iterable if there is `Symbol.iterator` defined (as a method):

```javascript
const iterable1 = {};

iterable1[Symbol.iterator] = function* () {
  yield 1;
  yield 2;
  yield 3;
};

console.log([...iterable1]);
// [1, 2, 3]
```

[Array's prototype has it defined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/@@iterator), but Object's prototype does not. _Iterability_ is **only** used by `for ... of`. The async variant is `Symbol.asyncIterator` and is used by **for-await-of**.

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
...
+```
+
+It might seem like `for...in` always visits the keys in the order in which they appear or were added to the object (insertion order).
+Nevertheless you should not rely on this because there are special cases where the order will differ from the insertion order.
+Always write your code as if the order would be arbitrary.
+If you need a reliable order, use a [`Map`][jsinfo-map] or a nested [array][concept-arrays] instead of an object.
+
+The for-in loop has a pitfall in case the object has [inherited keys][concept-inheritance].
+Those will also be visited by the for-in loop.
+You can avoid this by skipping over them like shown below.
```
SleeplessByte, line 209: Perhaps it should be mentioned that you almost _always_ want this.

SleeplessByte (commented):
I've only reviewed about.md, because I don't want to overload ya here, but see if it aids you in any way.

concepts/objects/about.md
junedev: Re the first issue: 
Interesting, I didn't know and it was not mentioned [on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions#description). Do you know why there is a difference or can you provide some link where I could read up on this? Tbh I would not hold off merging the PR until I got in. But I can put it into the design document as improvement idea. It might be a good fit once we have the special "sections" for about.md where you can mark something as advanced content.

Re the second issue:
True, but this is more about how function names are determined than about the shorthand notation imo. It could go into the functions concept but even there we have a lot more basic ground to cover first.

Its not worth holding off this PR for any of those. I would put both things into the design.md improvement ideas section of the respective exercise. They might be a good fit once we have the special "sections" for about.md where you can mark something as advanced content.

concepts/objects/about.md
junedev: "There are some more ways ... " probably applies to every other thing I cover in all the JavaScript concepts. :wink:

concepts/objects/about.md
junedev: Totally agree it should go in somewhere, that is why I wrote "Helper methods like `Object.assign` etc." should be covered in the about.md in the future in the design document. I rather wanted to move on to other concepts at some point. So in a next iteration on this I would see these helpers being explained in the `objects` about.md and then further down the road they could even be moved into some "objects advanced" or whatever exercise if we find more related things we maybe want to cover with real exercises and not just mention in about.md.

config.json
junedev: We discussed that we don't add the primitives types as prerequisites further down the concept tree in the JS track if they are not a core part of the current concept.

concepts/objects/about.md
junedev: @SaschaMann you are totally right with that first comment, in my head it was clear that we will cover that struct/class aspect of objects elsewhere but I forgot to also make this clear for the reader. Rewrote that section. Removed the "times" as well.

concepts/objects/about.md
junedev: Good point. Jeremy recommended to use links in these cases where we use terms that someone that already knows programming should know but in reality might not know.

concepts/objects/about.md
junedev: Provided a link where this is spelled out. I think that is sufficient.

exercises/concept/high-score-board/.meta/design.md
junedev: true, but I had in idea what to write there now so this became obsolete

exercises/concept/high-score-board/.meta/design.md
junedev: I just saw in the "stories" docs that Elixir is actually the reference implementation. Used that one as forked_from and in the notes now.

concepts/objects/about.md
```diff
@@ -0,0 +1,251 @@
+# About
+
+## Explanation
+
+Besides primitive data types like `number` and `string`, there is another important data type in JavaScript called `object`.
+Objects are collections of key-value pairs.
+In other languages they are referred to as maps or dictionaries and often times values need to have the same data type.
+In JavaScript, only the type of the key is restricted, it has to be a string.
+The values inside one object can have different types.
+They can be primitive types like numbers but also arrays, other objects or even functions.
```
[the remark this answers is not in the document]
junedev: I meant is as "but also entities like arrays" not as "but also types like arrays". I think it is ok as is. The "object basics" concept is not the right point for the "well actually functions are objects" point.

concepts/objects/about.md
junedev: I think just showing it below is enough here.

concepts/objects/about.md
junedev: Same category as "defineProperty", I added it the improvement ideas list.

concepts/objects/about.md
junedev: @SleeplessByte I would not go into the details of this in this concept. Left a note (and the link to this comment) on our concept tracking issue. https://github.com/exercism/javascript/issues/951#issuecomment-864452573

concepts/objects/about.md
junedev: Good point, I changed the wording accordingly.

exercises/concept/high-score-board/.docs/instructions.md
junedev: I realized the same thing at some point but forgot to fix it then. Did it now.

exercises/concept/high-score-board/.meta/config.json
junedev: see https://github.com/exercism/javascript/pull/1160#discussion_r654829051

exercises/concept/high-score-board/.meta/design.md
junedev: I removed this line. I was hinting at the common saying "everything is an object" in JavaScript but I found out that this is partly a misconception when I was looking into it in more detail. :see_no_evil:

concepts/objects/introduction.md
junedev: You are totally right. Fixed.

junedev:
> I've only reviewed about.md, because I don't want to overload ya here, but see if it aids you in any way.
@SleeplessByte No need to spare me. :wink: I addressed the existing comments by you and Sascha. Do you want to also review the rest?

concepts/objects/about.md
```diff
@@ -0,0 +1,266 @@
...
+Objects are collections of key-value pairs.
+As such, they can be used as what is referred to as maps or dictionaries in other languages.
+
+In other languages, all values in a map often need to have the same data type.
+In JavaScript, only the type of the key is restricted: it has to be a string.
+The values inside one object can have different types.
+They can be primitive types like numbers but also arrays, other objects or even functions.
+This makes objects very versatile so that they are also key entities for [object-oriented programming][oop] (OOP) in JavaScript.
+
+In the following we will focus on objects as collections or maps. Other use cases of objects will be covered in other concepts, see e.g., [Classes][concept-classes].
```
SaschaMann, line 15:
```suggestion
In the following we will focus on objects as collections or maps. Other use cases of objects will be covered in other concepts, see e.g. [Classes][concept-classes].
```

concepts/objects/about.md
SaschaMann: Depending on what comes first in the track, this may be worth mentioning, i.e. if `for..of` is explained in a previous exercise. That way, the student knows to watch out that `for..of` won't work here.

Thanks for the explanation :)

concepts/objects/about.md
```diff
@@ -0,0 +1,266 @@
+# About
+
+## Explanation
+
+Besides primitive data types like `number` and `string`, there is another important data type in JavaScript called `object`.
+Objects are collections of key-value pairs.
+As such, they can be used as what is referred to as maps or dictionaries in other languages.
```
SaschaMann, line 7:
```suggestion
As such, they can be used as what is often referred to as maps or dictionaries in other languages.
```

structs in other languages are also sometimes used purely as a collection. It's rare but imo making this statement slightly weaker clears up that it doesn't _always_ translate to a map or dictionary even if in many languages it does.

The difference to classes with methods in the context of this exercise is clearer to me now, though, so I think it'd be fine as it is too.

SaschaMann (approved):

SaschaMann:
(turns out I still have write access to this repo, so this is mergeable now, but it's probably best to wait for DJ's proper approval anyway. I meant to approve in the sense that my comments have all been addressed.)

concepts/objects/about.md
junedev: @SaschaMann Are you sure about this? I read [here](https://exercism.lol/docs/building/markdown/style-guide) that we use American English and from what read there you always put in the comma. 🤔

concepts/objects/about.md
junedev: fixed in all three places

concepts/objects/about.md
junedev: `for...of` comes later in our current unlock flow but we treat about.md as independent from that flow. So your point is totally valid. I added a small part on this. I used a link to the MDN docs for now. Once we have our own concept on enumerable and iterable we can link that instead.

exercises/concept/high-score-board/.meta/exemplar.js
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Normalizes a score with the provided normalization function.
+ *
+ * @param {Object} params
+ * @param {number} params.score
+ * @param {function} params.normalizeFn
+ * @returns {number} normalized score
+ */
+export function normalizeScore(params) {
+  return params.normalizeFn(params.score);
```
SleeplessByte, line 75:
```suggestion
  return params.normalizeFunction(params.score);
```

exercises/concept/high-score-board/.meta/exemplar.js
```diff
@@ -0,0 +1,76 @@
...
+
+  return scoreBoard;
+}
+
+/**
+ * Normalizes a score with the provided normalization function.
+ *
+ * @param {Object} params
+ * @param {number} params.score
+ * @param {function} params.normalizeFn
```
SleeplessByte, line 71:
```suggestion
 * @param {function} params.normalizeFunction
```

exercises/concept/high-score-board/high-score-board.js
```diff
@@ -0,0 +1,67 @@
...
+export function applyMondayBonus(scoreBoard) {
+  throw new Error('Please implement the applyMondayBonus function');
+}
+
+/**
+ * Normalizes a score with the provided normalization function.
+ *
+ * @param {Object} params
+ * @param {number} params.score
+ * @param {function} params.normalizeFn
```
SleeplessByte, line 62:
```suggestion
 * @param {function} params.normalizeFunction
```

SleeplessByte (approved):
Looks good. Few nitpicks.

concepts/objects/about.md
SaschaMann: I have never seen a comma after `e.g.`, hence my suggestion, but apparently you're right and American English puts a comma there indeed. The doc you linked to gets that wrong too funnily enough.

The more you know... :D

exercises/concept/high-score-board/.docs/hints.md
```diff
@@ -0,0 +1,38 @@
...
+## 2. Add players to a score board
+
+- Use bracket notation to add a key name that is stored in a variable (the argument).
+- Use the assignment operator `=` to set a value for the new key.
+- Return the score board parameter after you added the player.
+
+## 3. Remove players from a score board
+
+- Use the [delete operator][mdn-delete].
+- Reference the key like you have done in the task before (bracket notation).
```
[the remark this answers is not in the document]
junedev: I'm sorry but I don't understand the point about not using "you" in the hints. This is the "official" example from https://exercism.lol/docs/building/tracks/concept-exercises. I used that same style.
 ```markdown
 ## General

- You need to define a [constant][constant] which should contain the [integer][integers] value specified in the recipe.

## 1. Calculate the remaining oven time in minutes

- You need to define a [method][methods] with a single parameter for the actual time so far.
```
Are the docs wrong here?

File: exercises/concept/high-score-board/.docs/instructions.md [hunk 37]
```diff
@@ -0,0 +1,112 @@
...
+In this exercise, you are implementing a way to keep track of the high scores for the most popular game in your local arcade hall.
+
+You have 6 functions to implement, mostly related to manipulating an object that holds the high scores.
+
+## 1. Create a new high score board
+
+Create a function `createScoreBoard` that creates an object that serves as high score board.
+The keys of this object will be the names of the players, the values will be their scores.
+For testing purposes you want to directly include one initial entry in the object.
+There should player `The Best Ever` with a score of `1000000` already on the board.
```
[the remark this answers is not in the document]
junedev: rephrased it completely

exercises/concept/high-score-board/.docs/instructions.md
```diff
@@ -0,0 +1,112 @@
...
+
+## 6. Normalize a high score
+
+Different arcade halls award different score points.
+To celebrate the best arcade player in town, a player's score needs to be normalized so scores from different arcade halls become comparable.
+
+Write a function `normalizeScore`.
+To practice your object skills, instead of two parameters this function should accept one object as parameter.
+That object contains a key `score` and a player's score (a number) as a value.
+There is also a second key `normalizeFn` that has a function as value.
```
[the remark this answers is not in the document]
junedev: :+1: Fixed.

exercises/concept/high-score-board/.meta/exemplar.js
```diff
@@ -0,0 +1,76 @@
...
+    scoreBoard[player] += 100;
+  }
+
+  return scoreBoard;
+}
+
+/**
+ * Normalizes a score with the provided normalization function.
+ *
+ * @param {Object} params
```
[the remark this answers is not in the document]
junedev: done

junedev:
@SleeplessByte I fixed all the things besides the "you" issue, see [comment](https://github.com/exercism/javascript/pull/1160/files#r656490705).

exercises/concept/high-score-board/.docs/hints.md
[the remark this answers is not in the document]
SleeplessByte: The only reason I remove you when refering to a previous exercise or task is because they might not have done that themselves.

It's not a hard requirement, but a nice thing to do.

"You should know how to do this because you did it previously" becomes "it can be done like it has been done previously".

If the student asked for help on the previous implementation, they might go back to the person they've asked help from, without another negative pointer.

Not a blocker, feel free to ignore.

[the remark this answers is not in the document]
junedev: Ah ok, I misunderstood your comments/suggestions as "remove all the 'you'".

Nevertheless I would like to choose "ignore" here because of the "prefer active over passive voice" rule that was stated somewhere (can't find the reference anymore, wasn't moved to /docs on v3 website yet). That said, I still don't have merge rights so I need to you click the button. :see_no_evil:

PR #1163, opened 2021-06-01
Add prerequisites for kindergarten-garden

TomPradat:
I'd suggest adding a `spread-operator` concept, what do you think ?


SleeplessByte:
We have an exercise (concept) that teaches spreading, so that concept can be used.

SleeplessByte (approved):

Issue #1196, opened 2021-06-28
Check EXERCISM_PRETTIER_VERSION matches prettier in package.json


SleeplessByte:
`prettier` is configured in our workflows as `EXERCISM_PRETTIER_VERSION` (see #1195 for an example). It's version in our dependency list in `package.json` (currently `2.3.2`) must match the workflow setting exactly. Why? `prettier` will be installed if a maintainer is working on the track, and the workflow will check it against a version. If these don't match, a cycle of "updates" and "workflow failures" will occur.

In other words, given a `package.json` version of `2.3.2`, the ENV variable must be `2.3.2`.

Your task, should you accept it, is to add a check to the "precheck" scripts that run automatically on PR requests is to check if the package.json version matches the ENV variable or that the ENV variable is _set_ / _replaced_ with the `package.json` version.

laulujan:
Hello!! Is it ok if I give it a try??

SleeplessByte:
@laulujan go for it! Let me know / ping me if you need help.

laulujan:
Hi @SleeplessByte I opened a PR #1204 on this issue, could you tell me if I am on a good track or how should I approach this?  I am also wondering how to test this, Thanks in advance!!

PR #1204, opened 2021-07-06
Add check if EXERCISM_PRETTIER_VERSION matches prettier in package.json #1196

laulujan:
Issue #1196

Include check in the PR pre check to validate if ENV EXERCISM_PRETTIER_VERSION matches prettier in package.json, if not, set version in package.json to ENV.

### DIFF
--- a/.github/workflows/action-format.yml
+++ b/.github/workflows/action-format.yml
@@ -62,8 +62,6 @@ jobs:
 
       - name: 'Format code'
         run: ./bin/format.sh
-        env:
-          EXERCISM_PRETTIER_VERSION: '2.3.2'
 
       - name: 'Commit formatted code'
         run: |
--- a/.github/workflows/action-format.yml
+++ b/.github/workflows/action-format.yml
@@ -64,6 +64,8 @@ jobs:
         uses: actions/setup-node@25316bbc1f10ac9d8798711f44914b1cf3c4e954
         with:
           node-version: 14.x
+      - name: Install project development dependencies
+        run: npm i
       - name: 'Format code'
         run: ./bin/format.sh
 
--- a/.github/workflows/action-format.yml
+++ b/.github/workflows/action-format.yml
@@ -60,6 +60,10 @@ jobs:
           git clone $HEAD_REPO .
           git checkout -b "$HEAD_REF" "origin/$HEAD_REF"
 
+      - name: Use Node.js LTS (14.x)
+        uses: actions/setup-node@25316bbc1f10ac9d8798711f44914b1cf3c4e954
+        with:
+          node-version: 14.x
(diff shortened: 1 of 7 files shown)


SleeplessByte:
@laulujan Great start. You could check out your change by changing `package.json` in the root to `2.3.0`. It should then log your message in the "pre-check" job.

However, this *updates* the environment to be "correct", but it won't use the correct version when running the format workflow. Instead of "fixing" it, it's probably best to blow up (error), and/or post a comment to the PR that the version mis-matches.

laulujan:
Thanks, @SleeplessByte I committed a change that includes a message if the versions don't match, I also included a regex to compare only the version without the compatibility character, let me know if I understood correctly or if I still need to add something!

SleeplessByte:
Great! 

Now all we need is _throwing an error_ instead of commenting, so that it _surfaces_ on PRs :). After you've made that change, we can try it out.

laulujan:
Thanks @SleeplessByte I included the error in the new commit, let me know what you think!!

SleeplessByte:
Alright. Onto the next thing. 

Currently it's checking the _current_ environment (so the environment of when pr.ci.yml runs). However, `EXERCISM_PRETTIER_VERSION` is only set on a step in `action-format.yml` and `verify-code-formatting.yml`. We therefore _can't_ rely on `env` to read this value.

What it needs to do is:

- Read out _both_ values (`action-format.yml` and `verify-code-formatting.yml`)
- Check each of those instances

The rest is looking good :)

laulujan:
Thank you @SleeplessByte , I was wondering, is it ok if I  add https://github.com/nodeca/js-yaml as a dependency to read the yml files? Or do you know another approach I can take to read those files? Thanks in advance!!

SleeplessByte:
Yeah, we can _guess_ here and read the source files _as text_ (using node's regular `fs`). Why? Well, yes, we _could_ parse the `.yml` file, but we are really looking for `EXERCISM_PRETTIER_VERSION: x.y.z`. We know that that doesn't appear anywhere in those files except for the "right place".

> "But @sleeplessbyte, what about the future?"

Good point. When someone else adds another instance of `EXERCISM_PRETTIER_VERSION`, for example as a comment, your check will throw an error, and it will be up to that person to fix it ;). 

So in short: read the file as text, search for the value by parsing it as text, for example using regular expressions.

laulujan:
Great, thank you for your guidance, I included the changes in my latest commit, let me know what you think!

SleeplessByte:
@laulujan I've converted your idea and combined it with changes we have in https://github.com/exercism/javascript-test-runner among other repos. Thank you so much for working on this, it was the basis of those changes.

PR #1206, opened 2021-07-07
🤖 Simplify `hello-world` stub

ErikSchierboom:
Simplify the `hello-world` exercise's stub. The goal of this is to make things easier for students to get started and to introduce as little syntax as possible.

See https://github.com/exercism/v3-launch/issues/46

### DIFF
--- a/exercises/practice/hello-world/hello-world.js
+++ b/exercises/practice/hello-world/hello-world.js
@@ -3,4 +3,6 @@
 // convenience to get you started writing code faster.
 //
 
-export const hello = () => 'Goodbye, Mars!';
+export function hello() {
+  return 'Goodbye, Mars!';
+}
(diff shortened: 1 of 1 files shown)


SleeplessByte (approved):

exercises/practice/hello-world/hello-world.js
```diff
@@ -3,6 +3,4 @@
 // convenience to get you started writing code faster.
 //
 
-export const hello = () => {
-  throw new Error('Remove this statement and implement this function');
-};
+export const hello = () => 'Goodbye, Mars!';
```
SleeplessByte, line 6:
```suggestion
export function hello() {
  return 'Goodbye, Mars!';
}
```

I personally like this better, despite the `return`. What do you think @ErikSchierboom ?

ErikSchierboom: I definitely like that better. I just used whatever was in the example file.

PR #1226, opened 2021-07-24
Add icons for key features

angelikatyborska:
My proposal, feel free to change 🙂:

-  Run almost anywhere
-  Use any programming style
-  No types required
-  Concurrency is safe
-  Largest package repository
-  Designed by a committee

I'm assuming that "comity" was a typo because it's a law term according to wiki (https://en.wikipedia.org/wiki/Comity)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1706,32 +1706,32 @@
   ],
   "key_features": [
     {
-      "icon": "features-cross-platform",
+      "icon": "cross-platform",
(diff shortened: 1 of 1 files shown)


SleeplessByte:
Let's go with these. Changing them later is easy. Thank you! 

(And yes, typo, discussed earlier but didn't take the time to fix it)

PR #1227, opened 2021-07-24
Concept exercise review: annalyns-infiltration

angelikatyborska:
Two small problems:
- Hints for step 2 didn't show up because of a heading mismatch with the instructions.
- There were no hints for step 4. I don't know if you want any. I came up with one new hint that doesn't show up in the other steps that might make sense.


SleeplessByte (approved):

SleeplessByte:
Thank you!

PR #1097, opened 2021-04-16
[V3] Elyses Looping Enchantments: fix everything around this exercise

SleeplessByte:
# Reviewers, please spell/grammar check

Related #1017

Issue #945, opened 2020-07-10
Improve Concept Exercise: recursion


SleeplessByte:
## Getting started

Here you can read about what Concept Exercises are and how they are structured:

- [What are Concept Exercises?](https://exercism.org/docs/building/product/concept-exercises)
- [Concept Exercises](https://exercism.org/docs/building/tracks/concept-exercises)
- [Concepts](https://exercism.org/docs/building/tracks/concepts)

If you have not done so yet, it is probably also helpful to do a couple of "Learning Exercises" (this is how they are called on the site) yourself. You can also look at the code of an existing concept exercise like [`bird-watcher`](https://github.com/exercism/javascript/tree/main/exercises/concept/bird-watcher) (concept [`for-loops`](https://github.com/exercism/javascript/tree/main/concepts/for-loops)) for reference.

See the documentation _above_ (general documentation), as well as [How to implement a Concept Exercise in JavaScript](https://github.com/exercism/javascript/blob/main/reference/implementing-a-concept-exercise.md).

## Goal

[This PR](https://github.com/exercism/v3/pull/1420) has been merged [optimistically](https://github.com/exercism/v3/discussions/1725). The improvement goal is to implement:

- [ ] .docs/about.md
- [ ] .docs/hints.md
- [ ] .docs/instructions.md
- [ ] .docs/introduction.md
- [ ] languages/javascript/exercises/concept/README.md (check if updated)

Your mission, should you accept it, is to apply the story from the fork-original, or come up with a new one, write the step-by-step instructions, provide at least 1 hint per instruction (see other exercises for the format) and to provide some interesting facts about recursion or the ways to recurse in the `about.md` document. Optionally talk about tail-call optimisation/tail-recursion and the lack thereof in JavaScript.

> ⚠ The next message is of vital importance and _must_ be resolved using language in `instructions.md` and usage of `hints.md`. 

This exercise was **simplified** from the fork-original. This allows for an implementation with _recursion_ in the second fuction `orderPrice`. However, the test cases are structured in such a way so that the recursion solution will yield an error. These errors are more common than you think and this exercise introduces students to them. It is **vital** to name both errors in the _instructions.md_ and refer to _hints.md_ if a student needs help resolving them. 

```javascript
export function orderPrice([order, ...others]) {
  if (!order) {
    return 0
  }

  return pizzaPrice(order.pizza, ...order.extras) + orderPrice(others)
}
```

[...]

PR #1237, opened 2021-07-30
Rename conept exercise dirs: `closures` and `recursion`

angelikatyborska:
I noticed that there still are some concept exercises that are in directories named after concepts.

This PR renames:
- `closures` -> `coordinate-transformation`
- `recursion` -> `pizza-order`

There are 3 non-story directory names remaining, but I am not touching them:
- `array-loops` because it's already done in this draft PR https://github.com/exercism/javascript/pull/1097
- `errors` because it's not in the main `config.json` so it doesn't have a name. Should it be added there? is it waiting for improvements?
- `nullability` because it's deprecated


SleeplessByte (changes requested):
We probably need to fix more in those exercises, so putting it on hold

angelikatyborska:
In `recursion` yes, there's an issue: https://github.com/exercism/javascript/issues/945

`closures` looks pretty complete to me, except for the blurb.

Are you sure not merging this on its own is the best strategy? I'm not really volunteering to fix `recursion` so somebody would either need to branch off my branch or just redo the same changes. I don't really see any harm in merging just the dir renames.

SleeplessByte:
It's a good indication currently on if it's old or not :p.

That's my only concern.

I can probably make time this weekend.

SleeplessByte:
Narrator: I didn't

junedev:
@angelikatyborska I discussed with @SleeplessByte that I will have look whether anything else needs to be fixed.

junedev (approved):
Discussed with @SleeplessByte that we put the recursion exercise back to "wip" for now because too many things are missing.

Added some more small fixes and the missing blurbs.

PR #1238, opened 2021-07-30
Concept exrecise review: `closures`

angelikatyborska:
The resources in general hints wouldn't show up because nested lists are not allowed. I can't access https://whatthefork.is/closure at all. I assume it's not going back because "whatthefork" doesn't even show up as a website in my search engine, so it must have been dead for a while.

### DIFF
--- a/concepts/closures/links.json
+++ b/concepts/closures/links.json
@@ -12,6 +12,10 @@
     "description": "wiki-closure"
   },
   {
+    "url": "https://web.archive.org/web/20210702195218if_/https://whatthefork.is/closure",
+    "description": "what-the-fork-is-a-closure"
+  },
+  {
     "url": "https://javascript.info/closure",
     "description": "javascript-info-scope-closure"
   },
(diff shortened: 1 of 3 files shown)


angelikatyborska:
@SleeplessByte a review pretty please?

concepts/closures/links.json
```diff
@@ -11,10 +11,6 @@
     "url": "https://en.wikipedia.org/wiki/Closure_(computer_programming)",
     "description": "wiki-closure"
   },
-  {
-    "url": "https://whatthefork.is/closure",
```
SleeplessByte, line 15: It was such a good article.

How about we link to https://web.archive.org/web/20210702195218if_/https://whatthefork.is/closure instead?

SleeplessByte (approved):

concepts/closures/links.json
angelikatyborska: Sure, if it's that good 😁

SleeplessByte (approved):

SleeplessByte:
Thank you for this _exrecise_ review.

PR #1248, opened 2021-08-03
Improve Concept Exercise: Type Conversion

### LINKED ISSUE #947: [V3] Implement new Concept Exercise: type-conversion
This issue describes how to implement the `type-conversion` concept exercise for the `javascript` track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

See the documentation _above_ (general documentation), as well as [How to implement a Concept Exercise in JavaScript](https://github.com/exercism/v3/blob/master/languages/javascript/reference/implementing-a-concept-exercise.md).

## Goal

The goal of this exercise is to teach the student how they can "cast" between types.

## JavaScript specificness

ℹ Many languages have a way to _cast_ a value into a (different) type. JavaScript does not have such constructs. However, the global objects for truthy primitives (`String`, `Number`, `Boolean`) also exist as function (`String(val)`, `Number(val)`, `Boolean(val)`) which convert the input to the _primitive_ type (non-boxed). 

## Learning objectives

- Using (some) prototype functions:
  - `* ➡ string`: `.toString()`
  - `* ➡ Array<*>`: `[Symbol.iterator]` (destructuring [works with any iterable](https://exploringjs.com/impatient-js/ch_destructuring.html#array-destructuring-works-with-any-iterable))
   - `Array<*> ➡ string`: `.join`
   - `Date ➡ number`: `.getTime`
   - `Date ➡ string`: `.toISO8601`
   - `Object ➡ string`: `JSON.stringify` 
   - `Array ➡ string`: `JSON.stringify` 
   - `string ➡ Object `: `JSON.parse` 
   - `string ➡ Array`: `JSON.parse` 
- Using global functions:
  - `* ➡ boolean`: `Boolean(value)`
  - `* ➡ number`: `Number(value)`
  - `* ➡ string`: `String(value)`
  - `...* ➡ Array`: `Array(...values)` 
- Using operators:
  - ` * ➡ boolean`: `!value` and `!!value`
- Using constructors:   
  - `number ➡ Date`: `new Date(unix)`
  - `string ➡ Date`: `new Date(iso8601)`

If this list can not be turned into a single exercise, you can _remove_ the `Date` related ones, and put that into a `dates` concept exercise. Also **remove** the `JSON` related ones, and put that into a `json` concept exercise!

## Out of scope

- `number#toString(radix)` (probably need a different exercise about non-base-10 numbers)
- `new Function(value)` (we're not teaching `eval` here, and does MORE than type conversion)
- Loose type conversion that also does invalidation:
  - `parseInt` (probably want an input validation exercise, does MORE than type conversion)
  - `parseFloat` (probably want an input validation exercise, does MORE than type conversion)
- Unary operators that basically do an implicit type cast (such as `==` does)
  - `+string` (not idiomatic)
  - `~string` (not idiomatic)
- Binary operators that do an implicit type cast (actually `==`)
  - `string ^ string` (not idiomatic)
  - `string * string` (not idiomatic)

## Concepts

- `type-conversion`

## Prerequisites

- `basics`
- concepts about the types
- ... (do you think something is missing?)

## Help

You can choose to do this solo-style, or collaborate with multiple people on this. My suggestion would be is to 

1. **first accept this issue by saying "I'd like to work on this"** (you will not get a response from me. Just go with it) and optionally request that someone works with you (and wait for a second person to accept your request).
2. Use this issue to write down what you need (collect reference material) and discuss, if necessary, what to write and what not to write.
3. Draft a PR and request the feedback from at least one other JavaScript member (I recommend just pinging the channel on slack).

### LINKED ISSUE #1247: Improve Concept Exercise: Type Conversion (lucky numbers)
I will work on improving this exercise and concept to fit the learning curve we are aiming for, see https://github.com/exercism/javascript/issues/951#issuecomment-778834953.

I will also check everything fits the latest specs etc.

junedev:
Closes #1247 
Closes #947 (@SleeplessByte I diverged a bit from what you wrote there, hope that's ok for now.)

### Notes

- I don't cover the "behind the scenes" currently in about.md, e.g. boxed vs. non-boxed primitives, toPrimitive, details on toString. This is to save time right now but also because it is not clear to me yet whats the best place for this info in our concept tree.
- I also don't cover all the weirdness currently, like object + string. Not sure how much value there is going into this tbh.
- I replaced the third task because the old one didn't practice anything that wasn't already part of task 1 and 2. The new task now practices implicit conversion to boolean.
- There are lots of places where I use MDN links currently but once we have more concepts of our own (e.g. on JSON) we could link to those.
- No analyzer ideas yet in the design.md file, also to save some time before launch.


angelikatyborska:
FYI the PR description references this PR with "closes", I think you meant to reference the issue https://github.com/exercism/javascript/issues/1247 instead?

junedev:
Yes. 🙈 Fixed.

concepts/type-conversion/about.md
```diff
@@ -1,10 +1,272 @@
...
+```
+
+Note that because of the described rules, the following values are truthy in JavaScript although they might count as falsy in other languages.
+
+```javascript
+Boolean('0');
+// => true
+
+Boolean('false');
+// => true
```
angelikatyborska, line 44: I'm not much of a Polyglot but I'm curious which languages would consider those strings to be falsy?

In Ruby and Elixir, the concept of falsy is very limited, only `0` and `nil` are falsy.

concepts/type-conversion/about.md
```diff
@@ -1,10 +1,272 @@
...
+// => 1
+
+Number(false);
+// => 0
+
+Number(null);
+// => 0
+
+Number(undefined);
+// => NaN
```
angelikatyborska, line 93: So `Number(undefined)` is `NaN`, but `Number()` is `0`? 🤯

concepts/type-conversion/about.md
```diff
@@ -1,10 +1,272 @@
...
+```
+
+JavaScript also provides the functions `parseInt` and `parseFloat`.
+They apply laxer rules as to which strings can be converted to a number.
+Because of that, `Number` should be preferred as conversion function to avoid unexpected outcomes.
```
angelikatyborska, line 98: 😇 I learned something today... I have so many codebases to fix. I had so many bugs caused by forgetting the second (base) argument to `parseInt`...

concepts/type-conversion/about.md
```diff
@@ -1,10 +1,272 @@
...
+const obj = {
+  name: 'Gilda Guerra',
+  address: {
+    city: 'Goiânia',
+  },
+};
+
+JSON.stringify(obj);
+// => '{"name":"Gilda Guerra","address":{"city":"Goiânia"}}'
+```
```
angelikatyborska, line 158: Should this text also mention `JSON.parse`?

concepts/type-conversion/about.md
```diff
@@ -1,10 +1,272 @@
...
+  It returns one of the original operands (see [MDN on Logical Operators][mdn-logical-operators]).
+
+  ```javascript
+  null || 'hello';
+  // => 'hello'
+  ```
+
+### String Context
+
+If the addition operator `+` is used for primitive values and one operand is a string, the other one will be coerced into a string as well (if necessary).
```
angelikatyborska, line 205: I think the "if necessary" is already implied by saying "one operand is a string". If only one of the is a string, it is necessary.

```suggestion
If the addition operator `+` is used for primitive values and one operand is a string, the other one will be coerced into a string as well.
```

exercises/concept/lucky-numbers/.docs/instructions.md
```diff
@@ -1,21 +1,41 @@
 # Instructions
 
-## 1. Find the sum of two arrays
+Your friend Kojo is a big fan of numbers.
+He has a small website for playing around with numbers: www.fun-with-numbers.com.
```
angelikatyborska, line 4: I can't believe such a domain doesn't exist 😮

exercises/concept/lucky-numbers/.meta/config.json
```diff
@@ -1,6 +1,6 @@
 {
-  "blurb": "TODO: add blurb for lucky-numbers exercise",
-  "authors": ["shubhsk88"],
+  "blurb": "Practice type conversion while helping your friend Kojo with his website.",
```
angelikatyborska, line 2: I know blurbs are like "whatever", but... this one feels a bit weird to me because it doesn't have any connection to the exercise's name. Maybe:
```suggestion
  "blurb": "Practice type conversion while helping your friend Kojo with his website www.fun-with-numbers.com.",
```

angelikatyborska (approved):
I read through everything and tried to solve the exercise in my head before looking up the hints. It all makes sense to me, I didn't find any problems.

concepts/type-conversion/about.md
junedev: you are right but I didn't know this either :exploding_head:

exercises/concept/lucky-numbers/.docs/instructions.md
junedev: same, maybe someone who does the exercise will by it and implement the games :smile:

concepts/type-conversion/about.md
junedev: On this page https://javascript.info/type-conversions#boolean-conversion they say `'0'` is falsy in PHP. `'false'` is not really an example from another language, it is rather something that might be unexpected. I run into this last week when I wrote `if (<some environment variable>)`. Then I was confused why the if block was still executed even though the environment variable was set to false. Anyway, I will use the phrasing from the introduction instead so people don't start wondering about which languages are meant here.

concepts/type-conversion/about.md
junedev: I added a sentence about this behavior. I couldn't find an explanation though.
@SleeplessByte Do you know what's going on there?

concepts/type-conversion/about.md
junedev: I didn't want to dive too much into this because it will have a separate exercise/concept but I agree that if I show the one direction I should also mention the other way. I added a sentence to cover that.

concepts/type-conversion/about.md
junedev: fixed in all the places

exercises/concept/lucky-numbers/.meta/config.json
junedev: fixed, I do think blurbs are important

junedev:
@SleeplessByte I will merge this now. Feel free to also do another review here later if you want and I will create a PR with the fixes.

concepts/type-conversion/about.md
SleeplessByte: I do. It's specced

```
Number ( [ value ] )

Returns a Number value (not a Number object) computed by ToNumber(value) if value was supplied, else returns +0.
```

So this has a special case for _no_ argument, which is `+0`.

Then we check `ToNumber`:

Argument Type | Result
-- | --
Undefined | NaN
Null | +0
Boolean | The result is 1 if the argument is true. The result is +0 if the argument is false.
Number | The result equals the input argument (no conversion).
String | See grammar and note below.
Object | Apply the following steps:Let primValue be ToPrimitive(input argument, hint Number).Return ToNumber(primValue).

and _thus_ explicitly passing `undefined` is a different result (`NaN`) than passing _nothing_.

junedev: I got that part but I was wondering why it was specced that way.

SleeplessByte: Probably legacy / `new String`, `new Array`, `new Number` should be "the same" as in create an "empty" value of that type (`''`, `[]`, and `+0`). It would be weird if that created a `NaN`. However passing in `undefined` makes it subject to coercion. Something like that.

Issue #1013, opened 2021-02-26
CI should check if repository value of package.json is sane


SleeplessByte:
Related #1007.

## Description

We have two scripts that check the integrity of the repository files, or pr changed files respectively: [`scripts/ci-check`] and [`scripts/pr-check`](https://github.com/exercism/javascript/blob/main/scripts/pr-check).

These are their steps:

1. check if stub file exists
2. check if their configuration files have the right _integrity_ (check if the files in `exercises/*/*/` match those of the root, because we want the same `eslint`, `babel`, etc. configuration for all exercises).
3. check if the files pass the linter (`eslint`).

## Current behaviour

The integrity step correctly removes some fields from `package.json` before comparing the root file with the exercise file. However, this makes it possible to have a exercise `package.json` with incorrect values for the `repository` key.

Currently, the `repository` key's value is completely ignored.

## Expected behaviour

Instead, the `package.json` in the exercise directory should:

1. match `type` of the root `package.json` `repository`
2. match `url` of the root `package.json` `repository`
3. have the correct `directory` based on the path to the exercise

AbhilashJN:
I can work on this one.

PR #1254, opened 2021-08-05
Issue1013:  CI should check if repository value of package.json is sane

AbhilashJN:
- Added script `scripts/directory-check` to check if `directory` field in the package.json for each exercise is correct.
 - Added documentation in CONTRIBUTING.md for `directory-check` script
 - Run `directory-check` script from `ci-check` and `pr-check` scripts in the integrity check step.
 
Ref Issue: #1013

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -317,6 +317,27 @@ Run this script to check if package name in package.json of exercises is in expe
 
 Run this script to check if there is any duplicate package name.
 
+#### `directory-check`
+
+```js
+/**
+ * Run this script (from root directory): npx babel-node scripts/directory-check
+ *
+ * This will run following checks:
+ *
+ * 1. Package has the correct directory based on the path to the exercise.
+ *
+ * This script also allows fixing these directories: npx babel-node scripts/directory-check --fix
+ */
+```
+
+Run this script to check if package repository directory in package.json of exercises is in expected format or to fix it.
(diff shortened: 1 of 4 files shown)


AbhilashJN:
The referenced also mentions that
```
Instead, the package.json in the exercise directory should:

match type of the root package.json repository
match url of the root package.json repository
```

However, it seems to me that the `type` and `url` fields are already being checked as part of the checksum, so I am not sure if anything else needs to be done for this. Do let me know if I have misunderstood this part.

SleeplessByte:
Are we certain it's being checked as part of the checksum? It seems that if each `concept/***/slug` directory contains a `package.json` with a _different_ value for `repository.url`, it likely is **not** part of the checksum (nor should it), because it would be different 👍🏽

AbhilashJN:
Currently it looks like for all the exercises, the `repository.url` field has the value `https://github.com/exercism/javascript`. Is it supposed to be different for each exercise, perhaps something like `https://github.com/exercism/javascript/exercises/<exercise-path>` ?

SleeplessByte:
From the references issue:

```
Instead, the package.json in the exercise directory should:

match type of the root package.json repository
match url of the root package.json repository
have the correct directory based on the path to the exercise
```

So the url should be the same, the type should be the same, but the `directory` should be set.

AbhilashJN:
In that case, it looks like no further changes are required. The type and url are being compared to the root package.json as a part of the checksum.

PR #1257, opened 2021-08-05
fix(exercises): Retain balance in bank-account

ffflorian:


SleeplessByte (changes requested):
The test is called `reopened account does not retain balance`, so I don't think this _fixes_ anything.

ffflorian:
Oh no, I misread it! My bad... thanks for checking so quickly! :eyes:

PR #1268, opened 2021-08-14
Add missing commas between major and minor scale names

ramitmittal:
All scale names are rendered on the same line. The comma may not make sense in the markdown file but improves readability on the rendered page.


tejasbubane (approved):
Thanks a lot for spotting and fixing this.

PR #1277, opened 2021-08-25
Fix Concept Exercise: Errors

junedev:
This PR is to make the errors concept exercise fit the latest spec and sets it to "beta" status.
(see https://github.com/exercism/javascript/issues/1017)

It does not update the current content.


SleeplessByte (approved):
Looks very sensible. Thanks!

PR #1305, opened 2021-09-04
Fix links to old docs repo

iHiD:


SleeplessByte:
Good changes, but already applied!

Issue #1182, opened 2021-06-15
Check corejs version in babel configuration versus package.json


SleeplessByte:
[`core-js`](https://github.com/zloirock/core-js#babel) is configured in our `babel.config.js` / `.babelrc`. It's version in our dependency list in `package.json` (currently `3.14.x`) must match the babel setting without the patch version. 

In other words, given a `package.json` version of `3.14.x`, the configuration setting must be `3.14`.

Your task, should you accept it, is to add a check to the "precheck" scripts that run automatically on PR requests is to check if the package.json version matches the babel configuration version, up to the minor (major.minor.patch) version.

ffflorian:
I'd like to take over this task :slightly_smiling_face:

junedev:
@ffflorian You don't need to check all package.json and babel.config.js files, you only need to check those in the root directory. They are then synced from there to all the exercises via another script, so if they are correct in the root directory, they will be correct everywhere (or CI will complain about it).

ffflorian:
[Done](https://github.com/exercism/javascript/pull/1313) :slightly_smiling_face:

PR #1313, opened 2021-09-05
feat: Check corejs version in babel configuration versus package.json

ffflorian:
This closes https://github.com/exercism/javascript/issues/1182.


SleeplessByte (approved):
My bash-fu is weeak-ish (all the nastyness you saw thusfar is mostly my doing), but from what I could see, this is reading out the version (line 3 and 4), for example:

```bash
babel_version=3.17
package_version=3.17.2
```

...and then exploding each one into an array, and then comparing the first two parts.

Perhaps a less complex way to do this check that `package_version` starts with `babel_version` ?

Alas, the only thing I need to know before merge is if this will work in both bash and dash / all bourne shells :)

junedev:
@SleeplessByte The "starts with" version wouldn't really satisfy your requirements. `3.17.2` and `3` would then be valid entries for the Babel config.

SleeplessByte:
Meh. Yeah. That's true. And once you have to check for both starts with + at least one `.` and a number before and after it, it becomes as complex as this is. 

Okay. Let's do it!

PR #1318, opened 2021-09-06
Fix all ts-check issues in concept exercises

junedev:
I found out my editor did not show me the ts-check issues at all. I fixed the editor setting and then fixed all the typing issues.

This PR contains
* removing env.d.ts files as they are not needed anymore
* fix typing issues
* fix typing import paths in exemplar.js files


SleeplessByte (approved):
Nice!

PR #1324, opened 2021-09-07
reduced pythagorean-triplet's time complexity from cubic to quadratic 

eakray:
In the previous implementation the triplets function had a time complexity of O(n^3). This commit reduces it to O(n^2). I believe the current best known optimal solution for this type of problem is [O(n^2(log log n)/log ^2n)](https://en.wikipedia.org/wiki/3SUM) but in order to achieve that result we would need to make changes to this solution which would negatively impact readability/understandability. The final test still takes a long time (~16s) but I don't think that can be significantly improved for an input of that size.

### DIFF
--- a/exercises/practice/pythagorean-triplet/pythagorean-triplet.spec.js
+++ b/exercises/practice/pythagorean-triplet/pythagorean-triplet.spec.js
@@ -58,7 +58,7 @@ describe('Triplet', () => {
     ]);
   });
 
-  test.skip('triplets for large number', () => {
+  xtest('triplets for large number', () => {
     expect(tripletsWithSum(30000)).toEqual([
       [1200, 14375, 14425],
       [1875, 14000, 14125],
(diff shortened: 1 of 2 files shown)


exercises/practice/pythagorean-triplet/pythagorean-triplet.spec.js
```diff
@@ -58,7 +58,7 @@ describe('Triplet', () => {
     ]);
   });
 
-  test.skip('triplets for large number', () => {
+  xtest('triplets for large number', () => {
```
SleeplessByte, line 61: If it still takes ~ 16 s, we can't "unskip" this, because the web UI has a maximum of 10 s of run time. Apart from that, all these changes are great!

eakray: Would it be possible to run the test on a smaller input?

eakray: an input of 25000 takes ~9.5s and returns [ [ 5000, 9375, 10625 ] ]. Or do you mean i should mark the test as skipped but keep the code changes?

SleeplessByte: That last thing! So people can run it likely but it won't run in the web.

SleeplessByte (approved):
Looking great. I'll have the workflow format your code by writing:

`/format`

PR #1344, opened 2021-09-14
Update Annalyns Infiltration instructions.md

LoyalEnv0y:
I changed the 4th item in instructions because it says that 

> _the prisoner can only be freed if they were not asleep_

however in the 2nd test labeled as `when everyone is asleep and pet dog is not present` of the '`can free prisoner`' tests, it expects the student to return `true` when the prisoner is asleep.

### DIFF
--- a/exercises/concept/annalyns-infiltration/.docs/instructions.md
+++ b/exercises/concept/annalyns-infiltration/.docs/instructions.md
@@ -13,8 +13,9 @@ Having found the kidnappers, Annalyn considers which of the following actions sh
 - _Fast attack_: a fast attack can be made if the **knight** is **_sleeping_**, as it takes time for him to get his armor on, so he will be vulnerable.
 - _Spy_: the group can be spied upon if **at least one** of them is **_awake_**. Otherwise, spying is a waste of time.
 - _Signal prisoner_: the prisoner can be signalled using bird sounds if the **prisoner** is **_awake_** and the **archer** is **_sleeping_**, as archers are trained in bird signaling so they could intercept the message.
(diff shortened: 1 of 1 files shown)


junedev:
Hi there, I have trouble understanding your issue. If I look at the second test case at says "... and pet dog is present" (without "not"), so it fits the current instructions.
See https://github.com/exercism/javascript/blob/main/exercises/concept/annalyns-infiltration/annalyns-infiltration.spec.js#L156

The instructions say that the prisoner needs to be awake for the first condition but then it goes on with
> The prisoner can also be freed if the archer is sleeping and Annalyn has her pet dog with her
For the second condition there is no requirement whether the prisoner is awake or not.

LoyalEnv0y:
Hello, I am sorry i meant the 3rd* test instead of 2nd. It is labeled as 'when everyone is asleep and pet dog is not present'.
See https://github.com/exercism/javascript/blob/0d9bf4b176897e968d2c660a4aee44f00ba3c31b/exercises/concept/annalyns-infiltration/annalyns-infiltration.spec.js#L173

> _The prisoner can also be freed if the archer is sleeping and Annalyn has her pet dog with her_

i didn't think that this part of the instruction clearly states that we don't need the prisoner to be awake if Annalyn's dog is present. So i tried to rewrite it.

I just noticed that i missed this part of the instruction:

> _and the archer can't equip his bow fast enough to prevent the prisoner from being freed._

so it means, even if Annalyn startles the prisoner archer won't be able to stop them. It is my fault that i missed this part. 
Sorry if i wasted your time.

PR #1345, opened 2021-09-14
:bug: Fix 'promises' exercise specs

### LINKED ISSUE #1331: Improve Practice Exercise "Promises"
I was trying to help a student but the exercise makes no sense to me.  If I implement just a TINY implementation:

```
export const promisify = (fn) => {
  return (v) => {
    return new Promise(() => {
    });
  };
};
```

All the tests start turning green, despite the fact that I'm not even calling `fn`, not rejecting or resolving anything... if I can't trust the test suite to guide me then I'm not sure how any student is supposed to get thru this complex exercise.

If the function we're supposed to be wrapping is of the form `fn(...args, (error, result) => {})` then that should probably be mentioned somewhere in the instructions, perhaps with examples of before/after, etc.

TomPradat:
Fix #1331 

I've tested @joshgoebel  implementation of `promisify` and the specs for this function now fail correctly.

TODO :

- [x] Make sure all the other tests fail correctly

### DIFF
--- a/exercises/practice/promises/promises.spec.js
+++ b/exercises/practice/promises/promises.spec.js
@@ -26,12 +26,12 @@ describe('promises', () => {
     xtest("promisified function resolves to a callback's success value", () => {
       const SUCCESS = 'success';
       const fastPromise = promisify(fastCallbackFn);
-      expect(fastPromise(SUCCESS)).resolves.toEqual(SUCCESS);
+      return expect(fastPromise(SUCCESS)).resolves.toEqual(SUCCESS);
     });
 
     xtest("promisified function rejects a callback's error", () => {
       const failedPromise = promisify(failedCallbackFn);
-      expect(failedPromise(null)).rejects.toEqual(failedCallback);
+      return expect(failedPromise(null)).rejects.toEqual(failedCallback);
     });
   });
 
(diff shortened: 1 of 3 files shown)


exercises/practice/promises/promises.spec.js
```diff
@@ -26,12 +26,12 @@ describe('promises', () => {
     xtest("promisified function resolves to a callback's success value", () => {
       const SUCCESS = 'success';
       const fastPromise = promisify(fastCallbackFn);
-      expect(fastPromise(SUCCESS)).resolves.toEqual(SUCCESS);
+      return expect(fastPromise(SUCCESS)).resolves.toEqual(SUCCESS);
```
TomPradat, line 29: The `return` is needed, see https://jestjs.io/docs/26.x/asynchronous ("Be sure to return the promise - if you omit this return statement, your test will complete before the promise returned from fetchData resolves and then() has a chance to execute the callback.")

junedev:
As @joshgoebel said in the issue, the tests are only one of multiple issues with this exercise. So until someone has time to overhaul it more generally (clearer instructions, possibly simplify) I would agree we should set the status to "wip" for now. Could you include that in your PR?

SleeplessByte:
Yep. These changes are perfect @TomPradat , but please include the change to WIP

SleeplessByte (approved):

PR #1346, opened 2021-09-14
Update instructions.md

keshiki808:
I corrected the terminology in the table.
They are amino acids.


junedev:
@keshiki808 Thanks for the input but this description does not originate in this repository. This exercise (the instructions and test cases) originates from this repository https://github.com/exercism/problem-specifications. From there, the data is then "synced" into various tracks. Fixes to the instructions should be applied there so all tracks can benefit from it.

I will close this PR but feel free to open a new one in the problem-specification repository. When you do so, have a look at the rest of the exercise text. It looks to me, like there are more places where the wording needs to be corrected (Protein -> Amino Acid).

Be aware that changes to the problem specifications need a couple of approvals before they can be merged. I would recommend you include a couple more words explaining why the current text is incorrect.

PR #1349, opened 2021-09-16
Updating Elyses Analytic Enchantment JSDoc

arlm:
Enhancing the JsDoc description of the return value so that it matches what is described on the exercise description on the README.md file.


junedev (changes requested):
Thanks for your input! Could you please apply the same fixes to the .meta/exemplar.js file, the JSDoc comments also exist there.

junedev (approved):

PR #1355, opened 2021-09-17
Clarify the 'promises' exercise instructions

### LINKED ISSUE #1331: Improve Practice Exercise "Promises"

TomPradat:
Fix #1331 

TODO :
- [x] Make the exercise available again (by removing the wip status)

### DIFF
--- a/config.json
+++ b/config.json
@@ -984,8 +984,7 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 6,
-        "topics": ["promises", "error_handling", "higher_order_functions"],
-        "status": "wip"
+        "topics": ["promises", "error_handling", "higher_order_functions"]
       },
       {
         "slug": "yacht",
--- a/config.json
+++ b/config.json
@@ -983,7 +983,7 @@
         "core": false,
         "practices": [],
         "prerequisites": [],
-        "difficulty": 4,
+        "difficulty": 6,
         "topics": ["promises", "error_handling", "higher_order_functions"],
         "status": "wip"
       },
--- a/exercises/practice/promises/.docs/instructions.md
+++ b/exercises/practice/promises/.docs/instructions.md
@@ -1,11 +1,30 @@
 # Instructions
 
-Given the basic `Promise`, implement the following functions:
+The two objectives of this exercise are :
 
-- `promisify`: takes a callback that is turned into a function that returns a `Promise`
+1. Implement a `promisify` function that turns a function using the "callback pattern" into a function that returns a `Promise`. See the example below.
+
+```javascript
+function fetchProduct(productId, function(error, data) {
+    if (error) {
+        // Handle the error
+    } else {
+        // Make something with your data
+    }
+})
+
+const fetchProductAsPromise = promisify(fetchProduct);
+
+// Now you got a function `fetchProductAsPromise`
+// that returns a promise
+fetchProductAsPromise(productId)
+    .then((data) => {})
+    .catch((error) => {});
+```
+
+2. Re-implement the following built-ins `Promise` methods (without using them)
+
 - `all`: takes an array of promises and resolves when _all_ of them are resolved, or rejects when _one_ of them rejects.
 - `allSettled`: takes an array of promises and resolves when _all_ of them either resolve or reject.
 - `race`: takes an array of promises and resolves or rejects with the value of the _first_ promise that resolves or rejects.
 - `any`: takes an array of promises and resolves when _one_ of them resolves, or rejects when _all_ of them reject.
-
-Do not use built-in `Promise` functions other than creating a `Promise`, `then` and `catch`.
--- a/exercises/practice/promises/.docs/introduction.md
+++ b/exercises/practice/promises/.docs/introduction.md
@@ -8,13 +8,13 @@ A callback is a function that is passed as an argument to another function and w
 function callback(error, arg2, arg3) {}
 ```
 
-Now is it related to asynchronous code ?
+How is it related to asynchronous code ?
 
-Historically, callbacks have been used in order to allow people to do some work after an asynchronous task was done and without blocking the whole program.
+Historically, callbacks have been used in order to allow us to do some work after an asynchronous task was done and without blocking the whole program.
 
 ```javascript
 fetchProduct(productId, function (error, data) {
-  if (error !== null) {
+  if (error) {
     // Handle the error
   } else {
     // Do some work
--- a/exercises/practice/promises/.docs/introduction.md
+++ b/exercises/practice/promises/.docs/introduction.md
@@ -1,0 +1,25 @@
+# Introduction
+
+Before the `Promise` class was introduced, there was only one way to deal with asynchronous code : the _callback pattern_.
+
+A callback is a function that accept an "error" as first parameter (see example below)
+
+```javascript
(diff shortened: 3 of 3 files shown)


config.json
```diff
@@ -975,7 +975,7 @@
         "core": false,
         "practices": [],
         "prerequisites": [],
-        "difficulty": 4,
+        "difficulty": 6,
```
TomPradat, line 978: IMO the difficulty is underrated

exercises/practice/promises/.docs/introduction.md
```diff
@@ -0,0 +1,25 @@
+# Introduction
```
TomPradat, line 1: I'm not sure how this file is displayed on the website, will it be visible by default before the instructions?

exercises/practice/promises/.docs/instructions.md
```diff
@@ -1,11 +1,30 @@
...
+const fetchProductAsPromise = promisify(fetchProduct);
+
+// Now you got a function `fetchProductAsPromise`
+// that returns a promise
+fetchProductAsPromise(productId)
+    .then((data) => {})
+    .catch((error) => {});
+```
+
+2. Re-implement the following built-ins `Promise` methods (without using them)
```
TomPradat, line 25: Should we update the tests to make sure the pupils are not using the built-in methods?

junedev: We usually don't focus on testing that a student does not cheat so I don't think this should be a priority right now.

exercises/practice/promises/.docs/introduction.md
junedev: I am pretty sure it will look the same as all the concept exercises where you first see the introduction text and then the instructions.

junedev:
@joshgoebel Can you have a look whether you think the exercise is ok now?

exercises/practice/promises/.docs/introduction.md
```diff
@@ -0,0 +1,25 @@
...
+
+Before the `Promise` class was introduced, there was only one way to deal with asynchronous code : the _callback pattern_.
+
+A callback is a function that accept an "error" as first parameter (see example below)
```
junedev, line 5: I think a function with a different pattern can still be a callback, so maybe rephrase this, maybe like this:
```suggestion
A callback is a function that is passed as an argument to another function and will be called once some action in that other function has finished. A common pattern for those callback functions is that they accept an "error" as first parameter (see example below).
```

exercises/practice/promises/.docs/introduction.md
```diff
@@ -0,0 +1,25 @@
...
+function callback(error, arg2, arg3) {}
+```
+
+Now is it related to asynchronous code ?
+
+Historically, callbacks have been used in order to allow people to do some work after an asynchronous task was done and without blocking the whole program.
+
+```javascript
+fetchProduct(productId, function (error, data) {
+  if (error !== null) {
```
junedev, line 17: I am no expert on this pattern but shouldn't this be `if (error)` so it also works if error is `undefined` for example?

junedev (approved):
Left some minor comments.  
In general I would be ok to up-wip this and see whether students are ok with the exercise now.

exercises/practice/promises/.docs/introduction.md
TomPradat: Good catch, fixed in 614fa7060fd3cb084c481a90fef2ed0772c798bb

joshgoebel:
Let me pull this down and have a look.

joshgoebel:
Much much better I think.  

```
export const all = (promises) =>
  promises.reduce(
    async (acc, promise) => (await acc).concat(await promise),
    Promise.resolve([])
  );
```

Just out of curiosity this isn't concurrent, is it?  It seems like each promise stacks, having to wait until the one before it finishes before the array is resolved and then concat for the next item can be called, correct?

junedev (approved):

TomPradat:
> Much much better I think.
> 
> ```
> export const all = (promises) =>
>   promises.reduce(
>     async (acc, promise) => (await acc).concat(await promise),
>     Promise.resolve([])
>   );
> ```
> 
> Just out of curiosity this isn't concurrent, is it? It seems like each promise stacks, having to wait until the one before it finishes before the array is resolved and then concat for the next item can be called, correct?

I'm not used to using `reduce` with async functions but I have tested this in the browser and it looks like it is **CONCURRENT** :

```javascript
let all = (promises) =>
  promises.reduce(
    async (acc, promise) => (await acc).concat(await promise),
    Promise.resolve([])
  );
  
  let createPromise = () =>
      new Promise((resolve) => {
          setTimeout(() => {
               console.log(performance.now(), "hello");
               resolve();
          }, 1000);
      });

all([createPromise(), createPromise(), createPromise(), createPromise()]).then(() => console.log("finished"));
```

Result :

```
2540591.3000000003 'hello'
2540591.5000000005 'hello'
2540591.6 'hello'
2540591.6999999997 'hello'
finished
```

I must admit I don't quite understand why though (or maybe something is wrong with my test code).

TomPradat:
@junedev Should i remove the "wip" status before this gets merged ?

junedev:
yes please

joshgoebel:
@TomPradat Are you sure?  You're creating them all at once (in the array creating)... so their setTimeout will be concurrent, that doesn't mean `reduce` is... I don't see how it can be... it has to wait, one at a time - since array addition is sequential.  Perhaps this doesn't matter in practice though since the time spent in reduce is so small...?

TomPradat:
> @TomPradat Are you sure? You're creating them all at once (in the array creating)... so their setTimeout will be concurrent, that doesn't mean `reduce` is... I don't see how it can be... it has to wait, one at a time - since array addition is sequential. Perhaps this doesn't matter in practice though since the time spent in reduce is so small...?

You are perfectly right, I came to the same conclusion after some more thoughts. The `reduce` is sequential :

```javascript
let all = (promises) =>
  promises.reduce(
    async (acc, promise) => (await acc).concat(await promise()),
    Promise.resolve([])
  );
  
  let createPromise = () =>
      new Promise((resolve) => {
          setTimeout(() => {
               console.log(performance.now(), "hello");
               resolve();
          }, 1000);
      });

all([createPromise, createPromise, createPromise, createPromise]).then(() => console.log("finished"));
```
> Note : The difference with the previous sample is that the promise is being created in the `reduce` instead of being created in the array

Result :

```
3360643.1999999997 'hello'
3361644.1 'hello'
3362644.6999999997 'hello'
3363645.4 'hello'
finished
```

So to conclude, the `reduce` is sequential but it doesn't matter as the promises run in parallel. It adds a small overhead that is probably (IMO) negligible

TomPradat:
> yes please

Done in ddd27138

PR #1357, opened 2021-09-18
refactor: :pencil2: retry 3 times not 2

AndrewLawendy:
- It's asked on the next line not to exceed 3 times
- In the test case `it requests at most three times (does not retry thrice or more)` it excepts to throw an error if the tries are equal or greater than 3

### DIFF
--- a/exercises/concept/translation-service/.docs/instructions.md
+++ b/exercises/concept/translation-service/.docs/instructions.md
@@ -84,7 +84,7 @@ service.batch([]);
 // => Promise<...> rejects BatchIsEmpty()
 ```
 
-## 3. Request a translation, retrying at most 2 times
+## 3. Request a translation, retrying at most 3 times
 
 Implement a function `request(text)` that _requests_ a translation, with automatic retries, up to a total of **3 calls** for the same request.
 


SleeplessByte:
Retrying 3 times is calling it four times.

AndrewLawendy:
> Retrying 3 times is calling it four times.

Don't get your comment..

SleeplessByte:
The task asks to call it at most three times:

- call (ok)
- call (ok)
- call (ok)
- call (not ok)

When you say "retrying at most 3 times", you can call it four times:

- call (ok)
- call, this is the first retry (ok)
- call, this is the second retry (ok)
- call, this is the third retry (not ok)

So saying "you can retry three times" is the same as saying "you can _try_ it four times".

Is there a better way to word this entire thing so it makes more sense :D ?

AndrewLawendy:
> The task asks to call it at most three times:
> 
> * call (ok)
> * call (ok)
> * call (ok)
> * call (not ok)
> 
> When you say "retrying at most 3 times", you can call it four times:
> 
> * call (ok)
> * call, this is the first retry (ok)
> * call, this is the second retry (ok)
> * call, this is the third retry (not ok)
> 
> So saying "you can retry three times" is the same as saying "you can _try_ it four times".
> 
> Is there a better way to word this entire thing so it makes more sense :D ?

Haaa, OK
I get it now, but I still think it should be reworded somehow as it seems issue like..

PR #1366, opened 2021-09-21
Simplify tests for Poetry Club Door Policy

mmmmmrob:
Suggest removing some of the logic in the tests to make them more easily readable by the learners at this stage in the exercises.

### DIFF
--- a/exercises/concept/poetry-club-door-policy/.meta/exemplar.js
+++ b/exercises/concept/poetry-club-door-policy/.meta/exemplar.js
@@ -27,7 +27,7 @@
  * @returns {string}
  */
 export function frontDoorResponse(blurb) {
-  return blurb[0];
+  return blurb.trim()[0];
 }
 
 /**
--- a/exercises/concept/poetry-club-door-policy/.meta/exemplar.js
+++ b/exercises/concept/poetry-club-door-policy/.meta/exemplar.js
@@ -27,7 +27,7 @@
  * @returns {string}
  */
 export function frontDoorResponse(blurb) {
-  return blurb.trim()[0];
+  return blurb[0];
 }
 
 /**
--- a/exercises/concept/poetry-club-door-policy/door-policy.spec.js
+++ b/exercises/concept/poetry-club-door-policy/door-policy.spec.js
@@ -1,3 +1,5 @@
+/** @format */
+
 // @ts-check
 
 import {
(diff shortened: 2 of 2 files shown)


exercises/concept/poetry-club-door-policy/door-policy.spec.js
```diff
@@ -7,239 +9,117 @@ import {
...
-      });
+    test.each([
+      { poem: CODE_WORK, lastLetters: 'work', password: 'Work, please' },
+      {
+        poem: SHIRE_HORSE_WITH_SPACES,
+        lastLetters: 'horse',
+        password: 'Horse, please',
+      },
+    ])('should be correct for $password', ({ poem, lastLetters, password }) => {
+      expect(recite(poem, backDoorPatron).join('')).toBe(lastLetters);
```
junedev, line 121: Because of how this test works with the join, all tests will pass even if the student does not use `trim` in `backDoorResponse`. The tests should catch the missing `trim`, e.g. by also running `SHIRE_HORSE_WITH_SPACES` through the "output a character per line" test.

exercises/concept/poetry-club-door-policy/door-policy.spec.js
```diff
@@ -7,239 +9,117 @@ import {
...
-
-const SHIRE_HORSE_WITH_SPACES = new Poem(
-  `
-Stands so high  
-Huge hooves too      
-Impatiently waits for 
-Reins and harness  
+//Shire Horse, by Michael Lockwood
+const SHIRE_HORSE_WITH_SPACES = `
+Stands so high
```
junedev, line 29: Since the variable name includes "spaces" and this test case should help the student figure out they need to trim, I would suggest adding some obvious spaces like shown below (instead of just new line whitespaces).
```suggestion
     Stands so high
```

exercises/concept/poetry-club-door-policy/door-policy.spec.js
```diff
@@ -7,239 +9,117 @@ import {
...
+  .split('\n');
 
 describe('strings', () => {
-  describe('front door', () => {
-    const ShireGuard = new FrontDoorGuard(SHIRE_HORSE);
-
-    test('it outputs a character per line', () => {
-      const key = ShireGuard.recite();
+  describe('front door response', () => {
+    test('should output a character per line', () => {
```
junedev, line 70: General comment about all the test cases:  
Our test runner is not super intelligent yet so to get good output on the website, each test should include all the data it needs to run and be stand-alone. No table driven tests, no variables declared outside of the actual `test(...)` function (even if that means some duplication). You can see an example here that also has some longer data included: https://github.com/exercism/javascript/blob/304ebb8ba78283d09efce863ab104d02f7546476/exercises/concept/lasagna-master/lasagna-master.spec.js#L31-L77

Also we will connect the test cases to the tasks in the future (the one "intermediate" task the instructions currently have will be removed). So ideally there should be one `describe` block per imported function.

If you don't have time to fix those general things, I can also do it later as part of my cleanup ticket as this PR is definitely already a step in the right direction even without those fixes.

exercises/concept/poetry-club-door-policy/door-policy.spec.js
```diff
@@ -7,239 +9,117 @@ import {
...
 Exit with zero quick
-
-Derk-Jan Karrenbeld
-`.trim()
-);
+`
+  .trim()
+  .split('\n');
 
 describe('strings', () => {
```
junedev, line 68: this outer describe block does not add value and can be removed

junedev (commented):
Sorry for the wait, I provided the review now.

exercises/concept/poetry-club-door-policy/door-policy.spec.js
SleeplessByte: (it's coming. WIP here: https://github.com/exercism/typescript-test-runner/pull/27)

exercises/concept/poetry-club-door-policy/door-policy.spec.js
```diff
@@ -7,239 +9,117 @@ import {
...
-      test(`frontDoorPassword(${name})`, () => {
-        guard.assert();
-      });
-    });
+    test.each([
+      { poem: SUMMER, firstLetters: 'SUMMER', password: 'Summer' },
+      { poem: SOPHIA, firstLetters: 'SOPHIA', password: 'Sophia' },
+      { poem: CODE_WORK, firstLetters: 'CODE', password: 'Code' },
+    ])(
+      'should be correct for $password',
```
junedev, line 90: `$password` as text does not seem very descriptive, maybe add a name property to the test objects above that contains the name of the poem, that you can then use here

SleeplessByte: `test.each` _should_ work well but it doesn't. Can you make them three separate tests? Sorry! Our test-runner isn't great yet.

exercises/concept/poetry-club-door-policy/door-policy.spec.js
mmmmmrob: Restructured the tests to lose the `join()` and added a test using `SHIRE_HORSE_WITH_SPACES`.

exercises/concept/poetry-club-door-policy/door-policy.spec.js
mmmmmrob: Restructured the poems to make them simpler and whitespace more obvious.

exercises/concept/poetry-club-door-policy/door-policy.spec.js
mmmmmrob: One describe block per function, and removed `.each` blocks.

exercises/concept/poetry-club-door-policy/door-policy.spec.js
mmmmmrob: `$password` should have been outputting the password from the each block, but wasn't in this version of jest. Tests now split out.

exercises/concept/poetry-club-door-policy/.meta/exemplar.js
```diff
@@ -27,7 +27,7 @@
  * @returns {string}
  */
 export function frontDoorResponse(blurb) {
-  return blurb[0];
+  return blurb.trim()[0];
```
junedev, line 30: Oh no, I just realized the instructions only say that backdoor response trims at the end. This line does not fit the instructions. It needs to stay as it was. Instead the "with spaces" test should only have additional spaces at end. Sorry for getting you on the wrong track there with my "with spaces" example.

exercises/concept/poetry-club-door-policy/door-policy.spec.js
```diff
@@ -1,245 +1,142 @@
...
+  'Exit with zero quick',
+];
+
+describe('front door response', () => {
+  test('should output a character per line', () => {
+    const expectedLetters = recite(SHIRE_HORSE, frontDoorResponse);
+    expect(expectedLetters.length).toBe(SHIRE_HORSE.length);
+  });
+
+  xtest('should take the first character of each line', () => {
```
junedev, line 67: For concept exercises all tests should be unskipped as they are not meant to be done TDD style.

junedev (approved):

PR #1370, opened 2021-09-24
Use consistent "skip test" label

joshgoebel:
This seems to consistently fail in the online test runner... (but it's not skipped)... is there a reason for this discrepancy or is this a bug?
Screen Shot 2021-09-23 at 10 47 33 PM


SleeplessByte:
Yes, this is on purpose. `test.skip` tests are supposed to be skipped because they can't run online.

The typescript-test-runner was recently updated to actually not fail skipped tests, but the javascript-test-runner isn't updated yet.

PR #1379, opened 2021-09-25
Doc Correnction in Elyses Enchantments Problem Instructions

dhruvsharma1999:
In Javascript Exercise, the array's problem i.e Elyses Enchantments. The 5th point of the instructions explains the Removing of a card from the stack given.
Make a card disappear by removing the card the given `position` from the stack. Return the adjusted stack.

`const position = 2;
removeItem([3, 2, 6, 4, 8], position);
// => [3, 2, 4, 8]`

Th correct return of this example should be //=> [3,2,6,8], as the position from which the element needs to be removed is 2nd and not the third.

### DIFF
--- a/exercises/concept/elyses-enchantments/.docs/instructions.md
+++ b/exercises/concept/elyses-enchantments/.docs/instructions.md
@@ -56,7 +56,7 @@ Return the adjusted stack.
 ```javascript
 const position = 2;
 removeItem([3, 2, 6, 4, 8], position);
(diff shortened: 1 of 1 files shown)


junedev:
Position refers to the index of the element in the array, so index 2 is indeed the third element as the first element has position (index) 0. I agree that this is not clear from the instructions but it is already visible in the first example.

I will close this PR but feel free to open an issue or PR regarding improving the instructions.

PR #1380, opened 2021-09-25
Improve Concept Exercise Elyses Enchantments - Array

### LINKED ISSUE #1374: Improve Concept Exercise: Elyses Enchantments (`arrays`)
Related: https://github.com/exercism/support/issues/28#issuecomment-926606144

This is a tiny task. The stubbed JSDoc can be improved. 

## Current JSDoc

```javascript
/**
 * Remove the card at the 0-based position
 * 
 ...
 */
```

## Proposal

Change this JSDoc to make it clear that the card at the _given_ position must be removed. Perhaps we say: `position 0 is the first position` or something like that. 0-based and 1-based indexation isn't a concept that's necessarily clear, so making it more easy to discover might help future students.

pertrai1:
resolves #1374

### DIFF
--- a/exercises/concept/elyses-enchantments/.docs/instructions.md
+++ b/exercises/concept/elyses-enchantments/.docs/instructions.md
@@ -2,11 +2,11 @@
 
 As a magician-to-be, Elyse needs to practice some basics. She has a stack of cards that she wants to manipulate.
 
-To make things a bit easier she only uses the cards 1 to 10.
+To make things a bit easier she only uses the cards 1 to 10. Remember that array indices start at 0.
 
 ## 1. Retrieve a card from a stack
 
-In order to pick a card, return the card at position `position` from the given stack.
+In order to pick a card, return the card at index `position` from the given stack.
 
 ```javascript
 const position = 2;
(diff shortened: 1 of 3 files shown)


exercises/concept/elyses-enchantments/.docs/instructions.md
```diff
@@ -2,11 +2,11 @@
 
 As a magician-to-be, Elyse needs to practice some basics. She has a stack of cards that she wants to manipulate.
 
-To make things a bit easier she only uses the cards 1 to 10.
+To make things a bit easier she only uses the cards 1 to 10. Remember that array indices start at 0.
```
junedev, line 5: This sentence feels a bit out of place, I would suggest the following alternative:
```suggestion
To make things a bit easier she only uses the cards 1 to 10 so her stack of cards can be represented by an array of numbers. The position of a certain card corresponds to the index in the array. That means position 0 refers to first card, position 1 to the second card etc.
```

pertrai1: Thank you for the feedback. I have updated based on your suggestion.

PR #1381, opened 2021-09-25
Add a test to check that only the last result is remembered

### LINKED ISSUE #1373: Improve Concept Exercise: Coordinate Transformation (`closures`)
The goal of this task is to limit the number of different solutions are valid for the Concept Exercise "Coordinate Transformation".

As a refresher, or first time read, the docs can be found here:

Currently, the final task asks to write a `memoize` function, that only memoizes the _last_ result.
However, the tests don't enforce that final limitation. Technically memoizing _all_ calls is fine if you look at the grammar pedantically (it's undefined), but the many many ways to build a cache make it harder to write automated analyzers for, and we really want to force people to use `let` (or `var`) here.

## Solution that should not pass (but passes)

```javascript
export function memoizeTransform(f) {
  const cache = {}

  return function memoize(x, y) {
    const key = `${x}:${y}`
    
    return cache[key] || cache[key] = f(x, y)
  }
}
```

## Solution that should pass (and passes)

```javascript
export function memoizeTransform(f) {
  let lastX, lastY, lastResult

  return function memoize(x, y) {
    if (x === lastX && y === lastY) {
      return lastResult
    }
    
    lastX = x
    lastY = y
    
    return lastResult = f(x, y)
  }
}
```

samfundev:
I added a test that makes sure the student's code isn't remembering all the results and only the last one. It calls the student's function with some input, different input, and then the first input again. This should require the underlying function to be called 3 times if the student's code is implemented to only remember the last result.

Fixes #1373.

### DIFF
--- a/exercises/concept/coordinate-transformation/coordinate-transformation.spec.js
+++ b/exercises/concept/coordinate-transformation/coordinate-transformation.spec.js
@@ -119,4 +119,13 @@ describe('memoizeTransform', () => {
     expect(memoizedTransform(5, 5)).toEqual([1, 1]);
     expect(memoizedTransform(5, 5)).toEqual([1, 1]);
   });
+
+  test('should only remember the last result', () => {
+    const mockFunction = jest.fn((x, y) => [x * 2, y * 2]);
+    const memoizedTransform = memoizeTransform(mockFunction);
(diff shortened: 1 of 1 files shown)


SleeplessByte:
Looks good! Thanks.

PR #1388, opened 2021-09-28
Add concept 'array-transformations' and improve the concept's exercise

### LINKED ISSUE #1326: Improve Concept Exercise: Elyses Transformative Enchantments (array-transformations)
The goal of this task is to make sure the Concept Exercise "Elyses Transformative Enchantments" conform to the latest specification. As part of that, the concept `array-transformations` has to be created.

The docs can be found here

Additionally you can look at the `bird-watcher` concept exercise for reference.

The improvement should include changing the introduction.md file of the exercise to contain some minimal information so that the student is able to solve the exercise. For now it is ok to use that same content as introduction.md and about.md for the Concept as well.

Also ensure that
* there are good blurbs for the concept and exercise
* the tests are declarative (see https://github.com/exercism/javascript/issues/1019 for an example)
* each test case is "self-contained" so that the student sees meaningful output from the test runner (no forEach, no helper functions defined in other parts of the test file etc.)

Here is a general idea for the introduction that was discussed before:
Show how to do something like map or filter with a simple for loop, then show how to do it with the helper method instead. Afterwards provide a table with the common helpers. Just the method name as a link (e.g. to MDN) and a short explanation what this is for. We think its ok if the student reads up on some details of these methods on their own, we don't want to show and explain every single one in detail. Also the list can contain more things, than needed for the exercise (we count that info as "necessary to understand the concept" in contrast to "necessary to solve the exercise").

It might make sense to replace the `fill` task of the exercise (the last one) with `reverse` instead as that is probably more commonly used.

TomPradat:
Fix #1326


TomPradat:
@junedev @SleeplessByte Before I continue this PR, am I getting the concept right?

concepts/array-transformations/introduction.md
```diff
@@ -0,0 +1,66 @@
+# Introduction
+
+In javascript, it is very common to work with arrays. That explains why there are a lot of builtin methods to easy this operations. Some of them are used to transform the array into another array, a number, an object etc... We'll call these operations "transformations".
```
pertrai1, line 3: Seems to me that this should read differently:

> That explains why there are a lot of builtin methods to easy this operations

Just not sure how :-)

concepts/array-transformations/introduction.md
```diff
@@ -0,0 +1,66 @@
...
+}
+
+arr.filter(isOdd); // [2, 4]
+```
+
+This function is also pure.
+
+- `reduce`:
+
+Using an "accumulator", reduces the array in a single value by applying for each element of the array a function that takes as parameters the accumulator and the current element and returns the new accumulator.
```
pertrai1, line 43: do you think that there should be a different way of saying how each element is operated on? I found myself getting tripped up when I saw `for each` :-)

> applying for each element of the array

TomPradat: Good point :+1:. I took this opportunity to rewrite the sentence, what do you think?

```suggestion
Reduces the array to a single value using a function that takes as parameters an "accumulator" and the current element of the array. This function instructs how the current element must be merged into the accumulator and returns the accumulator that will be used on the next iteration.
```

pertrai1: That is easier to read and follow along with. Thanks for the update.

junedev:
@TomPradat I wrote in the issue description that I would recommend replacing fill with reverse. So the concept should introduce reverse instead of fill. 

Since there are a lot of methods in the exemplar, you can also introduce some of the easier methods just with a list. You can see an example here: https://github.com/exercism/javascript/blob/main/concepts/strings/introduction.md#concatenation-and-methods  
You don't have to provide the details/examples for all the methods from the exemplar, I would focus on the more difficult ones like reduce.

TomPradat:
Time flies by but I haven't forgotten this PR, I'll work on this soon hopefully :). By the way, I'll probably scope this PR to **adding the new concept only** in order to make this mergeable more quickly

junedev:
@TomPradat It seems I missed your question re the method list. Please proceed as you like which regards to that.

concepts/array-transformations/introduction.md
```diff
@@ -0,0 +1,114 @@
+# Introduction
```
junedev, line 1: The content of this file should be copied into about.md as well.

junedev:
@TomPradat I finished this up so it can be merged. Hope that's ok. It still misses `sort` I think but that can be added when working on the exercise part.

TomPradat:
> @TomPradat I finished up the concept so it can be merged first. Hope that's ok. It still misses `sort` I think but that can be added when working on the exercise part.

That's a good idea, thank you for the hand :+1:

PR #1396, opened 2021-10-03
Fixed two-bucket solution & errant tests (#905)

### LINKED ISSUE #901: two-bucket: fix implemenation

qianzhong516:
Hi, I have fixed the two-bucket solution and reverted all the `test.skip` to `xtest` in the spec.js file. Please check and let me know further issues, thanks!

Fixes #901

### DIFF
--- a/exercises/practice/two-bucket/.meta/proof.ci.js
+++ b/exercises/practice/two-bucket/.meta/proof.ci.js
@@ -95,10 +95,28 @@ export class TwoBucket {
     return mvCount;
   }
 
+  gcd(a, b) {
+    // greatest common divisor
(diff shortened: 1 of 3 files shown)


SleeplessByte:
Sweet! Thanks Janice!

PR #1397, opened 2021-10-03
Added Promises content in introduction & about

### LINKED ISSUE #1384: Write about and introduction for Promises concept
The about.md from concepts/promises contains a single line TODO. The introduction.md contains a single sentence.

Both need to be properly filled out with information about promises.

ProProgrammer2504:
Fixes #1384

### DIFF
--- a/concepts/promises/about.md
+++ b/concepts/promises/about.md
@@ -1,11 +1,11 @@
 # About
 
-The [Promise][promise-docs] object represents the eventual completion (or failure) of an asynchronous operation and its resulting value. It allows you to associate handlers with an asynchronous action's eventual success value or failure reason.
+The Promise object represents the eventual completion (or failure) of an asynchronous operation and its resulting value. It allows you to associate handlers with an asynchronous action's eventual success value or failure reason.
 
 A Promise is in one of these states:
 
 - **pending**: initial state, neither fulfilled nor rejected.
-- **fulfilled**: meaning that the operation was completed successfully.
+- **fulfilled**: meaning that the operation was completed successfully and the result is available.
 - **rejected**: meaning that the operation failed.
 
 When either of these options happens, the associated handlers queued up by a promise's `then` method is called. If the promise has already been fulfilled or rejected when a corresponding handler is attached, the handler will be called, so there is no race condition between an asynchronous operation completing and its handlers being attached.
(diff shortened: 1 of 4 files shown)


concepts/promises/about.md
```diff
@@ -1,3 +1,146 @@
 # About
 
-TODO: add information on promises concept
+The [Promise][promise-docs] object represents the eventual completion (or failure) of an asynchronous operation and its resulting value. It allows you to associate handlers with an asynchronous action's eventual success value or failure reason.
+
+A Promise is in one of these states:
+
+- **pending**: initial state, neither fulfilled nor rejected.
+- **fulfilled**: meaning that the operation was completed successfully.
```
junedev, line 8:
```suggestion
- **fulfilled**: meaning that the operation was completed successfully and the result is available.
```

concepts/promises/about.md
```diff
@@ -1,3 +1,146 @@
 # About
 
-TODO: add information on promises concept
+The [Promise][promise-docs] object represents the eventual completion (or failure) of an asynchronous operation and its resulting value. It allows you to associate handlers with an asynchronous action's eventual success value or failure reason.
```
junedev, line 3: Links can be distracting so they should be used sparingly and only if they add additional value compared to the existing content. E.g. the link to the promise docs above could be placed in the links.json instead (if its not included already). It will then render below the content of this file in the "Read More" section.

junedev (commented):
Thanks a lot for working on this. I left a couple of minor comments.

concepts/promises/about.md
```diff
@@ -1,3 +1,146 @@
...
+
+A Promise is in one of these states:
+
+- **pending**: initial state, neither fulfilled nor rejected.
+- **fulfilled**: meaning that the operation was completed successfully.
+- **rejected**: meaning that the operation failed.
+
+When either of these options happens, the associated handlers queued up by a promise's `then` method is called. If the promise has already been fulfilled or rejected when a corresponding handler is attached, the handler will be called, so there is no race condition between an asynchronous operation completing and its handlers being attached.
+
+The methods [`promise.then()`][promise-then], [`promise.catch()`][promise-catch], and [`promise.finally()`][promise-finally] are used to associate further action with a promise that becomes settled.
```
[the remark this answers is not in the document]
ProProgrammer2504: Should I also delete line number 11?

concepts/promises/about.md
ProProgrammer2504: There is already a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises
Should I also add another link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

concepts/promises/about.md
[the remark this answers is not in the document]
junedev: No, I think it is ok to keep it as general explanation.

concepts/promises/about.md
junedev: Yes, I think they are different enough so that makes sense.

ProProgrammer2504:
Anything more to add/change?

junedev (approved):

junedev:
Well, there are probably always minor things that could be discussed/rephrased/changed. But the main thing I look for is that the proposed change is an improvement to what was there before (which this definitely is), that it is applied consistently across the different files (like with the exercise introduction) and whether there is something major that needs to be fixed the the headlines in this case.  
Then the PR gets merged and of someone feels like improving the content further in the future there can be a new PR. This works better than turning around each sentence a couple of times.

Issue #898, opened 2020-12-18
palindrome-products: optimise implementation


SleeplessByte:
Don't work on this issue until #895 is merged, or use that PR as a base branch.

The "largest palindrome from four digit factors" test currently takes > 30s to complete. When fixed, replace `test.skip` with `xtest`.

tliebert:
This test case currently prevents the users code from passing the test suite - it returns this case as failed even though the test is skipped.

junedev:
@tliebert Thanks for reporting. We are aware of the issue. @SleeplessByte is working on fixing this.

PR #1398, opened 2021-10-04
Optimised palindrome-products solution

qianzhong516:
Hi, the original solution takes around 30s to pass all the tests on my machine. I have greatly optimized the solution and it only takes 1.5s to pass all the tests now.

### DIFF
--- a/exercises/practice/palindrome-products/.meta/proof.ci.js
+++ b/exercises/practice/palindrome-products/.meta/proof.ci.js
@@ -33,23 +33,39 @@ export class Palindromes {
   }
 
   get largest() {
-    let best = new Palindrome(this.minFactor, this.minFactor);
-    for (let m = this.maxFactor; m >= this.minFactor; m -= 1) {
-      let p = null;
-      for (let n = m; n >= this.minFactor && (!p || !p.valid()); n -= 1) {
-        p = new Palindrome(m, n);
-        if (p.valid()) {
-          if (best.value < p.value) {
-            best = p;
-          } else if (best.value === p.value) {
-            best = p.merge(best);
-          }
+    let left = this.maxFactor,
+      right = this.maxFactor,
+      best = new Palindrome(this.minFactor, this.minFactor);
+
+    while (right >= this.minFactor) {
+      let p = new Palindrome(left, right);
+
+      if (best.value && p.value < best.value) {
+        right--;
+        left = right;
+        continue;
+      }
+
+      if (p.valid()) {
+        if (best.value < p.value) {
+          best = p;
+        } else if (best.value === p.value) {
+          best = p.merge(best);
         }
       }
+
+      if (left - 1 < this.minFactor) {
+        right--;
+        left = right;
+      } else {
+        left--;
+      }
     }
+
     if (best.valid()) {
       return best;
     }
+
     return { value: null, factors: [] };
   }
 
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -133,6 +133,11 @@
   description: "Work on Documentation"
   color: "3c2d9f"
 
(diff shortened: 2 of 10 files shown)


SleeplessByte:
(@junedev added you for your eyes. @joshgoebel I know you find these really interesting)

surya-11dev (commented):
No its not working. still cant pass the largest number.

SleeplessByte:
@surya-11dev, that's a different issue. That's related to #898, and the test-runner trying to run the skipped test.

surya-11dev:
> @surya-11dev, that's a different issue. That's related to #898, and the test-runner trying to run the skipped test.

Thank u very much sir for making time to explain things.

surya-11dev:
> How so? Passes ALL tests here locally.

sorry sir, nothing but some misunderstanding.

joshgoebel:
@SleeplessByte Thanks for pinging me, I was mentoring this student so I had seen this early!

junedev:
@joshgoebel @SleeplessByte Could you two decide whether this can be merged please? I am not very familiar with this exercise.

joshgoebel (approved):
Has my vote.

exercises/practice/palindrome-products/palindrome-products.spec.js
```diff
@@ -79,7 +79,7 @@ describe('Palindromes', () => {
     expect(sortFactors(smallest.factors)).toEqual(expected.factors);
   });
 
-  test.skip('largest palindrome from four digit factors', () => {
+  xtest('largest palindrome from four digit factors', () => {
```
joshgoebel, line 82: Pretty sure this was intentional.

exercises/practice/palindrome-products/.meta/proof.ci.js
```diff
@@ -70,14 +68,22 @@ export class Palindromes {
   }
 
   get smallest() {
-    for (let m = this.minFactor; m <= this.maxFactor; m += 1) {
```
joshgoebel, line 57: _Isn't this the same exact algorithm just written in a much more complex manner?_  Isn't the loop still covering exactly the same territory?  Since you're just finishing the first item you can't get the same benefits you could by optimization the max loop.

exercises/practice/palindrome-products/palindrome-products.spec.js
qianzhong516: I think the previous solution caused some error when ran into this test case, so the maintainer skipped this test by `test.skip`. After the changes, it should work out fine, so I reverted `test.skip` to `xtest`.

exercises/practice/palindrome-products/.meta/proof.ci.js
qianzhong516: Please see #1407. I have updated the code after our discussion on exercism.io

exercises/practice/palindrome-products/palindrome-products.spec.js
joshgoebel: This is on purpose the test runner on the website is not supposed to run this because it can take too long.  There is a bug now where instead it fails.  That's a separate issue.

exercises/practice/palindrome-products/.meta/proof.ci.js
joshgoebel: Can't this be resolved by simply having the second for loop start with `m` instead of `minFactor`, avoiding looping over the same territory as prior passes?

joshgoebel: Adding complexity solely for symmetry is not a win IMHO.  If the simpler for loop solution work for minimum, lets not complicate it.

PR #1403, opened 2021-10-05
Tests for empty promise array and no arguments.

### LINKED ISSUE #1400: Promises Exercise may be missing tests
https://github.com/exercism/javascript/blob/ba702157951d269f3ec70d968bd8573efd3a2ade/exercises/practice/promises/promises.spec.js

The tests for the promises exercise ensures each function returns a promise, but does not test for two scenarios:

1. What should happen if the function is called with no arguments (or nullish) rather than empty array

2. That the promise resolves when called with an empty promises array.

I spotted these because my own implementations pass all the tests but do not resolve when given an empty array.

I am happy to add tests if we agree they would add value.

mmmmmrob:
Without these tests it's possible, easy even, to write promises that never resolve or reject when passed an empty array or no array.

fixes #1400

### DIFF
--- a/exercises/practice/promises/.meta/proof.ci.js
+++ b/exercises/practice/promises/.meta/proof.ci.js
@@ -5,30 +5,42 @@ export const promisify =
       fn(...args, (err, result) => (err ? reject(err) : resolve(result)));
     });
 
-export const all = (promises) =>
-  promises.reduce(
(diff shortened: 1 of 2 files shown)


tejasbubane (approved):

tejasbubane:
Thanks @mmmmmrob :+1:

PR #1413, opened 2021-10-07
Fixed fruit-picker test case (#1394)

### LINKED ISSUE #1394: Discrepancy between docs and test case
The comment on `purchaseInventoryIfAvailable` in fruit-picker.js says arg should be bool but a test case expects passing an object to return "PURCHASE" as if this is an expected type.

https://github.com/exercism/javascript/blob/669ddc8cc78c7b9e25c6ec350dd87d014c96bfcd/exercises/concept/fruit-picker/fruit-picker.js#L38

https://github.com/exercism/javascript/blob/669ddc8cc78c7b9e25c6ec350dd87d014c96bfcd/exercises/concept/fruit-picker/fruit-picker.spec.js#L74

The test case should be replaced by a bool or the comment should be updated to show different types are expected.

qianzhong516:
Fixes #1394

### DIFF
--- a/exercises/concept/fruit-picker/.docs/instructions.md
+++ b/exercises/concept/fruit-picker/.docs/instructions.md
@@ -20,7 +20,7 @@ The grocer's API provides a function to query their inventory called `checkInven
 
 The query takes the form of an _object_:
 
-```typescript
+```javascript
 const query = {
   variety: string,
   quantity: number,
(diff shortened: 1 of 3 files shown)


junedev:
I added some more fixes (hope that is ok):
* moved callback into the actual function (otherwise it does not show up in the website test runner output)
* found another place where `isAvailable` was an object instead of a boolean
* removed the unnecessary `setResesponse` calls
* change the JSDoc so the second argument of `purchaseInventoryIfAvailable` can be null as described in the instructions
* removed `typescript` tag from code block

Issue #1049, opened 2021-03-18
Improve Concept Exercise: Numbers (Freelancer Rates)


junedev:
## Getting started
Here you can read about what Concept Exercises are and how they are structured:

- [What are Concept Exercises?](https://exercism.org/docs/building/product/concept-exercises)
- [Concept Exercises](https://exercism.org/docs/building/tracks/concept-exercises)
- [Concepts](https://exercism.org/docs/building/tracks/concepts)

If you have not done so yet, it is probably also helpful to do a couple of "Learning Exercises" (this is how they are called on the site) yourself.

## Goal

The goal of this issue is to extend the existing concept exercise [Freelancer Rates](https://github.com/exercism/javascript/tree/main/exercises/concept/freelancer-rates) that teaches the `numbers` concept.

The exercise is already in good shape but currently only explicitly introduces numbers as a concept. **Additionally it should also introduce arithmetic operators incl. shorthand assignment operators.**

See [this list](https://github.com/exercism/javascript/issues/951#issuecomment-778834953) for details on the learning curve we are aiming for.

## Additional Learning Objectives
- How to do perform the following 6 arithmetic operations: addition, subtraction, multiplication, division, remainder, exponentiation
- What are shorthand assignment operators for those 6 operations and how to use them

(To avoid too many tiny concepts, both of these will be combined in one additional concept called `arithmetic-operators`.)

### Add Concept
- [x] Create a new folder  `arithmetic-operators` in `concepts` and add the concept to the general config.json file (incl. adding it to `concepts` for the `freelancer-rates` exercise there)
- [x] Write the introduction.md, about.md and links.json file for the concept

### Adjust the Concept Exercise
- [x] Add the introduction.md of the concept at the end of the introduction.md file of the exercise, make some adjustments in the content if necessary.
- [x] Adjust the design.md file to incorporate the new learning objectives.
- [x] Update the tasks in the exercise to include at least one use of one of the more unusual operators (`%` or `**`) and one sensible use of a shorthand assignment operator. You can add new additional tasks but you can also consider replacing/adjusting an existing task that might currently not teach much more than the other tasks already do. Updating the tasks includes the following files
  - instructions.md
  - hints.md
  - exemplar.js
  - stub file (freelancer-rates.js)
  - test file (freelancer-rates.spec.js)

### Fixes for the Existing Content
This part is optional polishing "while you are at it" but can also be tackled in another issue/PR later.

- [x] Fix broken link in "Built-in Object" section
[...]

PR #1420, opened 2021-10-09
Created arithmetic operators concept, as per #1049

JaPrad:
Part of #1049


JaPatGitHub:
Thank you for reviewing @junedev. I've incorporated your improvement suggestions.

PR #1426, opened 2021-10-11
🚧  Add practice exercise book store

0xBradock:
Hello @SleeplessByte and @junedev,
Could you tell me if I'm in the right direction?

I think I'm only missing on `.meta/proof.ci.js`, which should be done soon.

Thanks


SleeplessByte:
image

PR #1440, opened 2021-10-15
start freelancer

JoseToitio:


SleeplessByte:
If anything, this should be applied everywhere (according to [the jest docs](https://jestjs.io/docs/cli#--cache) we likely don't need this), but given [Chesterton's fence](https://fs.blog/2020/03/chestertons-fence/), I'd rather keep it, as it was here from before I joined.

PR #1441, opened 2021-10-15
Add missing [concept-arrays] tag in Functions concept instruction file

bwtlt:
The instruction file for the *"Functions"* concept was missing a tag for the link to the *"Arrays"* concept page.

### DIFF
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -107,7 +107,7 @@ noReturn(1);
 ```
 
 In JavaScript, you can only return exactly one value.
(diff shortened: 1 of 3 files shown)


junedev:
Thanks for pointing this out. If you submit more of these PRs, please always check for the mistake you found in all these files:
- exercise introduction.md
- concept about.md
- concept introduction.md

They are very similar, so usually a mistake that was made in one of them also appears in the others.
I did this for you here in this case: 72de65f4544822b385ada2eb28d63214652af308

PR #1442, opened 2021-10-15
Fix typo in door policy comment

carinablair:
Update the `backDoorResponse` comment to say 'back door'.


junedev:
That was my copy and paste error, I think. :see_no_evil:  Thanks for the fix!

Issue #983, opened 2021-02-15
[v3] Add prerequisites 2/6


TomPradat:
> ⚠️ If you have issues / are unsure, create ONE PR PER EXERCISE so it's easier to merge them as we can have discussions per exercise. If you're certain, it's fine to do a single one.

The idea is to split #960 into smaller chunks.

Here is the list of practices exercises to deal with for this part :

~collatz-conjecture~

~triangle~

~clock~

~meetup~

~etl~

~hamming~

~raindrops~

~nucleotide-count~

scrabble-score

allergies

word-count

bank-account

~difference-of-squares~

perfect-numbers

~complex-numbers~

luhn

~prime-factors~

grains

pythagorean-triplet

palindrome-products

TomPradat:
Update, here is the list of exercises that still need to be addressed :

scrabble-score

allergies

word-count

bank-account

perfect-numbers

luhn

grains

pythagorean-triplet

palindrome-products

SleeplessByte:
That's it I guess?!

PR #1468, opened 2021-10-26
Add prerequisites for practice exercises

TomPradat:
This PR is related to #983 , it will deal with the following exercises :

```
meetup

hamming

raindrops

nucleotide-count
```

In order to evaluate the prerequisites in the best way, I'm doing the exercise then I'm reviewing the exercise's proof of work and the submitted solutions.

### DIFF
--- a/config.json
+++ b/config.json
@@ -630,11 +630,10 @@
         "uuid": "be6f3ab3-1593-4c2d-8a35-5f39c1d5c91f",
         "practices": [],
         "prerequisites": [
-          "classes",
-          "array-transformation",
           "objects",
-          "strings",
-          "template-strings"
+          "template-strings",
+          "array-loops",
+          "errors"
         ],
         "difficulty": 4,
         "topics": [
--- a/config.json
+++ b/config.json
@@ -611,7 +611,7 @@
         "name": "Hamming",
         "uuid": "d773c4ef-c09e-40e4-a7fe-01456cb4a12a",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["comparison", "errors", "for-loops"],
         "difficulty": 2,
         "topics": ["conditionals", "loops", "equality", "strings"]
       },
--- a/config.json
+++ b/config.json
@@ -620,7 +620,7 @@
         "name": "Raindrops",
         "uuid": "f77ac2d1-cf3a-497d-bf04-b484a5a9cb37",
         "practices": [],
-        "prerequisites": ["objects", "array-transformations", "strings"],
+        "prerequisites": ["arithmetic-operators", "strings", "conditionals"],
         "difficulty": 2,
         "topics": ["conditionals", "integers", "strings", "transforming"]
       },
--- a/config.json
+++ b/config.json
@@ -629,7 +629,13 @@
         "name": "Nucleotide Count",
         "uuid": "be6f3ab3-1593-4c2d-8a35-5f39c1d5c91f",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "classes",
+          "array-transformation",
+          "objects",
+          "strings",
+          "template-strings"
+        ],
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -611,7 +611,7 @@
         "name": "Hamming",
         "uuid": "d773c4ef-c09e-40e4-a7fe-01456cb4a12a",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["comparison", "errors", "for-loops"],
```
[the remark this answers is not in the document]
junedev: I am a strong believer that a simple for loop is actually a very good way of solving this exercise, so I totally agree here. Later we could even set this exercise as "practices for-loops".

config.json
```diff
@@ -620,7 +620,7 @@
         "name": "Raindrops",
         "uuid": "f77ac2d1-cf3a-497d-bf04-b484a5a9cb37",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["objects", "array-transformations", "strings"],
```
junedev, line 623: I would set the prerequisites here looking at these types of solutions: https://github.com/exercism/javascript/blob/main/exercises/practice/raindrops/.meta/proof.ci.js
People can always use more fancy stuff if they want what with what is in there you can write a good solution for the exercise and there are far better exercises to use array transformation concepts on than this one.
Also the old "topics" seem to go into that direction.

```suggestion
        "prerequisites": ["arithmetic-operators", "strings", "conditionals"],
```

config.json
```diff
@@ -629,7 +629,13 @@
...
         "uuid": "be6f3ab3-1593-4c2d-8a35-5f39c1d5c91f",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "classes",
+          "array-transformation",
+          "objects",
+          "strings",
+          "template-strings"
+        ],
```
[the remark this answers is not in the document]
junedev: This exercise is a tough one to define. The proof file solution is not very good imo and the community solutions are full of awful solutions as well.

I digged around a bit and I would use something like this as guidance for the prerequisites: https://exercism.org/tracks/javascript/exercises/nucleotide-count/solutions/shybyte (with the change that instead of the spread operator (`...`) there could be a simple `split`).

That would leave us with classes, objects, strings, array-loops, conditionals and errors as prerequisites. What do you think about that?

junedev (commented):
Left my thoughts on the exercises in the PR. The issue description also mentions "meetup". Not sure what's up with that. :thinking:

config.json
```diff
@@ -629,7 +629,13 @@
         "name": "Nucleotide Count",
         "uuid": "be6f3ab3-1593-4c2d-8a35-5f39c1d5c91f",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "classes",
```
junedev, line 633: Update: "classes" can be removed here, the exercise was changed to remove the unnecessary class definition (https://github.com/exercism/javascript/pull/1488)

TomPradat: I meant to create an issue about this, good to know it has been fixed :)

config.json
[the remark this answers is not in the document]
TomPradat: Agreed, I've done something similar myself: https://exercism.org/tracks/javascript/exercises/nucleotide-count/solutions/TomPradat

I'm wondering about conditionals though, it seems to me that 4 prerequisites are long enough. What's more, conditionals should be known before learning about objects. What do you think?

config.json
TomPradat: I get what you're saying, let's make the prerequisites of this one easier :+1:

TomPradat:
> Left my thoughts on the exercises in the PR. The issue description also mentions "meetup". Not sure what's up with that. thinking

Oops, looks like I forgot to push this one's commit. It's done now :)

junedev (approved):

PR #1470, opened 2021-10-27
Update hamming exercise to the latest specs

### LINKED ISSUE #1467: Practice exercise 'hamming' : Confusing tests
IMO the following tests are confusing :

```javascript
  test('empty strands', () => {
    expect(compute('', '')).toEqual(0);
  });

  xtest('disallow left empty strand', () => {
    expect(() => compute('', 'G')).toThrow(
      new Error('left strand must not be empty')
    );
  });

  xtest('disallow right empty strand', () => {
    expect(() => compute('G', '')).toThrow(
      new Error('right strand must not be empty')
    );
  });
```
Why should we throw if a strand is empty but not if both are?

I'd suggest that an empty string should not cause an error to be thrown. Thus the 2 last test cases should be removed.

I can make a PR if you agree with me

TomPradat:
Fixes #1467

### DIFF
--- a/exercises/practice/hamming/.meta/proof.ci.js
+++ b/exercises/practice/hamming/.meta/proof.ci.js
@@ -1,8 +1,5 @@
 export const compute = (strand1, strand2) => {
-  const len1 = strand1.length;
-  const len2 = strand2.length;
-
-  if (len1 !== len2) {
+  if (strand1.length !== strand2.length) {
     throw new Error('strands must be of equal length');
   }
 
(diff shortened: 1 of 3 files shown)


exercises/practice/hamming/.meta/proof.ci.js
```diff
@@ -2,16 +2,8 @@ export const compute = (strand1, strand2) => {
   const len1 = strand1.length;
   const len2 = strand2.length;
```
junedev, line 3: Now that len1 and len2 are only used once, I think declaring these extra variables is not necessary anymore. The new line 5 could use `.length` directly.

junedev (approved):
Looks all good now! Just one minor, optional comment.

exercises/practice/hamming/.meta/proof.ci.js
junedev: I did this change myself so this can be merged.

TomPradat: Yes, sorry I just saw that, thanks :ok_hand:

PR #1475, opened 2021-10-30
Proposed enhancement - add eslint ignore rest siblings

SMCRunner88:
This exercise teaches array destructuring using rest and spread. The unused variable warnings may get confusing. 

Screen Shot 2021-10-30 at 11 10 08 AM
Screen Shot 2021-10-30 at 11 09 52 AM


junedev:
@SMCRunner88 The error message is valid and you can avoid it with code like this: `const [, second] = deck;`

Sidenote: The `.eslintrc` file has a "template" on the root level of the repository. From there, it is synced to every exercise. Changes should only be made to this root level file.

SMCRunner88:
Ahh, good to know! Thanks.

SleeplessByte:
This is something we can also add as analyzer commentary.

PR #1476, opened 2021-10-30
Add Concept Exercise: Windowing System (Classes)

### LINKED ISSUE #1015: Implement new Concept Exercise: Classes
## Getting started
Here you can read about what Concept Exercises are and how they are structured:

If you have not done so yet, it is probably also helpful to do a couple of "Learning Exercises" (this is how they are called on the site) yourself. You can also look at the code of an existing concept exercise like [`bird-watcher`](https://github.com/exercism/javascript/tree/main/exercises/concept/bird-watcher) (concept [`for-loops`](https://github.com/exercism/javascript/tree/main/concepts/for-loops)) for reference.

## Goal
The goal of this exercise is to introduce prototypes and classes.

## Concepts
The following concept pages need to be created. You can combine the introduction.md files from the concepts and use that as introduction.md for the concept exercise. No need to create different content at this point.

* `classes`

## Learning Objectives
In the concepts the student should learn about the following topics and then practice them in the concept exercise.

* Simple intro to OOP (e.g. "building objects that encapsulate values, called instance properties, and functions that work on these values, these functions are called methods") and why it is useful (structure code, re-use etc.)
* Javascript is a "prototype-based language", what does that mean?  
  The explanation from this [MDN link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Details_of_the_Object_Model#class-based_vs._prototype-based_languages) might be helpful in phrasing this:
  > you define a constructor function to create objects with a particular initial set of properties and values. [...] You use the new operator with a constructor function to create a new object.
* How to a constructor function to create a "template object" with properties and methods using prototype syntax, e.g.
    ```javascript
    function Person (name) {
        this.name = name;
        this.canTalk = true;
    };

    Person.prototype.greet = function() {
        if (this.canTalk) {
            console.log('Hi, I am ' + this.name);
        }
    };
    ```
* What does `this` mean/do
* What is an instance and how to create one with `new`
* How to write the same code as above with class syntax as syntactic sugar (explain what that means)

## Out of Scope
* Inheritance, `extends`, `super` (will be taught later)

## About.md of the Concept
Here some ideas of what could be mentioned.

* Why was class syntax introduced to the language
* Pitfall: `this` problem when nesting functions, mitigation via self/that or arrow functions
* Explain the meaning and status of private properties and static methods in JS 

## Prerequisites
* `objects`
* `functions`

At this point the student only knows about objects as simple key-value maps. `this` was not introduced before. See [this list](https://github.com/exercism/javascript/issues/951#issuecomment-778834953) for details on the learning curve we are aiming for.

## Exercise Idea
[C# Elons Toys Exercise](https://github.com/exercism/csharp/blob/main/exercises/concept/elons-toys/.docs/instructions.md) could serve as template.

## Help

You can choose to do this solo-style, or collaborate with multiple people on this. The suggested approach is to

1. **First accept this issue by saying "I'd like to work on this"** (no need to wait for a response, just go with it) and optionally request that someone works with you (and wait for a second person to accept your request).
2. Use this issue to discuss any questions you have, what should be included in the content and what not and to collect reference material.
3. Create a PR and set "exercism/javascript" as reviewers. Additionally you can write in #maintaining-javascript that your PR is ready for review. Once you incorporated any critical feedback that the reviewer might give you and the PR is approved, it will be merged by a maintainer.

junedev:
Resolves #1015


concepts/classes/about.md
```diff
@@ -0,0 +1,104 @@
...
+function Car() {
+  // ...
+}
+
+const myCar = new Car();
+const yourCar = new Car();
+```
+
+It is important to note that in JavaScript, the instances and the constructor function keep a relationship to each other even after the instances were created.
+Every object instance includes a hidden, internal object referred to as `[[prototype]]` in the language specification. It holds a reference to the value of the `prototype` key of the constructor function. Yes, you read that correctly, a JavaScript function can have key/value pairs because it is also an object behind the scenes. Nowadays `[[prototype]]` can be accessed via the key `__proto__` in many environments. It is important to not confuse the prototype of an instance (`[[prototype]]`/`__proto__`) with the `prototype` property of the constructor function.
```
pertrai1, line 27: do you prefer `__proto__` over `getPrototypeOf`

concepts/classes/about.md
```diff
@@ -0,0 +1,104 @@
...
+}
+
+const myCar = new Car('red', '2mt');
+myCar.color;
+// => 'red'
+myCar.engineRunning;
+// => false
+```
+
+The syntax above the usual way of adding fields. In theory, instance fields that do not depend on the parameters of the constructor can also be added via the `prototype` entry of the constructor function.
```
pertrai1, line 49: add *is* between *above* and *the* - *above is the*

junedev:
@pertrai1 Please wait with the review until the PR is un-drafted/ready for review. I often rewrite the texts along the way.

pertrai1:
@junedev sounds good. Was not sure if early feedback/discussion helped any.

junedev:
@SleeplessByte The about.md is done now, it was tough :sweat_smile:, resources were similarly bad/fragmented as with type conversion. It would be great if you could review it. Please focus on whether something is wrong or something extremely important is missing (I know there is also more we could cover at some point). Also let me know in case you have written any good articles on the topic that I should link. I am neither an expert nor a fan of OPP, so please also check whether the terminology seems ok.

@pertrai1 You can of course also review about.md now if you like.

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -26,60 +26,136 @@ screenSize.height ║                 |      │                      │
...
+
+The constructor should have no parameters.
 
-- `title` : `String`, Initial value is "New Window"
-- `screenSize` : `Size`, constant value with `width` = 800 and `height` = 600
-- `size` : `Size`, initial value is the default value of the `Size` struct
-- `position` : `Position`, initial value is the default value of the `Position` struct
-- `contents` : `String?`, initial value is `nil`
+```javascript
+const window = new Window();
```
SleeplessByte, line 91: `Window` exists in the browser as class, so I recommend a slightly new name. `ProgramWindow` or `UIWindow` for example?

junedev: Oh yes, good point! Will change that when I continue working on the exercise.

concepts/classes/.meta/config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "blurb": "JavaScript includes the capabilities for object-oriented programming. While it nowadays has a \"class\" keyword, it is still a prototype-based language.",
```
SleeplessByte, line 2:
```suggestion
  "blurb": "JavaScript allows for object-oriented programming. Despite it having a \"class\" keyword, it is still a prototype-based language.",
```

Don't need to change this, but I would make it less wordy. This is an example :)

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
+# About
+
+JavaScript includes the capabilities for object-oriented programming ([OOP][wiki-oop]).
+In OOP, you want to create objects (_instances_) from "templates" (_classes_) so that they include certain data and functionality.
+The data properties are called _fields_ in the OOP context, the function properties are called _methods_.
+
+JavaScript did not have classes at all before they were added to the language specification in 2015.
```
SleeplessByte, line 7:
```suggestion
JavaScript did not have classes at all before they were added to the language specification in 2015, but allowed for object-oriented programming using prototype-based inheritance. 
```

Perhaps we clarify why this is mentioned at all.

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+And even though a `class` keyword is available nowadays, JavaScript is still a _prototype-based_ language.
```
SleeplessByte, line 8: ...then this sentence makes more sense.

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+
+To understand what that means and how JavaScript actually works, we will go back to the time when there were no classes.
```
SleeplessByte, line 10:
```suggestion
To understand what it means to be a prototype-based language and how JavaScript actually works, we will go back to the time when there were no classes.
```

I recommend we don't refer back to the previous sentences because the _that_ is slightly ambiguous.

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+
+It is important to note that in JavaScript, the instances and the constructor function keep a relationship to each other even after the instances were created.
+Every instance object includes a hidden, internal property referred to as `[[prototype]]` in the language specification.
+It holds a reference to the value of the `prototype` key of the constructor function.
+Yes, you read that correctly, a JavaScript function can have key/value pairs because it is also an object behind the scenes.
+
+Since 2015, `[[prototype]]` can be accessed via `Object.getPrototypeOf()`.
+Before that, it was accessible via the key `__proto__` in many environments.
+
+Do not confuse the prototype of an object (`[[prototype]]`) with the `prototype` property of the constructor function.
```
SleeplessByte, line 37: Perhaps we recap?

```markdown
To summarize:

- Constructors in JavaScript are regular functions
- Constructing a new instance creates an object with a relation to its constructor called the prototype.
- Functions are objects (callable objects).
- Because functions are objects, it can have properties.
- The constructor's (function) prototype property will become the instance's prototype
```

Something like this? (Please check if correct).

Overall I like the `about` a lot, but I think it will help summarize the sections because it's a lot of "new" concepts.

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Instance Fields
+
+Often, you want all the derived objects to include some fields and pass some initial values for those when the object is constructed.
```
SleeplessByte, line 41:
```suggestion
Often, you want all the derived objects (instances) to include some fields and pass some initial values for those when the object is constructed.
```

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+This is facilitated via the [`this` keyword][mdn-this].
```
SleeplessByte, line 42:
```suggestion
This can be facilitated via the [`this` keyword][mdn-this].
```

Because the more advanced and flexible way is using `defineProperty` / `defineProperties`.

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+}
+
+const myCar = new Car('red', '2mt');
+myCar.color;
+// => 'red'
+myCar.engineRunning;
+// => false
+```
+
+### Instance Methods
```
SleeplessByte, line 62: I think this is a very clear section. Nice.

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+
+const myCar = new Car();
+myCar.startEngine();
+myCar.engineRunning;
+// => true
+```
+
+### The Prototype Chain
+
+`myCar` in the example above is a regular JavaScript object and if we would inspect it (e.g. in the browser console) we would not find a property `startEngine` with a function as value directly inside the `myCar` object.
```
SleeplessByte, line 92:
```suggestion
`myCar` in the example above is a regular JavaScript object and if we would inspect it (e.g. in the browser console), we would not find a property `startEngine` with a function as value directly inside the `myCar` object.
```

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+So how does the code above even work then?
+
+The secret here is called the _prototype chain_.
+When you try to access any property (field or method) of an object, JavaScript first checks whether the respective key can be found directly in the object itself.
+If not, it continues to look for the key in the object referenced by the `[[prototype]]` property of the original object.
+As mentioned before, in our example `[[prototype]]` points to the `prototype` property of the constructor function.
+That is where JavaScript would find the `startEngine` function because we added it there.
```
SleeplessByte, line 99: I think it makes sense to add the example in code again after this paragraph, to illustrate it.

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+
+And the chain does not end there.
+The `[[prototype]]` property of `Car.prototype` (`myCar.[[prototype]].[[prototype]]`) references `Object.prototype` (the `prototype` property of the `Object` constructor function).
+It contains general methods that are available for all JavaScript objects, e.g. `toString()`.
+In conclusion, you can call `myCar.toString()` and that method will exist because JavaScript searches for that method throughout the whole prototype chain.
```
SleeplessByte, line 104: Perhaps adding here that the prototype of `Object` is normally `null` which ends the chain?

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+You can find a detailed example in the [MDN article "Inheritance and the prototype chain"][mdn-prototype-chain-example].
+
+Note that the prototype chain is only travelled when retrieving a value.
```
SleeplessByte, line 107: Admonition here might be nice? To make it stand out.

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+Setting or deleting a property of an instance object only targets that specific instance.
```
SleeplessByte, line 108: "Unlike most class-based inheritance" or something similar? I think we should point out the differences to what people are used with.

Perhaps also changes to _setting a property directly or deleting a property_, because you _can_ set properties via the prototype chain, but it's convoluted (requires a method that sets it, effectively).

concepts/classes/about.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Dynamic Methods
```
SleeplessByte, line 110:
```suggestion
### Dynamic Methods
```

I don't know if this title is clear. Maybe it could be `Dynamic Methods / How to add methods to already existing instances, via their template`. I hate it, but that's the information I'm missing. Something explicit might also be written inside the paragraph below.

SleeplessByte (commented):
Looking great. Few suggestions.

concepts/classes/about.md
junedev: Good point!

concepts/classes/about.md
junedev: Great idea with the summary, I like it!

concepts/classes/about.md
junedev: Added the first part you suggested as a separate sentence as it was hard to squeeze in.

concepts/classes/about.md
junedev: :+1: I used "caution".

concepts/classes/about.md
junedev: You are right, it is not really clear what this is about. I adjusted the header and added a general sentence below.

junedev:
@SleeplessByte Thanks again for the review of the concept content. I liked all the suggestions and applied the respective changes. I filled in the missing introduction files now that the about content was fixed so this is now ready to go once someone reviewed the rest.

concepts/classes/about.md
```diff
@@ -0,0 +1,285 @@
...
+
+Often, you want all the derived objects (instances) to include some fields and pass some initial values for those when the object is constructed.
+This can be facilitated via the [`this` keyword][mdn-this].
+Inside the constructor function, `this` represents the new object that will be created via `new`.
+`this` is automatically returned from the constructor function when it is called with `new`.
+
+That means we can add fields to the new instance by adding them to `this` in the constructor function.
+
+```javascript
+function Car(color, weight) {
```
pertrai1, line 58: Not sure if it is helpful to show how this is constructed under the hood as an object:

```
function Car(color, weight) {
    let thisObj = Object.create(null);
    thisObj.color = color;
    thisObj.weight = weight;
    return thisObj;
}
```
`this` removes the need to explicitly set the object and return it. Might be overkill or not needed :-)

junedev: Since the student might not know about `Object.create` at this point, I felt this code is more confusing then helping. Imo the important part is that you don't have to explicitly return `this` and that is covered in the text above.

> `this` is automatically returned from the constructor function ...

pertrai1: Sounds great @junedev. This concept looks awesome and enjoyed reading the code when reviewing.

SleeplessByte: (fun fact, returning a _new_ object from the constructor function is in fact a way to get closures/private variables _and_ private functions from a constructor, however, you need to set the prototype at `Object.create` if you want to keep inheritance to work)

SleeplessByte:
*Let's simulate some windows :ocean::tornado:!*
In this exercise, you will be simulating a windowing based computer system. You will create some windows that can be moved and resized. Find your way around classes (and learn a bit about prototypes in the process)

PR #1493, opened 2021-11-03
fix math error

edwinruvalcaba:
Step 4: Adding 73 to player score should update to 12771081 instead of 12771091


junedev (approved):
Good catch, thanks!

PR #1494, opened 2021-11-05
Update introduction.md

RIKnight:
This document uses the term "Array.prototype" without defining it.   This is quite confusing and should be fixed.

### DIFF
--- a/concepts/arrays/about.md
+++ b/concepts/arrays/about.md
@@ -146,7 +146,7 @@ Read [this article][instanceof-vs-array-is-array] for more information.
 
 ## Array Methods
 
-Some of the [methods][array_methods] that are available on `Array.prototype` can be used to add or remove from the array.
+Some of the [methods][array_methods] that are available on every Array[^0] object can be used to add or remove from the array.
 Here are a few of them:
 
 ### push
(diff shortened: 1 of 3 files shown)


concepts/arrays/introduction.md
```diff
@@ -33,7 +33,7 @@ numbers;
 
 ## Methods
 
-Some of the [methods][array_methods] that are available on `Array.prototype` can be used to add or remove from the array.
+Some of the [methods][array_methods] that are available on `Array.prototype`[^0] can be used to add or remove from the array.
```
junedev, line 36: I think the best way forward here would be to write "... that are available on every array can be used...". We have a separate concept that introduces `prototype` in detail later on.

Would be great if you could apply this fix to all the 3 files mentioned in the automated PR comment.

junedev (changes requested):

tejasbubane (approved):
Thanks! :+1:

concepts/arrays/about.md
```diff
@@ -146,7 +146,7 @@ Read [this article][instanceof-vs-array-is-array] for more information.
 
 ## Array Methods
 
-Some of the [methods][array_methods] that are available on `Array.prototype` can be used to add or remove from the array.
+Some of the [methods][array_methods] that are available on every Array[^0] object can be used to add or remove from the array.
```
junedev, line 149: The word [methods] already links to the page you have in the footnote and footnotes are only used for quotes.

RIKnight: ok, removed them!

SleeplessByte (approved):

junedev (approved):

PR #1499, opened 2021-11-06
Lasagna Master: Check that original recipe is not modified

### LINKED ISSUE #1496: Lasagna Master: Improve `scaleRecipe` tests
The instructions for `scaleRecipe` in the concept exercise `Lasagna Master` mention that the object that is passed in should not be modified. However, the tests currently don't check for that. That should be fixed.

(Thanks to @SethGunnells for pointing this out.)

SethGunnells:
This adds a check that, for a number of people not equal to 2, the original object is deeply equal to a clone of itself made earlier. This I think is more accurate than simply checking that the original recipe and the resulting object are not equal by reference since it is possible to modify the original without returning it at the end. This check also allows them to mess with the original object but end up with the same values at the end which is functionally equivalent to leaving it unmodified.

Closes #1496

### DIFF
--- a/exercises/concept/lasagna-master/lasagna-master.spec.js
+++ b/exercises/concept/lasagna-master/lasagna-master.spec.js
@@ -223,6 +223,21 @@ describe('scaleRecipe', () => {
   test('works for an empty recipe', () => {
     expect(scaleRecipe({})).toEqual({});
   });
+
+  test('does not modify the original recipe', () => {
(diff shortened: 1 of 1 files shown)


junedev (approved):
Thanks for identifying this issue and fixing it!

iHiD:
Nice work @SethGunnells. Thanks for contributing! :)

PR #1519, opened 2021-11-14
fix: remove unnessary class in rotational-cipher

### LINKED ISSUE #1511: Rotational Cipher: Remove unnessary class
The Rotational Cipher exercise stub currently contains a class definition that serves no purpose.

The exercise should be changed so that the stub and exemplar only use a simple arrow function:
```javascript
export const rotate = () => {
  // ...
};
```
The following files need to be updated:
* `<exercise>.js`
* `<exercise>.spec.js`
* `.meta/proof.ci.js`
* root level `config.json` (remove "classes" from the prerequisites of the exercise in case it is present there)

You can find general information about the structure of practice exercises here: https://exercism.org/docs/building/tracks/practice-exercises

kerolloz:
fixes #1511

### DIFF
--- a/exercises/practice/rotational-cipher/.meta/proof.ci.js
+++ b/exercises/practice/rotational-cipher/.meta/proof.ci.js
@@ -1,17 +1,15 @@
-export class RotationalCipher {
-  static rotate(text, shift) {
-    return [...text]
-      .map((c) => {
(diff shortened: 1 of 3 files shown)


kerolloz:
@junedev thank you for your patience.

junedev (approved):

PR #1522, opened 2021-11-14
Remove unnecessary class in Pig-Latin

### LINKED ISSUE #1510: Pig Latin: Remove unnessary class
The Pig Latin stub currently contains a class definition that serves no purpose.

The exercise should be changed so that the stub and exemplar only use a simple arrow function:
```javascript
export const translate = () => {
  // ...
};
```
The following files need to be updated:
* `<exercise>.js`
* `<exercise>.spec.js`
* `.meta/proof.ci.js`
* root level `config.json` (remove "classes" from the prerequisites of the exercise in case it is present there)

You can find general information about the structure of practice exercises here: https://exercism.org/docs/building/tracks/practice-exercises

asadrjhk:
Remove unnecessary class in Pig-Latin exercise.
closes #1510


junedev (approved):
Thanks for the PR!

PR #1527, opened 2021-11-17
Fix minor inconsistency with default Error

goldendemise:
The name of the function is `updateScore` rather than `addToScore` - this update changes the thrown error to match the function name.


junedev (approved):
Good catch!

PR #1530, opened 2021-11-18
Added new exercise: rest-api

### LINKED ISSUE #923: Add new exercise: rest-api
[problem specification ](https://github.com/exercism/problem-specifications/tree/master/exercises/rest-api)| [description](https://github.com/exercism/problem-specifications/blob/master/exercises/rest-api/description.md) | [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/rest-api/canonical-data.json)

You may find it helpful to follow [this checklist](https://github.com/exercism/javascript/blob/83feda2af955df1ef9f3a8080338f506f0fdf696/CONTRIBUTING.md#new-exercise) from the [CONTRIBUTING.md](https://github.com/exercism/javascript/blob/83feda2af955df1ef9f3a8080338f506f0fdf696/CONTRIBUTING.md) resource.

lpizzinidev:
close #923

### DIFF
--- a/config.json
+++ b/config.json
@@ -1686,7 +1686,7 @@
         "name": "Rest API",
         "uuid": "75b4816d-a5d7-4840-92d6-7249c9e8beeb",
         "practices": [],
-        "prerequisites": ["arrays", "classes", "for-loops", "maps"],
+        "prerequisites": ["arrays", "classes", "array-loops", "objects"],
         "difficulty": 6,
         "topics": ["algorithms", "logic"]
       }
--- a/config.json
+++ b/config.json
@@ -1682,13 +1682,13 @@
         "topics": ["parsing", "strings"]
       },
       {
-        "slug": "rail-fence-cipher",
-        "name": "Rail Fence Cipher",
-        "uuid": "5f0352e5-0988-434a-82c7-f6810d41d63a",
+        "slug": "rest-api",
+        "name": "Rest API",
+        "uuid": "75b4816d-a5d7-4840-92d6-7249c9e8beeb",
         "practices": [],
-        "prerequisites": ["arrays", "strings", "for-loops", "conditionals"],
-        "difficulty": 3,
-        "topics": ["parsing", "strings", "iteration"]
+        "prerequisites": ["arrays", "classes", "for-loops", "maps"],
+        "difficulty": 6,
+        "topics": ["algorithms", "logic"]
       }
     ]
   },
--- a/config.json
+++ b/config.json
@@ -1689,6 +1689,15 @@
         "prerequisites": ["arrays", "classes", "array-loops", "objects"],
         "difficulty": 6,
         "topics": ["algorithms", "logic"]
+      },
+      {
+        "slug": "rest-api",
+        "name": "Rest API",
+        "uuid": "75b4816d-a5d7-4840-92d6-7249c9e8beeb",
+        "practices": [],
+        "prerequisites": ["arrays", "classes", "for-loops", "maps"],
+        "difficulty": 6,
+        "topics": ["algorithms", "logic"]
       }
     ]
   },
--- a/config.json
+++ b/config.json
@@ -1682,20 +1682,20 @@
         "topics": ["parsing", "strings"]
       },
       {
-        "slug": "rest-api",
-        "name": "Rest API",
-        "uuid": "75b4816d-a5d7-4840-92d6-7249c9e8beeb",
+        "slug": "rail-fence-cipher",
+        "name": "Rail Fence Cipher",
+        "uuid": "5f0352e5-0988-434a-82c7-f6810d41d63a",
         "practices": [],
-        "prerequisites": ["arrays", "classes", "array-loops", "objects"],
-        "difficulty": 6,
-        "topics": ["algorithms", "logic"]
+        "prerequisites": ["arrays", "strings", "for-loops", "conditionals"],
+        "difficulty": 3,
+        "topics": ["parsing", "strings", "iteration"]
       },
       {
         "slug": "rest-api",
         "name": "Rest API",
         "uuid": "75b4816d-a5d7-4840-92d6-7249c9e8beeb",
         "practices": [],
-        "prerequisites": ["arrays", "classes", "for-loops", "maps"],
+        "prerequisites": ["arrays", "classes", "array-loops", "objects"],
         "difficulty": 6,
         "topics": ["algorithms", "logic"]
       }
--- a/exercises/practice/rest-api/.docs/instructions.append.md
+++ b/exercises/practice/rest-api/.docs/instructions.append.md
@@ -1,0 +1,9 @@
+# Instructions append
+
+## Implementation
+
+Implement the `get` and `post` methods from the `RestAPI` class.
+
+- no actual HTTP server should be implemented, just the handler functions should be written
+- an in-memory object can serve as "database"
+- the constructor function needs to accept a seed for the database as argument (and set up a sensible default if nothing was passed)
--- a/exercises/practice/rest-api/.docs/instructions.append.md
+++ b/exercises/practice/rest-api/.docs/instructions.append.md
@@ -4,6 +4,8 @@
 
 Implement the `get` and `post` methods from the `RestAPI` class.
 
-- no actual HTTP server should be implemented, just the handler functions should be written
-- an in-memory object can serve as "database"
-- the constructor function needs to accept a seed for the database as argument (and set up a sensible default if nothing was passed)
+You should write only the handler functions, without implementing a real HTTP server.
+You can mock the database using an in-memory object that will contain all stored users.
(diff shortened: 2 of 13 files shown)


SleeplessByte (commented):
Overall looks good. 

However, on the problem-spec repo I had previously commented this. Whilst the newest RFCs regarding HTTP allow GETs with bodies, the semantics are lost. For REST-apis (and especially HATEOS and earlier), GETs just don't have bodies / should not interpret bodies. 

I'm strongly in favour of url encoding the payload, and/or use `encodeUriComponent`.

lpizzinidev:
@SleeplessByte 
Ok, so I should change the `get` function in the `proof.ci.js` file and remove the `payload` parameter.
Also, I should expect an URL with query params, like for example `/users?users=Bob`.
Did I understand it correctly?

SleeplessByte:
Yes that sounds good

lpizzinidev:
@SleeplessByte I've applied the suggested edits. Let me know if further changes are needed.

exercises/practice/rest-api/rest-api.spec.js
```diff
@@ -0,0 +1,345 @@
...
+      expect(restAPI.get('/users?users=Bob')).toEqual({
+        users: [
+          {
+            name: 'Bob',
+            owes: {},
+            owed_by: {},
+            balance: 0.0,
+          },
+        ],
+      });
```
junedev, line 44: General comment regarding test file: I would recommend thinking about the readability of the tests. The statements currently span so many lines that is is hard to grasp the actual statements. I think it would help to maybe write some of the objects in one line and to use variables to define them. Here an example how this could look like applying both ideas:
```javascript
      const seedUsers = [
        { name: 'Adam', owes: {}, owed_by: {}, balance: 0.0 },
        { name: 'Bob', owes: {}, owed_by: {}, balance: 0.0 },
      ];
      const restAPI = new RestAPI({ users: seedUsers });
```

exercises/practice/rest-api/.docs/instructions.md
```diff
@@ -0,0 +1,42 @@
...
+| List of user information | GET         | /users | `{"users":["Adam","Bob"]}`                                                | `{"users":<List of all User objects>}` | `{"users":<List of User objects for <users> (sorted by name)}`                  |
+| Create user              | POST        | /add   | `{"user":<name of new user (unique)>}`                                    | N/A                                    | `<User object for new user>`                                                    |
+| Create IOU               | POST        | /iou   | `{"lender":<name of lender>,"borrower":<name of borrower>,"amount":5.25}` | N/A                                    | `{"users":<updated User objects for <lender> and <borrower> (sorted by name)>}` |
+
+## Other Resources
+
+- [https://restfulapi.net/](https://restfulapi.net/)
+- Example RESTful APIs
+  - [GitHub](https://developer.github.com/v3/)
+  - [Reddit](https://www.reddit.com/dev/api/)
```
junedev, line 42: General comment regarding the instructions: It seems you applied the suggestion from SleeplessByte to not have a payload (http request body) for the GET request. However, currently the student that would attempt this exercise would not know about this. I would recommend to fix this by adding an instructions.append.md file that explains how the payload is provided in case of the GET request.

Also there are other things that are currently not mentioned anywhere and should probably be clarified in the append file:
- no actual HTTP server should be implemented, just the handler functions should be written
- an in-memory object can serve as "database"
- the constructor function needs to accept a seed for the database as argument (and set up a sensible default if nothing was passed)

In addition is would probably be good to add the parameter names to the stub so the student understands what is provided as input (e.g. `post(url, payload)`).

config.json
```diff
@@ -1680,6 +1680,15 @@
         "prerequisites": ["arrays", "strings", "for-loops", "objects"],
         "difficulty": 3,
         "topics": ["parsing", "strings"]
+      },
+      {
+        "slug": "rest-api",
+        "name": "Rest API",
+        "uuid": "75b4816d-a5d7-4840-92d6-7249c9e8beeb",
+        "practices": [],
+        "prerequisites": ["arrays", "classes", "for-loops", "maps"],
```
junedev, line 1689: `maps` is the concept for this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map
Are you sure this is a prerequisite here? Do you maybe mean `objects` instead (this concept covers using objects as simple key/value collections).

The for-of loop is covered in the concept `array-loops`. Maybe you want to use that here instead of `for-loops`.

junedev (commented):
Sorry you had to wait a week for the review! Thanks for all the work, looks good overall!

lpizzinidev:
@junedev 
Regarding the `instructions.md` file.
I've split the methods table into a `GET` one, with the `/users` method and its query params, and a `POST` one, with the other two methods like before.
I think that this clarifies how parameters should be handled with the `GET` method.
Let me know if there are other changes to be made.

junedev:
@lpizzinidev It is better to never modify the instructions.md file when it was copied from problem spec. Soon, configlet will include functionality to sync this file with the version in the problem specification repo. Imagine there was an update to the instructions in problem spec and someone runs the command and the reviewer does not notice information got lost. Then all the modifications are gone and the instructions are incomplete. This is what happened for several exercises in the Go track and it took some time to resolve this. This is why I am so cautious here.

I get the good intentions of course but I would recommend to only use the append file instead. Write something like "For this implementation, in case of a `GET` request, the payload should be part of the url ..." and whatever else you want to explain there.

exercises/practice/rest-api/.docs/instructions.append.md
```diff
@@ -0,0 +1,9 @@
+# Instructions append
+
+## Implementation
+
+Implement the `get` and `post` methods from the `RestAPI` class.
+
+- no actual HTTP server should be implemented, just the handler functions should be written
+- an in-memory object can serve as "database"
+- the constructor function needs to accept a seed for the database as argument (and set up a sensible default if nothing was passed)
```
junedev, line 9: Could you try to convert these into your own proper sentences? Currently it looks weird because they address different things and the relation to the sentence above is unclear.

exercises/practice/rest-api/rest-api.spec.js
```diff
@@ -0,0 +1,153 @@
...
+      const restAPI = new RestAPI({ users: seedUsers });
+      const expectedUsers = [
+        { name: 'Adam', owes: { Bob: 3.0 }, owed_by: {}, balance: -3.0 },
+        {
+          name: 'Bob',
+          owes: { Chuck: 3.0 },
+          owed_by: { Adam: 3.0 },
+          balance: 0.0,
+        },
+      ];
```
junedev, line 89: Use the same format for all objects in the array. So if one user does not fit in one line, write all of them like line 84-87.

exercises/practice/rest-api/rest-api.spec.js
```diff
@@ -0,0 +1,153 @@
...
+          owes: { Adam: 3.0, Chuck: 3.0 },
+          owed_by: {},
+          balance: -6.0,
+        },
+      ];
+      expect(
+        restAPI.post('/iou', {
+          lender: 'Adam',
+          borrower: 'Bob',
+          amount: 3.0,
```
junedev, line 70: I would recommend to use another variable to separate out the request payload here. Then you get one readable line that specifies the assertion. `expect(restAPI.post('/iou', payload)).toEqual(...)`. The student can then check what's in the payload or expected users variable depending on what they want to check out.

exercises/practice/rest-api/rest-api.spec.js
```diff
@@ -0,0 +1,153 @@
...
+      const restAPI = new RestAPI({ users: [] });
+      expect(restAPI.get('/users')).toEqual({ users: [] });
+    });
+    test('add user', () => {
+      const restAPI = new RestAPI({ users: [] });
+      expect(restAPI.post('/add', { user: 'Adam' })).toEqual({
+        name: 'Adam',
+        owes: {},
+        owed_by: {},
+        balance: 0.0,
```
junedev, line 15: Writing all numbers as `x.0` instead of just `x` does not seems very idiomatic to me since there is no integer vs. float distinction in JavaScript. But it is not wrong in terms of this file making the tests pass of course. :woman_shrugging:

junedev:
Thanks for the update! I did another round of review and left some more comments. Would be great if you could address those as well (in addition to the instructions.md issue) and then this can be merged.

lpizzinidev:
@junedev Applied suggested changes, let me know if something is missing

SleeplessByte (approved):
Hehe. That proof is _very_ dense. 

Only request is if you can add a newline between the tests. It's difficult to read right now because there is no good vertical rhythm or separation.

SleeplessByte (approved):

SleeplessByte:
Wonderful. Thanks Luca!!

SleeplessByte:
image

PR #1545, opened 2021-11-30
Fix two small typos for elyses looping enchantments

khendrikse:
I just saw these two tiny typo's. So here is a fix.


SleeplessByte (approved):
Thanks!

Issue #1234, opened 2021-07-29
Improve Concept Exercise: Functions (Lasagna Master)


junedev:
## Getting started
Here you can read about what Concept Exercises are and how they are structured:
- [What are Concept Exercises?](https://exercism.org/docs/building/product/concept-exercises)
- [Concept Exercises](https://exercism.org/docs/building/tracks/concept-exercises)
- [Concepts](https://exercism.org/docs/building/tracks/concepts)

If you have not done so yet, it is probably also helpful to do a couple of "Learning Exercises" (this is how they are called on the site) yourself.

## Goal

The goal of this issue is to improve the existing concept exercise [Lasagna Master](https://github.com/exercism/javascript/tree/main/exercises/concept/lasagna-master) that teaches the `functions` concept.

The exercise is already in good shape but to save some time before V3 launch, the about.md is very limited at the moment and some other bits and pieces are missing. Here a more detailed list of what could be improved.

### Exercise Improvements
- There is currently no task related to function expressions. It would be nice to have one but it should not go to far into the territory of e.g. the `callbacks` concept.

### About.md Additions
The about.md file is currently mostly a copy of introduction.md with some minor additions. Here some ideas what could be added.
- The arguments array and its pitfalls
- How to mitigate pass by reference issues (how to avoid modifying the input by mistake), e.g. copy object
- Maybe some more details on when to use function expressions
- Explain that function declarations are hoisted and function expressions are not
- Add a reference to the recursion concept
- The scope section should link to a variable concept that explains variable shadowing
- IIFE
- https://github.com/exercism/javascript/pull/1183/files#r684702295

Possibly we could also include some more general recommendations regarding functions.
- Good names for functions
- One functions / one purpose
- Avoid side effects / pure functions
- How to use functions with good names to clean up code/as comments
- How function names are determined (see also https://github.com/exercism/javascript/pull/1160#discussion_r654697390)

## Collect Ideas for Analyzer Comments
If you feel like thinking about analyzer feedback but don't feel confident to build it directly, you can collect ideas for the analyzer in the design.md file. Just add a `## Analyzer` section. You can see an example [here](https://github.com/exercism/javascript/blob/79082d00a90e3a6ffd7b47cfc4228c91b0f205b6/exercises/concept/bird-watcher/.meta/design.md#analyzer).

## Help

You can choose to do this solo-style, or collaborate with multiple people on this. The suggested approach is to

[...]

PR #1550, opened 2021-12-02
Translation Service: remove instructions as steps

khendrikse:
While doing the translation service exercise I was really confused because it seemed that the first few instructions about the api interface were reading as steps:

image

As this is already a concept that can be hard to grasp, I thought if we removed them as headers and kept them as part of the instructions (as I assume they are?) it will be easier to do this exercise. 

Maybe there is another way to make this happen that is semantically more correct? But right now it appears that a h3 or h2 will turn into a step?

### DIFF
--- a/exercises/concept/translation-service/.docs/instructions.md
+++ b/exercises/concept/translation-service/.docs/instructions.md
@@ -4,11 +4,11 @@ In this exercise you'll be providing a `TranslationService` where paid members h
 
 You have found an out-of-space translation API that is able to fulfill any translation _request_ in a reasonable amount of time, and you want to capitalize on this.
 
-## The API interface
+**The API interface**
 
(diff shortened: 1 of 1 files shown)


junedev (approved):
Thanks for reporting this bug and providing the fix. I also wasn't aware h3 headers are also counted as tasks. I think you fix is good.

junedev:
Pro tip: When you mention the issue in the PR description with `Fixes #1234` than GitHub will automatically link the PR to the issue and close the issue once the PR is merged. See [GitHub docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) for more details.

(I manually added the link this time.)

PR #1570, opened 2021-12-17
Add note in the instructions.append.md

### LINKED ISSUE #1564: Gigasecond problem is confusing
I spent a lot of time on the gigasecond problem in JavaScript (https://exercism.org/tracks/javascript/exercises/gigasecond), just trying to understand it. It mentions that the input is a number, but it is a string. It also doesn't have any of the hints that the learning track exercises have. 

I can certainly understand scaffolding and removing the scaffold as a part of the learning process but the wording of the problem and the lack of hints or other information or even just examples of input and output made it very difficult to understand what was expected. 

I'd like to suggest updating this problem at least in JS with the following:
- Indicating that the input is a date string
- Perhaps adding a hint or two about using the JavaScript date object and getTime as well as that getTime expects milliseconds (I got stuck here for a while trying not understanding why adding 1 bil seconds to getTime didn't have the expected effect)
- That the output is expected to be a date object
- A small table of example input and output

asadrjhk:
Add a note in the instructions.append.md file that the input and output of the gigasecond function is of type Date.
closes #1564

### DIFF
--- a/exercises/practice/gigasecond/.docs/instructions.append.md
+++ b/exercises/practice/gigasecond/.docs/instructions.append.md
@@ -1,3 +1,6 @@
 # Instructions append
 
 It is possible to return a correct value for this exercise by mutating the solution function argument. Although there are legitimate use cases for mutating function arguments, this is usually undesirable, and in the case of this exercise, clearly unexpected. For this reason, the test suite has a test that fails in case the argument has been modified after the function execution.
+
+## Notes
+The input and output of the gigasecond function is of type [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date).


exercises/practice/gigasecond/.docs/instructions.append.md
```diff
@@ -1,3 +1,6 @@
 # Instructions append
 
 It is possible to return a correct value for this exercise by mutating the solution function argument. Although there are legitimate use cases for mutating function arguments, this is usually undesirable, and in the case of this exercise, clearly unexpected. For this reason, the test suite has a test that fails in case the argument has been modified after the function execution.
+
+## Notes
+The input and output of the gigasecond function is of type [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date).
```
junedev, line 6: * Put your two new lines before line 3. Line 3 is also a note but yours is more important.
* Wrap `gigasecond` in backticks, see https://exercism.org/docs/building/markdown/markdown#h-code
* See https://exercism.org/docs/building/markdown/markdown#h-links for how we provide links in the markdown

Currently, the CI pipeline fails because the formatting is not right. See https://github.com/exercism/javascript/blob/main/CONTRIBUTING.md#format for how to run the formatter.

junedev (commented):
Thanks! I left some comments how to make the addition adhere to the guidelines we have.

asadrjhk:
sorry, about that but this time i try to make the PR using my local system, which makes the new pull request, i think i need to close this.

PR #1571, opened 2021-12-17
Add note in the instructions.append.md

### LINKED ISSUE #1564: Gigasecond problem is confusing

asadrjhk:
Add a note in the instructions.append.md file that the input and output of the gigasecond function is of type Date.
closes #1564

### DIFF
--- a/exercises/practice/gigasecond/.docs/instructions.append.md
+++ b/exercises/practice/gigasecond/.docs/instructions.append.md
@@ -1,10 +1,9 @@
 # Instructions append
 
 ## Notes
+
 The input and output of the `gigasecond` function is of type [Date][date-link].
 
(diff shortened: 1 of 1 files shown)


junedev (approved):

junedev:
Thanks for addressing the review comments!

PR #1572, opened 2021-12-17
Why I can't use functional approach? [javascript/exercises/concept/bird-watcher]

dotnil:
I don't want to use ancient "for", thanks, I want my guard clauses and map

```
export function fixBirdCountLog(birdsPerDay) {
  return birdsPerDay.map(function (day, index) {
    const isOdd = index % 2 === 0;
    if (isOdd) {
      return day + 1;
    }
    return day;
  });
}
```

2021-12-17-19-40-59-region

### DIFF
--- a/exercises/concept/bird-watcher/bird-watcher.spec.js
+++ b/exercises/concept/bird-watcher/bird-watcher.spec.js
@@ -53,14 +53,6 @@ describe('bird watcher', () => {
       expect(fixBirdCountLog(birdsPerDay)).toEqual(expected);
     });
 
-    test('does not create a new array', () => {
-      const birdsPerDay = [2, 0, 1, 4, 1, 3, 0];
-
-      // This checks that the same object that was passed in is returned.
-      // https://jestjs.io/docs/expect#tobevalue
-      expect(Object.is(fixBirdCountLog(birdsPerDay), birdsPerDay)).toBe(true);
-    });
-
     test('works for a short bird count list', () => {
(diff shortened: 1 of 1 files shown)


junedev:
I am sorry but I have to reject this change.

Here the reasoning:
* A lot of students like this exercise as is and people request mentoring for it regularly (meaning it is not too trivial for everyone to just go over it).
* To be a well rounded JS developer, it is good to practice "manipulate an array without making a copy". `map` can be quite performance intensive when working with big arrays.
* Learning exercises are not meant to be solved in the best way possible like practice exercises. They are meant to be solved with the concepts they are teaching to practice. Also learning exercises are primarily made for students that do not know JavaScript yet. If you are already familiar with the language, you can either solve the early learning exercises quickly if you are interested in some of the later concepts or you can just turn off learning mode altogether and focus on the practice exercises.
* For loops are by no means something from the past. If you are writing code where performance is critical, where an existing array is modified or where the length of the array can change during the iteration process, you will still see for loops being used in modern JavaScript code.

SleeplessByte:
To expand on @junedev 's comment, which I agree with, the _most_ important reason is really that this exercise _is about `for` loops_. I want to re-iterate that. It doesn't matter if you already know `for` loops or understand them. As part of the syllabus, you must show your knowledge of `for` loops.

In the same way, we also have an exercise where you're forced to use a nice abstraction like `map`, where the goal is to use a `map`.

In practice exercises, however, do what you want :)

Issue #984, opened 2021-02-15
[v3] Add prerequisites 3/6


TomPradat:
```
anagram

acronym

high-scores

isogram

matching-brackets

phone-number

scale-generator

series

largest-series-product

transpose

grep

rectangles

spiral-matrix

ocr-numbers

saddle-points

forth

food-chain

house

isbn-verifier

pig-latin
```

AnishDe12020:
I am willing to take this issue up (at least some of the exercises). Just let me know if I should make the PRs, one for each exercises, one for a batch of exercises or one for all of them.

TomPradat:
I'd suggest that you make a PR for each exercise but if you feel like you want to do a lot of exercises you can take 4-5 by PR.

Trying to do all of them at once is not a good idea IMO, it's indigestible for both you and the reviewer.

Idk if this might help but here is the steps I used when doing some :

1 - Doing/trying the exercise
2 - Looking at the exercise's proof of work
3 - Looking at the submitted solutions.
4 - Adding the prerequisites

AnishDe12020:
> I'd suggest that you make a PR for each exercise but if you feel like you want to do a lot of exercises you can take 4-5 by PR.
> 
> Trying to do all of them at once is not a good idea IMO, it's indigestible for both you and the reviewer.
> 
> Idk if this might help but here are the steps I used when doing some :
> 
> 1 - Doing/trying the exercise 2 - Looking at the exercise's proof of work 3 - Looking at the submitted solutions. 4 - Adding the prerequisites

Sounds good to me

TomPradat:
Here is an updated list (25/10/2022) :

```
matching-brackets

scale-generator

series

largest-series-product

transpose

grep

rectangles

spiral-matrix

ocr-numbers

saddle-points

forth

food-chain

house

isbn-verifier

pig-latin

```

safwansamsudeen:
Is this still open? Would love to work on it if it is!

TomPradat:
@safwansamsudeen Yes ! See https://github.com/exercism/javascript/issues/984#issuecomment-984682274 for suggestions on how to do it. Let me know if you have any questions

junedev:
@safwansamsudeen If you work on this, please start out with a small batch, e.g. 3 exercises max.
[...]

PR #1575, opened 2021-12-22
add prerequisities for phone number exercise

AnishDe12020:
I have added the prerequisites for the phone number exercise

references #984

### DIFF
--- a/config.json
+++ b/config.json
@@ -821,7 +821,13 @@
         "name": "Phone Number",
         "uuid": "82775adb-eabe-4d44-91f5-4080b8834a4a",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "strings",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -821,7 +821,13 @@
         "name": "Phone Number",
         "uuid": "82775adb-eabe-4d44-91f5-4080b8834a4a",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "strings",
+          "arrays",
```
junedev, line 826: Can you say a sentence why you think arrays are needed as prerequisite?

Everything else looks good to me.

AnishDe12020: Oh, I had actually used arrays in my solution but now that I see it, it can be done without arrays. I will remove it.

PR #1578, opened 2021-12-23
Prerequisites for isogram practice exercise

AnishDe12020:
I have added the prerequisites for the isogram practice exercise.

references #984


config.json
```diff
@@ -779,7 +779,7 @@
         "name": "Acronym",
         "uuid": "440d78d1-9dea-466f-9bd4-935eed067409",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["strings", "arrays"],
```
junedev, line 782: Same as for acronym, I don't understand the selection here. I would have expected strings and regular expressions.

AnishDe12020: This exercise can be solved without regular expressions (although it will make it a bit longer). Also, I have added arrays because sets are used in the solving of this (my solution as well as top solutions which solve the latest version of the solution use it). I can add regular expressions if required.

SleeplessByte: https://exercism.org/tracks/javascript/exercises/isogram/solutions/junedev
https://exercism.org/tracks/javascript/exercises/isogram/solutions/sleeplessbyte

Here are ours.

I did check the first few community solutions, and it seems that _each one_ uses regular expressions. So even though you can solve it without, I think it should be included. The majority of one liners don't use arrays, so I don't think it's a prerequisite to finish the exercise, but I can see how you must be able to "see" the string as an "array of characters" to solve it. I would be _okay_ leaving `arrays` in.

AnishDe12020: Cool, I have added it in https://github.com/exercism/javascript/pull/1578/commits/1a7e3f165124771f1d0eea0ad21947da572f5f30

junedev: @AnishDe12020 Looks like there is a small mistake. You added something to "topics" which is not used anymore.

AnishDe12020: I got a bit confused actually. Nevertheless, I have fixed it in https://github.com/exercism/javascript/pull/1578/commits/39e603536ee8287a89f8d788ea74ce58b0ee0860. Do let me know if it is fine

junedev: I noticed this comment actually belongs to line 805.

AnishDe12020: Yeah, I have fixed it there itself.

Issue #1528, opened 2021-11-17
Implement new Concept Exercise: Dates


junedev:
Note: The concept was already created, only the exercise is missing now.

## Getting Started

If you have not yet contributed to concept exercises before, this task requires some upfront reading to acquire the necessary background knowledge.

Here you can read about what Concept Exercises are and how they are structured:

- [What are Concept Exercises?](https://exercism.org/docs/building/product/concept-exercises)
- [Concept Exercises](https://exercism.org/docs/building/tracks/concept-exercises)
- [Concepts](https://exercism.org/docs/building/tracks/concepts)

If you have not done so yet, it is probably also helpful to do a couple of "Learning Exercises" (this is how they are called on the site) yourself. You can also look at the code of an existing concept exercise like [`bird-watcher`](https://github.com/exercism/javascript/tree/main/exercises/concept/bird-watcher) (concept [`for-loops`](https://github.com/exercism/javascript/tree/main/concepts/for-loops)) for reference.

See the documentation _above_ (general documentation), as well as [How to implement a Concept Exercise in JavaScript](https://github.com/exercism/javascript/blob/main/reference/implementing-a-concept-exercise.md).

Also be aware of these general guidelines.

- [Exercism Markdown Specification](https://exercism.org/docs/building/markdown/markdown)
- [Exercism Formatting and Style Guide](https://exercism.org/docs/building/markdown/style-guide)

## Goal

The goal here is to create a new concept exercise that teaches working with dates via `Date`.

The concept needs to be written from scratch, the exercise can be ported from another track (see details below).

## Concepts

The following concept needs to be created. You can use the `introduction.md` file of the concept also as `introduction.md` file of the exercise. No need to create different content at this point. Additionally, if you want to save some time it is ok to not have an extensive `about.md` for now. It can also be mainly the `introduction.md` content, maybe with some additions you would like to make.

- `dates` (maybe use "Date and Time" as display name in the config)

## Learning Objectives

In the concepts the student should learn about the following topics and then practice them in the concept exercise.

- how dates/timestamps are represented internally, see e.g. [MDN link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_ecmascript_epoch_and_timestamps)
- how to create dates with `new Date`, incl. some of the commonly used [variants](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#syntax)
- `Date.parse`
- `Date.now`
[...]

PR #1579, opened 2021-12-23
Creating Dates Concept

JaPatGitHub:
As per #1528


JaPatGitHub:
Some things to let you know:
i) I have copied the content of `about.md` to `introduction.md` as it is for now. I'll edit the introduction once I've decided how the concept exercise will be.
ii) I still haven't added the #1528 recommended additional content in `about.md`. I wish to work on the exercise first, as that is of higher priority I believe, and then do this?

concepts/dates/about.md
```diff
@@ -0,0 +1,256 @@
...
+
+```exercism/note
+January 1st, 1970 at 00:00:00 UTC is referred to as the Unix epoch.
+Unix is an operating system originally developed in the 1960s.
+Early Unix engineers picked that date arbitrarily because they needed to set a uniform date for the start of time, and New Year's Day, 1970, seemed most convenient.
+```
+
+### Timestamp string
+
+A string value representing a date is used, specified in a format recognized by `Date.parse()` method, about which you will learn soon in this article.
```
junedev, line 42:
```suggestion
You can pass a string value representing a date to the `Date` constructor.
The string needs to follow a format that is recognized by the `Date.parse()` method.
You will learn more about this below.
```

concepts/dates/introduction.md
```diff
@@ -0,0 +1,256 @@
+# Introduction
```
junedev, line 1: Remember that the introduction should be a condensed version of the about file that only contains the most vital information. Let's get the about file sorted first but once we got that, think about what could be removed/shortened here. There should be just enough info to understand the concept and solve the exercise. We are not too strict about this atm but I think there are some quick fixes we could do here to make this more "introduction" like, e.g. remove the note on the 1.1.1970.

package-lock.json
```diff
@@ -1,7 +1,14824 @@
 {
```
junedev: Your PR should not contain any changed to the package-lock.json file.

junedev (changes requested):
Thanks for the first draft of the concept. I didn't check everything in detail yet but just left some initial comments for you to work on.

concepts/dates/about.md
```diff
@@ -0,0 +1,256 @@
...
+
+`Z` - This _optional_ part denotes the time zone.
+If `Z` is present, the `Date` will be set to UTC.
+If `Z` is not present, it will be Local Time.
+
+If the format is invalid, `NaN` is returned
+
+Shorter variants are also possible, like `YYYY-MM-DD` or `YYYY-MM` or even `YYYY`. However not that these variants **set the `Date` to UTC**, even though `Z` not mentioned. To understand what exactly happens check out [this section][mdn-diff-assumed-timezone] of a MDN page.
+
+TODO complete eg below
```
junedev, line 125: ?

concepts/dates/about.md
```diff
@@ -0,0 +1,256 @@
...
+This can be used as an argumect for the Date object.
+
+```javascript
+// 0 means 01.01.1970 UTC+0
+const Jan01_1970 = new Date(0);
+console.log(Jan01_1970);
+
+// adding 24 hours, we get 02.01.1970 UTC+0
+const Jan02_1970 = new Date(24 * 3600 * 1000);
+console.log(Jan02_1970);
```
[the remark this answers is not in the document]
JaPatGitHub: I'm sorry about infringing the copyrights. Will credit the source.
However, there is some content I have copied but modified. The javascript.info license says that should be mentioned. How do I do that?

package-lock.json
JaPatGitHub: I think this change was made by the GitHub actions bot when I commented `/format` in the PR as I could not format it locally.
Is there a way I can revert changes in only one file?

concepts/dates/about.md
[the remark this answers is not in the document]
junedev: I would also use a footnote in this case and then say something like "Paragraph adapted from ..." there.

package-lock.json
junedev: @SleeplessByte Do you know why the formatting workflow suddenly started to change the package.json file? Some new npm version and the file is not up-to-date maybe?

@JaPatGitHub You can revert the change like this https://fettblog.eu/snippets/git/reverting-a-single-file/ but you need to find the right commit for this. So it might be faster just to copy and paste from main and then commit that change.

SleeplessByte: image

This is why, so yes.

junedev: I proposed a fix here: https://github.com/exercism/javascript/pull/1607

JaPatGitHub:
@junedev , @SleeplessByte ...the formatting ci check failed because I used tildes for exercism note and caution boxes. So anything possible to do to make prettier ignore this?

SleeplessByte:
@JaPatGitHub I think so, using `prettier-ignore` See: https://prettier.io/docs/en/ignore.html#markdown

junedev (approved):
I changed some small formatting stuff and shortened the introduction a bit. The introduction should only contain the most important info (in general and to later solve the exercise).

I will merge this now. If there are more things to fix, we can still do that later when working on the exercise.

junedev:
For the tildes we just used the version that prettier suggests in the other concepts and that worked fine so I went with the same here.

sean-m-foley:
You may want to tag this into the 'gigasecond' javascript exercise. It's very difficult to determine how to accomplish that without 'getting offsite' and googling for a long period of time to find the exact functions you have to call to accomplish the test conditions. (It also helps immensely with reading the test js and understanding what it is doing.)

JaPatGitHub:
@junedev, will what @sean-m-foley suggests automatically happen after we merge the concept exercise once it's ready?

PR #1584, opened 2021-12-26
Reordered task instructions in Elyses Enchantments exercise

nasch92:
The exercise "Elyses Enchantments" had instructions that did not match the order in which they appeared in the online editor.

I switched the position of the instructions based on the order they had in the code. I also reordered the hints for each task so that they matched the new order.


pertrai1:
LGTM. Thanks for the update @nasch92

junedev (approved):

junedev:
@nasch92 Good catch!

PR #1587, opened 2021-12-28
Improve Elyses Transformative Enchantments

### LINKED ISSUE #1326: Improve Concept Exercise: Elyses Transformative Enchantments (array-transformations)

junedev:
Resolves #1326 (the remaining todos)

Note that the exercise might have some ambiguities currently to whether the original array can be modified or not. For some cases, like when we expect the student to use `map`, I added a test case to clarify the expected behavior. For other cases it is left up to the student for now. Let's see how this is received first. If a lot of students are confused, we can still add more clarification later.

### DIFF
--- a/exercises/concept/elyses-transformative-enchantments/.docs/introduction.md
+++ b/exercises/concept/elyses-transformative-enchantments/.docs/introduction.md
@@ -5,9 +5,9 @@ These methods make it much easier to convert one array into another than it othe
 
 Some methods are [_pure_][pure-function-definition], meaning that they no not modify the original array.
 They return a new one instead.
-Other methods however manipulate the array they are called on and not **not** return the modified array.
+Other methods however manipulate the array they are called on and do **not** return the modified array.
 
-Here is a presentation of some of the most commonly used methods.
+Some of the most commonly used methods to transform arrays are presented below.
 A full list of available methods can be found in the [MDN documentation][array-methods].
 
 ### map (pure)
@@ -117,7 +117,7 @@ arr.splice(2, 0, '3');
 console.log(arr);
 // => ['1', '2', '3', '5', '6']
 
-// Remove 2 elements, starting at index 3 an insert 2 elements
+// Remove 2 elements, starting at index 3 and insert 2 elements
 const removed = arr.splice(3, 2, '4', '5');
 console.log(removed);
 // => ['5', '6']
--- a/exercises/concept/elyses-transformative-enchantments/.docs/introduction.md
+++ b/exercises/concept/elyses-transformative-enchantments/.docs/introduction.md
@@ -1,23 +1,181 @@
 # Introduction
 
-In JavaScript, the `Array` class has many powerful built-in functions for transforming arrays. These functions make it much easier to do things than it otherwise would be using a simple for loop or more direct manipulation.
+In JavaScript, the `Array` class has many powerful built-in methods for transforming arrays.
+These methods make it much easier to convert one array into another than it otherwise would be using a simple for loop or more direct manipulation.
 
-Example of transformation using a for loop and second array:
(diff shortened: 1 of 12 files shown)


exercises/concept/elyses-transformative-enchantments/.docs/introduction.md
```diff
@@ -1,23 +1,181 @@
 # Introduction
 
-In JavaScript, the `Array` class has many powerful built-in functions for transforming arrays. These functions make it much easier to do things than it otherwise would be using a simple for loop or more direct manipulation.
+In JavaScript, the `Array` class has many powerful built-in methods for transforming arrays.
+These methods make it much easier to convert one array into another than it otherwise would be using a simple for loop or more direct manipulation.
 
-Example of transformation using a for loop and second array:
+Some methods are [_pure_][pure-function-definition], meaning that they no not modify the original array.
+They return a new one instead.
+Other methods however manipulate the array they are called on and not **not** return the modified array.
```
pertrai1, line 8: Should this have **not** listed twice?

exercises/concept/elyses-transformative-enchantments/.docs/introduction.md
```diff
@@ -1,23 +1,181 @@
...
+
+Here is a presentation of some of the most commonly used methods.
```
pertrai1, line 10: Should there be a link to the presentation

exercises/concept/elyses-transformative-enchantments/.docs/introduction.md
```diff
@@ -1,23 +1,181 @@
...
+
+```javascript
+const arr = ['1', '2', '5', '6'];
+
+// Insert an element at index 2
+arr.splice(2, 0, '3');
+console.log(arr);
+// => ['1', '2', '3', '5', '6']
+
+// Remove 2 elements, starting at index 3 an insert 2 elements
```
pertrai1, line 120: I think **an** should be **and**

exercises/concept/elyses-transformative-enchantments/.docs/introduction.md
junedev: What is meant here is that the methods are presented below. I rephrased the sentence to make this more clear.

exercises/concept/elyses-transformative-enchantments/.docs/introduction.md
junedev: "not not" should be "do not"

junedev:
@pertrai1 Thanks for the review! I fixed the things you mentioned in all the places.

PR #1592, opened 2021-12-29
Fix error in slice example

sumedh-inamdar:
arr.slice(1, 2) should be [2] instead of [1]


junedev (approved):
Good catch, thanks for the fix!

PR #1593, opened 2021-12-31
Update high-score-board.spec.js with expected value

alanmynah:
We expect `2674` on: 
https://github.com/exercism/javascript/blob/main/exercises/concept/high-score-board/high-score-board.spec.js#L90
but update the value to `1337` on:
https://github.com/exercism/javascript/blob/main/exercises/concept/high-score-board/high-score-board.spec.js#L94
that leads to a failing test on the main track. No way around via online editor. The only option is to update the test and submit from local. 

Updating the function to `updateScore` with expected values.


alanmynah:
nvm, it's just me being daft

PR #1594, opened 2021-12-31
Position moveTo() should be move() as per test spec

sumedh-inamdar:
Test spec and Hints specify 'move()' method of Position object instead of 'moveTo()'.


orien (approved):

junedev (approved):
Thanks!

PR #1601, opened 2022-01-01
Update introduction.md

ngl4:
Fixed the typo in the 'reduce (pure)' section to make sure the accumulator is returning the numbers to the current property in the accumulator object. The if statement is updated to 'even' and the else statement is updated to 'odd'.


SleeplessByte (approved):
Thanks for fixing this :)

PR #1606, opened 2022-01-04
Improve instructions for task 3

### LINKED ISSUE #1604: Why does Exercism test for state mutation?
In JS track Bird Watcher exercise Test 9 is testing for object equality. The only way for me to pass the test is to mutate the array. My initial solution where I returned a new array with the modified value triggers this test failure. It means that my map solution was not accepted. Is this intended behaviour?

```
const birdsPerDay = [2, 0, 1, 4, 1, 3, 0];
expect(Object.is(fixBirdCountLog(birdsPerDay), birdsPerDay)).toBe(true);
```

```
Error: expect(received).toBe(expected) // Object.is equality
Expected: true
Received: false
```

junedev:
Fixes #1604


pertrai1:
👍 LGTM

pertrai1 (approved):

SleeplessByte (approved):
Yeah, this makes sense to me.

PR #1610, opened 2022-01-05
Added new exercise: Satellite

### LINKED ISSUE #924: Add new exercise: satellite
[problem specification ](https://github.com/exercism/problem-specifications/tree/master/exercises/satellite)| [description](https://github.com/exercism/problem-specifications/blob/master/exercises/satellite/description.md) | [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/satellite/canonical-data.json)

You may find it helpful to follow [this checklist](https://github.com/exercism/javascript/blob/83feda2af955df1ef9f3a8080338f506f0fdf696/CONTRIBUTING.md#new-exercise) from the [CONTRIBUTING.md](https://github.com/exercism/javascript/blob/83feda2af955df1ef9f3a8080338f506f0fdf696/CONTRIBUTING.md) resource.

lpizzinidev:
closes #924

### DIFF
--- a/config.json
+++ b/config.json
@@ -1770,7 +1770,13 @@
         "name": "Satellite",
         "uuid": "bfc08dea-2a6b-4696-be20-9b55e282388d",
         "practices": [],
-        "prerequisites": ["arrays", "array-transformations", "objects", "conditionals", "recursion"],
+        "prerequisites": [
+          "arrays",
+          "array-transformations",
+          "objects",
+          "conditionals",
+          "recursion"
+        ],
         "difficulty": 6,
         "topics": []
       }
--- a/config.json
+++ b/config.json
@@ -1764,6 +1764,15 @@
         ],
         "difficulty": 5,
         "topics": ["algorithms", "logic"]
+      },
+      {
+        "slug": "satellite",
+        "name": "Satellite",
+        "uuid": "bfc08dea-2a6b-4696-be20-9b55e282388d",
+        "practices": [],
+        "prerequisites": ["arrays", "sets", "objects", "conditionals"],
+        "difficulty": 6,
+        "topics": ["algorithms", "trees", "recursion"]
       }
     ]
   },
--- a/config.json
+++ b/config.json
@@ -1770,9 +1770,9 @@
         "name": "Satellite",
         "uuid": "bfc08dea-2a6b-4696-be20-9b55e282388d",
         "practices": [],
-        "prerequisites": ["arrays", "sets", "objects", "conditionals"],
+        "prerequisites": ["arrays", "array-transformations", "objects", "conditionals", "recursion"],
         "difficulty": 6,
-        "topics": ["algorithms", "trees", "recursion"]
+        "topics": []
       }
     ]
   },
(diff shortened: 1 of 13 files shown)


config.json
```diff
@@ -1764,6 +1764,15 @@
         ],
         "difficulty": 5,
         "topics": ["algorithms", "logic"]
+      },
+      {
+        "slug": "satellite",
+        "name": "Satellite",
+        "uuid": "bfc08dea-2a6b-4696-be20-9b55e282388d",
+        "practices": [],
+        "prerequisites": ["arrays", "sets", "objects", "conditionals"],
```
junedev, line 1773: * `sets` and nice here but not needed as prerequisite I think. A student could also use some other way to check for duplicates.
* I would add `array-transformations` to the list as I think those are needed to a reasonable solution (e.g. like yours).
* What do you think about adding `recursion`? Seems like a good fit to me.

config.json
```diff
@@ -1764,6 +1764,15 @@
...
+        "difficulty": 6,
+        "topics": ["algorithms", "trees", "recursion"]
```
junedev, line 1775: FYI: There is no need to add the topics key. It is not used anymore. Once we filled all prerequisites and practices fields, all the topics keys will be removed.

junedev (commented):
Thanks for another nice exercise. Sorry for the wait, I finally got around to do the review.

exercises/practice/satellite/satellite.spec.js
```diff
@@ -0,0 +1,51 @@
...
+    };
+    expect(treeFromTraversals(preorder, inorder)).toEqual(expected);
+  });
+
+  test('Reject traversals of different length', () => {
+    const preorder = ['a', 'b'];
+    const inorder = ['b', 'a', 'r'];
+    expect(() => {
+      treeFromTraversals(preorder, inorder);
+    }).toThrowError('traversals must have the same length');
```
junedev, line 33: * From the documentation, it looks the code you wrote means the error message somewhere contains the provided string. We want to check the error/message matches exactly. See last box in this section for how to set this up: https://jestjs.io/docs/expect#tothrowerror
* In most of the other tests we use the `toThrow` version instead of `toThrowError` (they do the same thing), maybe it makes sense to stick with that.

junedev: Sorry, there was a misunderstanding here. With the link to the documentation, I just meant to go from
`toThrowError('traversals must have the same length');`
which only checks the message is contained somewhere in the error message to
`toThrow(new Error('traversals must have the same length'));`
which checks the error messages are actually the same

I didn't see there was a completely different example with custom error classes at the end of that example box. Sorry for not being more precise here earlier.

lpizzinidev: No problem.
I should have asked for clarifications before implementing the code.
I'll fix the issue and submit the changes.

junedev (approved):

SleeplessByte:
*Special space satellite*
Practice arrays, conditionals, objects, and recursion, when trying to transmit a binary tree to a satellite approaching Alpha Centauri.

PR #1614, opened 2022-01-09
Bird Watcher: Add clarification that for loop should be used

junedev:
Besides some people just hating for loops, it also seems a lot of students are not aware that for loops are the concept "under practice" here. For example because they skip over the introduction or in general because people don't have a good understanding of what learning exercises are, some just seem to treat them as "oh, a new practice exercise in the JS track".

I realized that the instructions currently don't mention that a for loop should be used at all. Since the tasks itself don't really force that either, I added a sentence to the instructions to clarify this. Maybe this will help.


SleeplessByte (approved):

SleeplessByte:
Perhaps use an admonition instead to make it stand out.

PR #1615, opened 2022-01-09
Fixes grammar: replaces "a used vehicle" by "an used vehicle"

vibraniumdev:
Replaces all "a used vehicle" by "an used vehicle"


junedev:
The rule here is a bit confusing but the current version is correct. You can find an explanation here: https://www.quora.com/Should-the-word-a-or-an-be-used-before-the-word-unique

vibraniumdev:
Well observed, the "u" sound in "unique" and "used" is called a semivowel since it's a rising diphthong. It makes sense. Thanks!

PR #1616, opened 2022-01-09
Adds a test case that can be solved using `hasOwnProperty` to the High Score Board exercise

jamespeilunli:
The `README.md` in the High Score Board exercise mentions `hasOwnProperty`, but it was never used in the rest of the exercise, so I added a test case that can be solved using `hasOwnProperty`. 

This is my first PR to this repository, so if I'm missing something, let me know. I'm not clear about what may or may not need to be updated on the problem-specifications repository, or in the hidden directories (like `.meta`).

### DIFF
--- a/exercises/concept/high-score-board/.docs/instructions.md
+++ b/exercises/concept/high-score-board/.docs/instructions.md
@@ -110,3 +110,20 @@ const params = { score: 400, normalizeFunction: normalize };
 normalizeScore(params);
 // => 810
 ```
+
+## 7. Confirm player is on score board
+
+Players want to know if they are on the score board.
+
+Write a function `hasPlayer`.
+The function takes a scoreboard, and a name of a player. It should return a boolean indicating if that player is on the score board.
+
+```javascript
+const scoreBoard = {
+  'Amil Pastorius': 99373,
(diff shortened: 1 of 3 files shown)


junedev (commented):
@jamespeilunli Thanks for bringing this up. Usually, it is better to open an issue and discuss your idea with the maintainers before creating a PR.

There are two main reasons this is currently not part of the exercise.
* The exercise already has 6 tasks and it is not necessary that every part of the concept needs to be covered by the exercise. So I decided to leave out this very simple task to not make the exercise too long.
* To properly understand the difference between `hasOwnProperty` vs. `in` or just checking the value for the key is not `undefined`, the student would need the knowledge that is covered in the `classes` concept. This concept comes much later in the syllabus because it depends on some other concepts that need to be tackled first. So `hasOwnProperty` is mentioned for completeness but I don't like the student to practice without the proper understanding of what it does.

We will have an "advanced objects" concept at some point where this topic could be properly deep-dived because that would come after `classes`.

Let me know what you think now that you have some background on the topic.

jamespeilunli:
All right, thanks for letting me know. I'll close this now.

PR #1622, opened 2022-01-14
Elyses Transformative Enchantments exercise typos

FordSmh:
Tests and hints use array with even amount of cards, yet `sandwichTrick` function phpdoc says _deck with 10 cards_ . Probably a copypaste typo


junedev (approved):
Good catch! Thanks for the fix.

Issue #985, opened 2021-02-15
[v3] Add prerequisites 4/6


TomPradat:
> ⚠️ If you have issues / are unsure, create ONE PR PER EXERCISE so it's easier to merge them as we can have discussions per exercise. If you're certain, it's fine to do a single one.

The idea is to split #960 into smaller chunks.

Here is the list of practices exercises to deal with for this part :

```
proverb

twelve-days

yacht

beer-song

resistor-color-trio

dominoes

say

diamond

rational-numbers

sublist

binary-search-tree

custom-set

binary-search

circular-buffer

simple-linked-list

word-search

variable-length-quantity

two-bucket

alphametics

connect
```

pertrai1:
I would like to take this one

junedev:
@pertrai1 Thanks! Please don't do more than 4 or 5 per PR as these are quite difficult to review.

junedev:
Here some hints from someone else who worked on this: https://github.com/exercism/javascript/issues/984#issuecomment-984682274

PR #1623, opened 2022-01-15
#985 - Add Prerequisite to Exercises - Part 2

pertrai1:
Add `prerequisites` for the following:

```
dominoes
say
diamond
rational-numbers
sublist
```

relates to #985

### DIFF
--- a/config.json
+++ b/config.json
@@ -1195,7 +1195,7 @@
         "name": "Sublist",
         "uuid": "bb00e97f-8dcf-40ba-b8c2-c3cb72bbaf97",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["array-analysis", "comparison"],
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -1178,7 +1194,7 @@
         "name": "Sublist",
         "uuid": "bb00e97f-8dcf-40ba-b8c2-c3cb72bbaf97",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["array-analysis", "comparison"],
```
junedev, line 1197: Needs `conditionals` as well, I didn't see a solution without an `if` statement. The inputs are arrays so might be good to add `arrays`.

junedev (commented):
Thanks for the new batch. I left my suggestions what to change potentially.

PR #1641, opened 2022-01-27
javascript array concept minor typo

kroitor:


junedev (commented):
Thanks for the fix! Please take note of the first point in the automated response you got. It applies here.

junedev (approved):

Issue #1549, opened 2021-12-02
Bug: Translation Service - Untranslatable is not defined


khendrikse:
Hi! I have noticed through a friend and myself doing this exercise that the error `Untranslatable` that has to be used in step 4 is not defined in the `service.js` file. It is in the `errors.js` file, but it is not coming through. This makes the following error happen when the error is used in step 4: 

```
ReferenceError: Untranslatable is not defined
    at <solution>/service.js:105:30
```

khendrikse:
@junedev I’m afraid this is not fixed by #1550. That was another problem that was so easy to fox I did not make an issue for it :)

junedev:
Oh, sorry. My mistake. 🙈 Somehow I thought there was an issue for that PR as well.

SleeplessByte:
errors.js is likely missing from the appriopiate files / solution / editor / whatever array in the config. This exercise was created before any of that existed cc @junedev

junedev:
To me it looks like it's possible to solve the exercise without knowledge of these specific error classes as they are not used in our example solutions. My guess is that using the errors explicitly is nevertheless a reasonable solution so yes, we should make sure they are available.

@khendrikse It would be great if you could let us know whether you saw this error when solving the exercise locally or on the website and also provide me with the solution that produces that error.

junedev:
The errors file shows up in the editor just fine so that is not the issue here. I checked a couple of community solutions and even though some import the errors, they don't actually use them. I will need some more time to investigate this and determine why the errors are not needed to solve the exercise even though the instructions sound like they are needed (might be an issue with the test cases) and try to reproduce the error reported in this issue.

@khendrikse In case you are still around, would be great if you could give some more details where and how you encountered the issue (see request above).

khendrikse:
Hey @junedev sorry for the long delay! I encountered it while solving the exercise on the website.

This solution triggered the error:

```
/// <reference path="./global.d.ts" />
// @ts-check
//
// The lines above enable type checking for this file. Various IDEs interpret
// the @ts-check and reference directives. Together, they give you helpful
// autocompletion when implementing this exercise. You don't need to understand
// them in order to use it.
//
// In your own projects, files, and code, you can play with @ts-check as well.

export class TranslationService {
  /**
   * Creates a new service
   * @param {ExternalApi} api the original api
   */
  constructor(api) {
    this.api = api;
  }

  /**
   * Attempts to retrieve the translation for the given text.
   *
[...]

PR #1644, opened 2022-01-29
exercises/concept/translation-service imports and a minor typo

kroitor:
Without those imports, the tests will always fail if you attempt to catch those errors by their type (without string comparisons). On the JavaScript track, the concept of importing is not introduced before this exercise, so it makes sense to add those imports into the JS template for less confusion.

Also a minor typo in the instructions.

### DIFF
--- a/exercises/concept/translation-service/.docs/instructions.md
+++ b/exercises/concept/translation-service/.docs/instructions.md
@@ -99,7 +99,7 @@ service.request("jIyajbe'");
 
 ## 4. Fetch a translation, inspect the quality, or request it
 
-Implement the function `premium(text, quality)` for premium users, which fetches a translation, request it if it's not available, and only returns it if it meets a certain threshold.
+Implement the function `premium(text, quality)` for premium users, which fetches a translation, requests it if it's not available, and only returns it if it meets a certain threshold.
 
 - If `api.fetch` resolves, check the quality before resolving
 - If `api.fetch` rejects with `NotAvailable`, _request_ the translation instead
--- a/exercises/concept/translation-service/service.js
+++ b/exercises/concept/translation-service/service.js
@@ -8,6 +8,8 @@
 //
 // In your own projects, files, and code, you can play with @ts-check as well.
 
+import { NotAvailable, Untranslatable } from './errors';
+
 export class TranslationService {
   /**
    * Creates a new service


kroitor:
The failing check is not related to this PR, it is most likely an intermittent dependency issue.

junedev (changes requested):
This is a good point you are making regarding the imports. However, there is an issue with consistency know and also there seems to be some general issue with the exercise. Maybe you can help to shed some light on this.

Currently, the exemplar solution passes all the tests without these imports. Even student solutions import the errors but then don't end up using them.
Examples:
- https://exercism.org/tracks/javascript/exercises/translation-service/solutions/m42dax
- https://exercism.org/tracks/javascript/exercises/translation-service/solutions/Elmaletas

I don't fully understand why this is. Maybe the test cases are not good enough here, maybe not checking for the specific errors is really the better solution, maybe there is some bug related to the importing (see https://github.com/exercism/javascript/issues/1549) that keeps students from using the import. :thinking: 

In any case, if you feel like a good solution to this exercise should import (and use) those errors, then please adjust the exemplar.js file accordingly. Currently the stub and the exemplar don't match up anymore. The exemplar file is shown to the mentors as "reference solution" for concept exercises, that is why it is important it fits.

kroitor:
@junedev i'll do my best to change the exemplar, but, as you've noted, the reference solution doesn't use those imports. This entire exercise was the most confusing so far on my JS track, which is partially due to incomplete instructions, and partially due to those error imports. Without those imports this iteration would not build: 

- https://exercism.org/tracks/javascript/exercises/translation-service/iterations

I think, the tests for this exercise are lacking an edge case, where the user should not call `api.request` if the resource is known to be `Untranslatable` from `api.fetch` right away. If we add that edge case to the test, the existing exemplar will fail, cause it sends an api.request on all failures. Which is why I think the exemplar is wrong, despite that it satisfies the tests.

junedev:
@kroitor Thanks for you advice on what's wrong with the exercise. Given that information, I will merge this PR as is and create a new issue for adding the missing test case and fixing the examplar solution. Feel free to pick up that issue as well in case you are interested in working on it.

The link you posted only works for you but I found the code you wanted to show on your public profile: https://exercism.org/tracks/javascript/exercises/translation-service/solutions/kroitor

junedev (approved):

PR #1646, opened 2022-01-29
Update the two-bucket proof and tests.

### LINKED ISSUE #1551: Two-Bucket has an incorrect test
The number of moves for the 6th test is wrong. 
```javascript
  describe('Measure using bucket one of size 2 and bucket two of size 3', () => {
    xtest('start with bucket one and end with bucket two', () => {
      const twoBucket = new TwoBucket(2, 3, 3, 'one');
      expect(twoBucket.moves()).toEqual(4);
      expect(twoBucket.goalBucket).toEqual('two');
      expect(twoBucket.otherBucket).toEqual(1);
    });
  });
```

The number of moves should be 2:
1. fill the start bucket
2. since the other bucket size == the goal, fill it
3. there is no move 3.

Following that change, the `otherBucket` amount should be 2 (the start bucket is full).

I humbly suggest [my solution][mine] as a reference solution.

This was changed in #1396.

[mine]: https://exercism.org/tracks/javascript/exercises/two-bucket/solutions/glennj

glennj:
The moves() function now returns an object with keys: moves, goalBucket, otherBucket.
This removes the need for goalBucket and otherBucket properties.
I also fixed the "bug" where moves() can return after the 2nd move if the second bucket is the same size as the goal (fixes #1551).

### DIFF
--- a/exercises/practice/two-bucket/.docs/instructions.append.md
+++ b/exercises/practice/two-bucket/.docs/instructions.append.md
@@ -2,7 +2,7 @@
 
 ## Output format
 
-The `moves()` method is expected to return an object with these properties:
+The `solve()` method is expected to return an object with these properties:
 
 - `moves` - the number of bucket actions required to reach the goal
   (includes filling the start bucket),
--- a/exercises/practice/two-bucket/.docs/instructions.append.md
+++ b/exercises/practice/two-bucket/.docs/instructions.append.md
@@ -10,10 +10,11 @@ The `moves()` method is expected to return an object with these properties:
 - `otherBucket` - the amount contained in the other bucket.
 
 Example:
+
 ```json
 {
-  moves: 5,
-  goalBucket: 'one',
-  otherBucket: 2
+  "moves": 5,
+  "goalBucket": "one",
+  "otherBucket": 2
 }
 ```
--- a/exercises/practice/two-bucket/.docs/instructions.append.md
+++ b/exercises/practice/two-bucket/.docs/instructions.append.md
@@ -1,0 +1,19 @@
+# Instructions.append
+
+## Output format
+
+The `moves()` method is expected to return an object with these properties:
+
+- `moves` - the number of bucket actions required to reach the goal
+  (includes filling the start bucket),
+- `goalBucket` - the name of the bucket that reached the goal amount,
+- `otherBucket` - the amount contained in the other bucket.
+
+Example:
+```json
+{
+  moves: 5,
+  goalBucket: 'one',
+  otherBucket: 2
+}
+```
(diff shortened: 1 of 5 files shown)


junedev (commented):
Thanks for picking up the issue!

I didn't look at the code in detail yet but noticed two things:
* You mentioned before that the tests in problem spec for this exercise were updated so I was expecting this PR would include a change to the tests.toml file due to the new sync via configlet. :thinking:
* I saw you changed the output format for the exercise. I think it is definitely better then before.
  * I was wondering whether the stub really needs to contain a class definition with the new changes you made. Would it make sense that the stub is just a function (e.g. `solve`) that returns the result object you defined? Then the student would be free too choose what classes they think they need in the implementation. (You don't have to change anything if you don't want to  but I would like to hear your thoughts.)
  * Whatever we decide for, the format of the input and expected output should be described in the instructions.append.md file so the student does not have to guess the format from the tests. You can also create an issue for this and it can be addressed later.

glennj:
> * You mentioned before that the tests in problem spec for this exercise were updated so I was expecting this PR would include a change to the tests.toml file due to the new sync via configlet. 🤔

There was no changes necessary to tests.toml. This PR is actually reflecting the actual problem-spec: https://github.com/exercism/problem-specifications/blob/87cd20ff0b2f078d5eb2da716dc7d74fb71ea258/exercises/two-bucket/canonical-data.json#L85


> * I saw you changed the output format for the exercise. I think it is definitely better then before.

Based on @joshgoebel's suggestion. It certainly simplifies things.

>   * I was wondering whether the stub really needs to contain a class definition with the new changes you made. Would it make sense that the stub is just a function (e.g. `solve`) that returns the result object you defined? Then the student would be free too choose what classes they think they need in the implementation. (You don't have to change anything if you don't want to  but I would like to hear your thoughts.)

I don't have strong opinions about it. I think this problem naturally points to an OO solution, so I'm fine with what's there. I didn't want to make overly large changes to what was there already.

>   * Whatever we decide for, the format of the input and expected output should be described in the instructions.append.md file so the student does not have to guess the format from the tests. You can also create an issue for this and it can be addressed later.

I see something similar for the satellite exercise. I'll add it to this PR.

junedev:
> I think this problem naturally points to an OO solution, so I'm fine with what's there.

@glennj Thanks for your feedback. Then let's keep it as is. One small thing though, what do you think about renaming the `moves` method to `solve`? Currently `moves` returns a result object which then has `moves` again as property. That seems a bit strange.

junedev (approved):
I think this is a great improvement for the exercise, not only the bug regarding the tests but also the proper, document output format. Thanks a lot!

PR #1647, opened 2022-01-30
add .gitignore inside of all exercises folder were are missing

herminiotorres:


junedev:
I assume you are proposing this change because you want to commit your solutions to git/GitHub and don't want these files/folders included.

This change would mean all students that download the exercise would also download the .gitignore file even though a lot of them might not use version control for their solutions at all, some students might not even know what git is yet. Because of this, I don't think this PR is the right way to solve the problem above.

You can solve the problem on your end in various ways. If you have all your JavaScript exercises in one folder that you track in git, you can add a .gitignore file at the root level of that directory. If that doesn't work for you for some reason, you can also set up a global gitignore file for your machine. The setup might depend on your operating system, here are instructions how it works for Linux and Mac: http://egorsmirnov.me/2015/05/04/global-gitignore-file.html.

NobbZ:
On top of what junedev said, you can even specify a "global" gitignore file which is only effective for repositories within the  exercism workspace.

```ini
[includeIf "gitdir:~/exercism/**"]
    path = "~/.config/git/exercism.gitconfig"
```

and then set the global ignore file in that secondary configuration.

PR #1648, opened 2022-01-31
Docfix(amusement-park): Fix ticketStatus function wording

rawalplawit:
The function name on question no.4 example should be simpleTicketStatus instead of ticketStatus.


junedev (approved):
Great catch!

PR #1652, opened 2022-02-02
Small fix: Removed unnecessary export from Zebra Puzzle exercise

lpizzinidev:


junedev:
I totally missed that in the initial review. Thanks for the fix!

PR #1655, opened 2022-02-02
exercises/concept/translation-service edits

### LINKED ISSUE #1649: Translation Service: Missing test case
From @kroitor in #1644:

> I think, the tests for this exercise are lacking an edge case, where the user should not call api.request if the resource is known to be Untranslatable from api.fetch right away. If we add that edge case to the test, the existing exemplar will fail, cause it sends an api.request on all failures even when it already knows that the resource is Untranslatable. Which is why I think the exemplar is wrong, despite that it satisfies the tests.

The alternative to fixing the tests and exemplar solution would be to change the instructions. My current feeling is there is value in following the current instructions though.

kroitor:
fix #1649


junedev (approved):

junedev:
Thanks for taking care of this right away!

PR #1658, opened 2022-02-03
Fix broken footnote link

rijavecb:
The footnote on line 27 wasn't clickable

image


SleeplessByte (approved):
We didn't support footnotes when this was written! 

Thanks for the fix :)

PR #1660, opened 2022-02-05
minor grammar fix#3

k0ne:


SleeplessByte:
This is actually not incorrect. You can verify that by changing the word order: 

> in common year (2015), year not divisible by 4

This particular exercise, like most in the `practice` folder, relies on shared problem specifications, found here:
<https://github.com/exercism/problem-specifications/blob/main/exercises/leap/canonical-data.json>

If you'd like to propose a change, it must be done in that repo first. If approved, I'm happy to re-open this and have it applied!

PR #1661, opened 2022-02-05
unreachable [0] index in the 1st example

k0ne:
Line 13 using Math.ceil() in that manner never generates 0, so that the value '2' which is sampleData[0] remains unreffered.
This is a confusing randomization result.
One of the possible solutions (not quiet sure if suitable) is:
let randomNumber = Math.floor(Math.random() * (sampleData.length + 1));
The problem appears in Translation Service exerсise as well.


SleeplessByte (approved):
I like this change (I'm not a big fan of the entire document -- which I didn't author, which could use a rewrite in general).

Thanks!

PR #1662, opened 2022-02-05
[ exercises/concept/fruit-picker ] Add missing return keyword in applyToSquare function

SnowJambi:
I was struggling with understanding callbacks so tried the first example in the introduction in a code playground only to find that `applyToSquare(areaOfSquare);` actually returns `undefined` rather than the expected `25`. Added the missing return keyword so it works as expected (unless I am misunderstanding something?)


SleeplessByte (approved):
You understood correctly. 

I think the author of this concept initially wrote something else that did not return from the callback, but this example most definitely does.

PR #1663, opened 2022-02-05
Upgrade to ESLint 8

SleeplessByte:
This upgrades the entire track to ESLint 8.

If you're having trouble locally (when contributing) with an `eslint` plugin (like in VS Code) and you've already:
- tried turning it off and on again
- tried re-installing it (to force the latest version)

Run the following:
```shell
rm -rf node_modules
npm i
```


junedev:
CI wasn't triggered here again (so it does not seem to be related to the format action), will close and open.

junedev (approved):

PR #1664, opened 2022-02-06
Use @exercism/babel-preset-javascript

SleeplessByte:
This uses `@exercism/babel-preset-javascript` for the babel configuration.

- Simplifies the dependency list
- Simplifies the babel configuration
- We can consolidate "updating" babel packages in that repo (once it's up), so it doesn't pollute this one anymore.

Technically this also opens up `@exercism/babel-plugin-javascript`, should we ever want some custom shit, including injecting stuff for the test-runner.


junedev:
@SleeplessByte Where is the code for `@exercism/babel-preset-javascript`? I can't find an Exercism repo with this name. I know last time you also publishes something before Jeremy created the repo but I can't remember where you kept the code.

SleeplessByte:
I don't have a repo yet!

You can pull the file directly from npm. It's just a single file.

junedev (commented):
I have some more questions to be sure this works.

1. The code in `@exercism/babel-preset-javascript` contains presets and an empty plugins array and then the babel.config.js file contains `presets` and and empty plugins array again. Is that correct/intended?

2. When corejs was updated in the past, you said we also need to update the babel.config.js file. Does this mean after this change e.g. when @tejasbubane want's to merge the update, he would need to update the new npm package first (which is currently impossible for him) and then include the updated version in the repo and update the corejs version in the repo?

3. I thought we had CI to check that the corejs version in the babel.config.js fits the one in the package.json. Probably this CI check needs to be changed/removed now (even if it might not fail if the corejs key can't be found.

junedev:
CI did not run for this, I will try close and re-open.

SleeplessByte:
Good questions! 

1. This has several parts. The empty array is intended. I wanted to be explicit that we currently don't load any plugins directly in our preset, nor do we directly in our config. We used to have the big int plugin here, but it made it to stage 3 (and 4?) And thus was included in preset env, which is a transitive preset we load! 

2. Yes. The idea is that that change will come in on the repo that's yet to be created. The check will indeed need to be added to that repo. The step is then to merge the dep in that repo, and then trigger a publish of the package. Publishing the package will make dependabot ask this repo to be updated (same as now). Publishing the package can be a GitHub action which the three of us could trigger.

3. Yes.

junedev (approved):

PR #1668, opened 2022-02-06
Add Prerequisites Part 4

### LINKED ISSUE #985: [v3] Add prerequisites 4/6

pertrai1:
Final exercises to add prerequisites for to resolve #985

### DIFF
--- a/config.json
+++ b/config.json
@@ -1295,7 +1295,13 @@
         "name": "Word Search",
         "uuid": "0125f5a7-8883-4553-a6c1-45f19544af5e",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "arrays",
+          "array-loops",
+          "array-transformations",
+          "classes",
+          "conditionals"
+        ],
         "difficulty": 8,
         "topics": [
           "arrays",
@@ -1312,7 +1318,12 @@
         "name": "Variable Length Quantity",
         "uuid": "6be39d1f-f68d-4a52-bd72-85a38ffd509e",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "array-transformations",
+          "arrow-functions",
+          "functions",
+          "rest-and-spread"
+        ],
         "difficulty": 5,
         "topics": ["bitwise_operations", "transforming"]
       },
@@ -1321,7 +1332,12 @@
         "name": "Two Bucket",
         "uuid": "0bc6b478-40a8-47ab-889b-c403b922f7e5",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "array-analysis",
+          "classes",
+          "conditionals",
+          "functions"
+        ],
         "difficulty": 6,
         "topics": [
           "algorithms",
@@ -1338,7 +1354,13 @@
         "name": "Alphametics",
         "uuid": "a602bd33-69fc-4b67-a3d3-95198853095e",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "array-analysis",
+          "array-transformations",
+          "conditionals",
+          "for-loops",
+          "functions"
+        ],
         "difficulty": 7,
         "topics": ["algorithms", "games"]
       },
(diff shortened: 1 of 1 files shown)


junedev:
Sorry for the wait regarding the review here. Didn't really have the time/energy this week.

config.json
```diff
@@ -1295,7 +1295,13 @@
         "name": "Word Search",
         "uuid": "0125f5a7-8883-4553-a6c1-45f19544af5e",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "arrays",
+          "array-loops",
+          "array-transformations",
```
junedev, line 1301: I don't think transformations are needed/recommended to solve this, e.g. have a look at this solution: https://exercism.org/tracks/javascript/exercises/word-search/solutions/James-Deciutiis

config.json
```diff
@@ -1312,7 +1318,12 @@
         "name": "Variable Length Quantity",
         "uuid": "6be39d1f-f68d-4a52-bd72-85a38ffd509e",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "array-transformations",
+          "arrow-functions",
+          "functions",
+          "rest-and-spread"
```
junedev, line 1325: * `errors` is missing
* most solutions I saw use `for-loops`
* I would recommend to add `arrays` as well because that is where we teach push etc which is often used here
* I don't see the reason why `functions` is mentioned here
* I would recommend to add `bitwise-operators` even though we don't have the concept yet but they seem to be often used here. (All the bugs with exercises not unlocking because of something like this have been fixed now.)

config.json
```diff
@@ -1321,7 +1332,12 @@
         "name": "Two Bucket",
         "uuid": "0bc6b478-40a8-47ab-889b-c403b922f7e5",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "array-analysis",
+          "classes",
+          "conditionals",
+          "functions"
```
junedev, line 1339: For this one I would go by the current example solution. It was recently updated and is rather readable. There I see classes, arithmetic-operators, errors, conditionals, while-loops and objects (for the return value).

config.json
```diff
@@ -1338,7 +1354,13 @@
         "name": "Alphametics",
         "uuid": "a602bd33-69fc-4b67-a3d3-95198853095e",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "array-analysis",
+          "array-transformations",
+          "conditionals",
+          "for-loops",
+          "functions"
```
junedev, line 1362: Can you show me a solution that works with just those? Going by the example solution I would add `rest-and-spread`, `arithmetic-operators` and `objects`.

config.json
```diff
@@ -1347,7 +1369,12 @@
         "name": "Connect",
         "uuid": "2fa2c262-77ae-409b-bfd8-1d643faae772",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "array-analysis",
+          "array-transformations",
+          "functions",
+          "for-loops"
```
junedev, line 1376: * classes and conditionals are missing
* I would recommend array-loops instead of for loops here because the iteration is over elements of the array
* arrays would probably also be a good fit here

junedev (commented):
After a very busy time at work I am back and finished the review.

junedev:
CI is buggy again, I need to close and re-open the PR.

junedev (approved):

PR #1670, opened 2022-02-06
Add custom flags to config

SleeplessByte:
Supports https://github.com/exercism/javascript-test-runner/pull/79.

This adds hand-written flags to the `.meta/config.json` files to be consumed by our tooling.

```json
"version.tests.compatibility": "jest-27"
```

This doesn't do anything by itself, but jest 27 is technically a breaking change. Adding this to our config files allows us to:

- separate exercises that have been updated vs. those that are on pre-27 versions
- give special error messages if something timeout related happens. Why? Jest 27 imposes a standard 5s timeout on each individual `test`.
- track unupdated exercises the next time jest is breaking

```json
"flag.tests.task-per-describe": true
```
This should generally be `true` for all `concept` exercises and `false` for all `practice` exercises. When `true`, the test-runner can calculate the `task_id` based on the `describe` in the test files. This PR also contains updates to the spec files to match this newly imposed format.

```json
"flag.tests.may-run-long": false
```
When we see students complain about timeouts on naive implementations, we should turn this to true. Doesn't do something right now, but if we see it a lot, we can use it to actually impose a _shorter_ timeout (in the runner) and then show guided help.

```json
"flag.tests.includes-optional": false
```
When true, skipped tests with `test.skip` are present and optional. This allows us to:
- enforce in the test-runner that there are only skipped tests if this is true
- add a message in the analyzer that there are optional tests that the student can run via the CLI. We can add an informational message with data how to do that, even if they did run this via the CLI.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,3 +2,6 @@
 /.github/workflows/sync-labels.yml
 exercises/**/README.md
 !/README.md
+
+# These are formatted via configlet and will not match prettier
+exercises/**/.meta/config.json
--- a/exercises/concept/amusement-park/.meta/config.json
+++ b/exercises/concept/amusement-park/.meta/config.json
@@ -7,5 +7,11 @@
     "test": ["amusement-park.spec.js"],
     "exemplar": [".meta/exemplar.js"]
   },
-  "forked_from": ["ruby/amusement-park"]
+  "forked_from": ["ruby/amusement-park"],
+  "custom": {
+    "version.tests.compatibility": "jest-27",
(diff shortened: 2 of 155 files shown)


junedev (approved):
Skimmed this. LGTM

SleeplessByte:
Yeah, I want to add it there, and also in our CONTRIBUTING guide, right?

SleeplessByte:
I am kinda leaning towards adding everything to the `exercism.org/` javascript docs.

PR #1671, opened 2022-02-07
Make translation service more resilient against bad implementations

SleeplessByte:
With this change, a faulty implementation when calling the API results in the respective test failing with the following message:

```text
Error: Expected callback function when calling fetch(text, callback), actual undefined.
    at new BadRequest (<solution>\api.js:122:5)
    at ExternalApi.request (<solution>\api.js:76:13)
    at TranslationService.request (<solution>\service.js:61:21)",
```

Without this change, the entire test run blows up and returns:

```json
{
  "status": "error",
  "message": "Expected to run at least one test, but none were found. This can happen if the test file(s) (.spec.js) are missing or empty. These files are normally not empty. Revert any changes or report an issue if the problem persists.",
  "tests": [],
  "version": 3
}
```

Because under the hood it shows:

```text
Determining test suites to run...<solution>\api.js:97       
      setTimeout(() => callback(this.values[text][0] ? undefined : makeRandomError()), 1);
                       ^

TypeError: callback is not a function
    at Timeout._onTimeout (<solution>\api.js:86:15)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7)
```

### DIFF
--- a/exercises/concept/translation-service/api.js
+++ b/exercises/concept/translation-service/api.js
@@ -39,6 +39,12 @@ export class ExternalApi {
    * @returns {Promise<Translation>}
    */
   fetch(text) {
+    if (typeof text !== 'string') {
+      throw new BadRequest(
+        `Expected text when calling fetch(text), actual ${typeof text}.`
+      );
+    }
+
     // Check if client is banned
     if (mutex.current) {
       return rejectWithRandomDelay(new AbusiveClientError());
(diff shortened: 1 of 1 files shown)


junedev (approved):
I would recommend you move all errors into api.js as discussed. Otherwise it seems a bit weird now that BadRequest is in api.js but the others are in errors.js.

But we can also do this later of course, so I'll approve anyway and you can decide how to proceed.

Issue #1666, opened 2022-02-06
For practice exercises, all but the first test case should be marked as skipped


junedev:
This issue can be worked on by someone without any experience in contributing to Exercism. (Experienced contributors, please leave this for someone else.) 

In the test files of the practice exercises, our rule is that the test cases should be marked as skipped and only the first one should be active (un-skipped). The test files are all the `<exercise-slug>.spec.js` files in the folder `exercises/practice/<exercise-slug>`.

Skipping the test case is done by using `xtest` instead of `test` or `xit` instead of `it`.

Most test files should already follow this rule but we forgot to apply this to some newer exercises. It would be great to get this sorted. The reason all but the first test is skipped is that the student is supposed to work through the exercise TDD style, step-by-step (get one test passing, unskip the next one, get that one passing etc.).

If you are new to making contributions via GitHub, here is a step by step guide: https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github

Sidenote: This task is labelled as "size:medium" not because it is hard, just because it is a bit tedious.

pertrai1:
@junedev - if no first timers come forward to work on this, I can pick it up to get the test consistent across all exercises.

flawnn:
I'll work on that!

flawnn:
https://github.com/exercism/javascript/pull/1672

flawnn:
Just pushed the PR!

PR #1672, opened 2022-02-07
Fix up test case checks

flawnn:
In the test files of the practice exercises, our rule is that the test cases should be marked as skipped and only the first one should be active (un-skipped).
The test files are all the <exercise-slug>.spec.js files in the folder exercises/practice/<exercise-slug>.


flawnn:
Fixing up https://github.com/exercism/javascript/issues/1666

PR #1677, opened 2022-02-12
Add tips for function naming

Aung1998:
Added improvements to `concepts/functions/introduction.md` and `exercises/concept/lasagna-master/.docs/introduction.md`, part of [issue #1234](https://github.com/exercism/javascript/issues/1234)

- Added naming function section.


junedev (commented):
I am sorry you had to wait so long for feedback on this, I had a busy time at work.

I discussed with another maintainer, how to best deal with the new content you provided here because it is very general, it would even apply for other languages than JavaScript.
Nevertheless we think it is valuable content.

The best solution we found was that you create a new file `PROGRAMMING_TIPS.md` in the docs folder of this repo and move your content there. Then we can link to that from the functions concept and also collect more similar general content in the future.

To be visible on the website, the new docs page also needs an entry in the config, I will add that for you once the PR is adjusted.

Also I would ask you to double check your bullet points and try to make them as concise as possible. Try not to repeat the same content with different words.

Lastly, we have the convention to create a new line for each sentence. See https://exercism.org/docs/building/markdown/markdown#h-layout for reference.With bullet points that works by adding two spaces at the beginning of the new line. With that, the sentence gets added to the bullet point.

junedev:
@Aung1998 Would it be ok for you if I push a change into your branch to adjust the wording a bit? I could write comments instead so you can make the changes but that doesn't really benefit anyone.

As mentioned before, I will also add the update to the configuration.

Aung1998:
@junedev Sure. I'm ok with it.

RTurek:
@Aung1998 @junedev - I have some re-wording suggestions as well, but this would be my first contribution to Exercism and I'm unfamiliar with contribution best practices. Is it appropriate to push commits, or should I just leave a bunch of comments? I likely have similar thoughts as @junedev as far as re-wording.

junedev:
@RTurek I am sick currently so it would be great if you could help out with improving the wording here. However, only maintainers can add commits. Maybe you could use the GitHub suggestion feature for the whole text (you don't have to go line by line) or post the improved text as a comment here.

junedev:
@RTurek Are you still interested in helping with the rewording here?

RTurek:
@junedev yes I am, I have this tab open still and hope to get to it in the next 2 weeks, I have just been buried under work and family life.

I have it maybe half drafted up. Sorry to be so slow!

docs/PROGRAMMING_TIPS.md
```diff
@@ -0,0 +1,25 @@
...
+  It will return the student grade.
+- Don't make the name too short or too long. 
+  It is good to be clear and precise but take care not to make too long name. 
+  `checkStudentFailOrPassOrMeritOrDistinction(mark)` is an example of a unneccessarily long function name.
+  The longer the name is, the harder it get to remember and the more space it takes on the computer screen.
+  On the other hand. making the name too short will also cause confusion. 
+  `checkFPMD(mark)` is an example of the bad shortened name because when another programmer look at the function, he would get confused by what 'FPMD' mean.
+- Use the name that everyone understand. 
+  Your function name should be understandable by everyone who read your code. 
+  For example, don't use `DeathStar()` when you write delete item function. Use `deleteItem()` instead.
```
minimalsm, line 20:
```suggestion
### Use the name that everyone understands. 

Your function name should be understandable by everyone who reads your code. 
For example, an obscure function name like `DeathStar()` might be amusing to you, but the name doesn't communicate its purpose.

```

docs/PROGRAMMING_TIPS.md
```diff
@@ -0,0 +1,25 @@
...
+- Pick one word per one concept and stick with it. 
+  For example, if you named getStudentName(), don't change get to retrieve like retrieveStudentGrade() when naming another function.
+- Avoid using the same word for two purposes. 
+  For example, don't keep using 'add' when you want to put a group of items to a collection. 
+  Use 'insert' or 'append' instead.
```
minimalsm, line 25:
```suggestion
### Avoid using the same word for two purposes

For example, don't keep using 'add' when you want to put a group of items into a collection. 
Use 'insert' or 'append' instead.
```

minimalsm (changes requested):
Hey folks :-)

I noticed this has been hanging around for a while and I thought I'd take a pop at tidying up the language a little bit.

I also converted each bullet point to headings (I hope this is okay 😀). This makes it easier for readers to scan to get a holistic understanding of the points without having through every point first.

---

FWIW I disagree with the long function names example (it is a bad function name, but not because of the length) but I've left it in.

docs/PROGRAMMING_TIPS.md
minimalsm: I don't think this point makes sense, I wasn't entirely sure what you meant @Aung1998 so I've just fixed a typo and left the rest as is.

Aung1998: Let me give you an example with JavaScript functions. array.push() and array.concat() Why JavaScript programmer use two different words form the two similar functions? Because they are not the same. In other words, don't mistake 'similar' with 'same'.

junedev: Maybe the point can be clarified by putting something like "If you already used `add` for summing up integers, don't use it for ..." in the beginning.

Besides that, I would recommend backticks instead of the single quotes.

Aung1998:
Sure I will do that

docs/PROGRAMMING_TIPS.md
```diff
@@ -0,0 +1,25 @@
...
+  When you name your function, you should be clear on what it does. 
+  On Function name `get(studentId, subject)`, you can't tell what it is geting mark for or what it is getting or whether it is getting Studnet name or Age or their marks. 
+  Function `getStudentGrade(studentId, subject)` told you everything you need to know about what it does. 
+  It will return the student grade.
+- Don't make the name too short or too long. 
+  It is good to be clear and precise but take care not to make too long name. 
+  `checkStudentFailOrPassOrMeritOrDistinction(mark)` is an example of a unneccessarily long function name.
+  The longer the name is, the harder it get to remember and the more space it takes on the computer screen.
+  On the other hand. making the name too short will also cause confusion. 
+  `checkFPMD(mark)` is an example of the bad shortened name because when another programmer look at the function, he would get confused by what 'FPMD' mean.
```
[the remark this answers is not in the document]
Aung1998: @minimalsm if you have a better idea for longer name, you can put a suggestion.

junedev (approved):

junedev:
@Aung1998 Thanks for your patience on this. I'm very sorry it took so long to get this merged. As discussed before, I added the adjustment in the config file.

PR #1678, opened 2022-02-12
Minor fix

qweered:
corresponding https://github.com/exercism/javascript/blob/main/exercises/concept/elyses-enchantments/.docs/instructions.md

### DIFF
--- a/exercises/concept/elyses-enchantments/.docs/instructions.md
+++ b/exercises/concept/elyses-enchantments/.docs/instructions.md
@@ -28,7 +28,7 @@ Return the adjusted stack.
 
 ```javascript
 const position = 2;
-const newCard = 6;
+const replacementCard = 6;
 setItem([1, 2, 4, 1], position, replacementCard);
 // => [1, 2, 6, 1]
 ```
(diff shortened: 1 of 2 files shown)


junedev:
@qweered Can you say a couple of words why you think this change is necessary? `newCard` in the instructions is just a variable name so there is no actual need for the two names to match.

qweered:
@junedev I think it removes any misunderstanding like "what is replacement card?". Cause everywhere in code it names newCard

SleeplessByte:
It would become:

````markdown
## 2. Exchange a card in the stack

Perform some sleight of hand and exchange the card at index `position`
with the replacement card provided.
Return the adjusted stack.

```javascript
const position = 2;
const replacementCard = 6;
setItem([1, 2, 4, 1], position, replacementCard);
// => [1, 2, 6, 1]
````

## 2. Exchange a card in the stack

Perform some sleight of hand and exchange the card at index `position`
with the replacement card provided.
Return the adjusted stack.

```javascript
const position = 2;
const replacementCard = 6;
setItem([1, 2, 4, 1], position, replacementCard);
// => [1, 2, 6, 1]

qweered:
Sounds good

qweered:
i think its ready to merge

SleeplessByte (approved):
Thanks for the change!

PR #1681, opened 2022-02-16
Update queen-attack to be clearer for non-chess players

mahmoud-moursy:
- Added examples
- Added explanation for why the queens are labelled B and W

### DIFF
--- a/exercises/practice/queen-attack/.docs/instructions.md
+++ b/exercises/practice/queen-attack/.docs/instructions.md
@@ -23,5 +23,39 @@ _ _ _ _ _ _ _ _
 ```
 
 You'd also be able to answer whether the queens can attack each other.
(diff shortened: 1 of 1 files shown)


SleeplessByte:
Thank you for your submission. These seem very sensible changes.

The description for this exercise is sourced from https://github.com/exercism/problem-specifications/blob/main/exercises/queen-attack/description.md. Would you be so kind to open a PR in that repo so we can determine if this change is nice for all tracks!

Then, we can come back to this PR and actually pull it from there.

SleeplessByte:
It didn't show up in my notifications list 😅 . 

Let's see what happens there 🗡️

SleeplessByte:
Closed because of https://github.com/exercism/problem-specifications/pull/1965#issuecomment-1528794281

PR #1687, opened 2022-02-18
Added /.idea to .gitignore for JetBrains based IDEs.

Metallifax:
Accommodates contributors who use WebStorm and other JetBrains IDEs, since it will auto-generate environment info. In line with other repos like exercism/python for PyCharm and exercism/java for IntelliJ.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /bin/configlet.exe
 /pnpm-lock.yaml
 /yarn.lock
+/.idea


SleeplessByte:
Hi there!

Thanks for your contribution. Following https://github.com/exercism/javascript/pull/1647#issuecomment-1025199554 I will close this PR for the same reason. We absolutely understand why you'd submit this, and the comment linked provides an alternative way to accomplish the same.

PR #1688, opened 2022-02-18
fixes #1685

### LINKED ISSUE #1685: Elyses Enchantments currently not passable due to errors in enchantments.spec.js
Some of the below tests compare equality between expected and expected. Some tests require mutation in order to pass. Some have both problems.

[test('remove the middle two cards'](https://github.com/exercism/javascript/blob/1b66a23cfa0de7fe9bd2b9d336a6c46c52a9b7e3/exercises/concept/elyses-enchantments/enchantments.spec.js#L145-L154)
[test('remove two cards from the top',](https://github.com/exercism/javascript/blob/1b66a23cfa0de7fe9bd2b9d336a6c46c52a9b7e3/exercises/concept/elyses-enchantments/enchantments.spec.js#L169-L177)
[test('remove two cards from the bottom'](https://github.com/exercism/javascript/blob/1b66a23cfa0de7fe9bd2b9d336a6c46c52a9b7e3/exercises/concept/elyses-enchantments/enchantments.spec.js#L191-L200)
[test('adding multiple cards to the stack to the bottom'](https://github.com/exercism/javascript/blob/1b66a23cfa0de7fe9bd2b9d336a6c46c52a9b7e3/exercises/concept/elyses-enchantments/enchantments.spec.js#L235-L244)

david-abell:
I've updated the tests errors in Elyses Enchantments so that:

- tests no longer check equality between `expect => to be strictly equal to expected` but instead check `result stack => to be strictly equal to expected`
- tests that call one function multiple times to remove multiple cards now set equality on stack variable each time they run so that mutation is not required in order to pass strict equality test

### DIFF
--- a/exercises/concept/elyses-enchantments/enchantments.spec.js
+++ b/exercises/concept/elyses-enchantments/enchantments.spec.js
@@ -191,8 +191,8 @@ describe('make the top card disappear', () => {
   test('remove two cards from the bottom', () => {
     let stack = [1, 2, 3];
 
-    stack = removeItemFromTop(stack);
-    stack = removeItemFromTop(stack);
+    stack = removeItemAtBottom(stack);
+    stack = removeItemAtBottom(stack);
 
     const expected = [3];
     expect(stack).toStrictEqual(expected);
--- a/exercises/concept/elyses-enchantments/enchantments.spec.js
+++ b/exercises/concept/elyses-enchantments/enchantments.spec.js
@@ -97,10 +97,10 @@ describe('make cards appear at the top', () => {
   });
 
   test('adding multiple cards to the stack at the top', () => {
-    const stack = [1];
+    let stack = [1];
 
-    insertItemAtTop(stack, 5);
-    insertItemAtTop(stack, 9);
+    stack = insertItemAtTop(stack, 5);
+    stack = insertItemAtTop(stack, 9);
 
     const expected = [1, 5, 9];
     expect(stack).toStrictEqual(expected);
@@ -143,13 +143,13 @@ describe('make cards disappear', () => {
   });
 
   test('remove the middle two cards', () => {
-    const stack = [1, 2, 3, 4];
+    let stack = [1, 2, 3, 4];
 
-    removeItem(stack, 1);
-    removeItem(stack, 1);
+    stack = removeItem(stack, 1);
+    stack = removeItem(stack, 1);
 
     const expected = [1, 4];
-    expect(expected).toStrictEqual(expected);
+    expect(stack).toStrictEqual(expected);
   });
 });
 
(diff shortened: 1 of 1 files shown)


exercises/concept/elyses-enchantments/enchantments.spec.js
```diff
@@ -97,10 +97,10 @@ describe('make cards appear at the top', () => {
   });
 
   test('adding multiple cards to the stack at the top', () => {
-    const stack = [1];
+    let stack = [1];
 
-    insertItemAtTop(stack, 5);
-    insertItemAtTop(stack, 9);
+    stack = insertItemAtTop(stack, 5);
+    stack = insertItemAtTop(stack, 9);
```
SleeplessByte, line 103: This test is not incorrect.

exercises/concept/elyses-enchantments/enchantments.spec.js
```diff
@@ -189,13 +189,13 @@ describe('make the top card disappear', () => {
...
+    let stack = [1, 2, 3];
 
-    removeItemFromTop(stack);
-    removeItemFromTop(stack);
+    stack = removeItemAtBottom(stack);
+    stack = removeItemAtBottom(stack);
 
     const expected = [3];
-    expect(expected).toStrictEqual(expected);
+    expect(stack).toStrictEqual(expected);
```
SleeplessByte, line 198: Instead of changed the top part, only the last line should be changed (which you did, thanks!)

exercises/concept/elyses-enchantments/enchantments.spec.js
```diff
@@ -233,10 +233,10 @@ describe('make cards appear at the bottom', () => {
   });
 
   test('adding multiple cards to the stack to the bottom', () => {
-    const stack = [1];
+    let stack = [1];
 
-    insertItemAtBottom(stack, 5);
-    insertItemAtBottom(stack, 9);
+    stack = insertItemAtBottom(stack, 5);
+    stack = insertItemAtBottom(stack, 9);
```
SleeplessByte, line 239: This test is not incorrect.

SleeplessByte (changes requested):

david-abell:
Hmm the changes pass both my test answers as well as if I run it against `exemplar.js`. What am I missing?

SleeplessByte:
You're trying to change the tests so that you can use immutability, correct? The exercise is _purposefully_ expecting mutation on some of the tests. 

- `expect(expected).toXXX(expected)` is a mistake, you fixed those (great 😄).
- Changing the mutability is unnecessary. 
- `exemplar.js` should pass without changing the mutability! If it doesn't, we'd like to know :)

david-abell:
Yes I had not realized that immutability was not a goal in these tests. I removed those pieces and it appears to be passing with `exemplar.js` now

PR #1689, opened 2022-02-19
Update windowing-system.spec.js

PauloHenriqueMelo:
Hello. In instructions you drew how the system works. You chose the >upper<  left of the window as the reference point for Window. As such, the maximum value for position.y for the window is the screenSize.height itself. And the minimum value for the position.y is the height of the window. Think about it. For your tests to make sense you would have to have chosen the lower left of the window as the reference point for the positon of the Window.


junedev:
I am not sure I get your point here. I checked the illustration again but I don't see how it makes a statement about in which direction the value increases if that is what you mean. The text for the position states `and y values getting larger as you move down`.
:thinking:

junedev:
I am closing this as it seems abandoned.

PR #1692, opened 2022-02-22
added two words

Atomic-Germ:
I ended up writing the wrong function, just moving the one card to the end.
To make the last exercise clear, I've added "and vice-versa".


Atomic-Germ:
Oh holy cow it's actually worse than I thought.
OKAY. So, when it says "so that the top card ends up at the bottom", you _apparently_ mean "reverse the whole deck".

So. Figure that out, yea?

PR #1698, opened 2022-02-25
Fix wrong test case in Translation Service

Taucher2003:
The first call to `premium` already requested and resolved the translation. Therefore the quality check after requesting was never tested.

This caused code snippets like this to pass the tests, when it shouldn't
```javascript
async premium(text, minimumQuality) {
    try {
      const translation = await this.api.fetch(text);

      if (translation.quality >= minimumQuality) {
        return translation.translation;
      }

      throw new QualityThresholdNotMet(text);
    } catch (error) {
      if (error instanceof NotAvailable) {
        await this.request(text);
        const { translation } = await this.api.fetch(text);
        return translation;
      }

      throw error;
    }
  }
```
(Credits for this snippet to dominguezcelada as part of the mentoring discussion)

The snippet above does not check the quality after requesting it and still passed the test. \
The correct way would be to check the quality also after requesting it, which can be done by recalling the `premium` function because the translation has been requested and resolved.

```javascript
async premium(text, minimumQuality) {
        try {
            const translation = await this.api.fetch(text);
            if (translation.quality >= minimumQuality) {
                return translation.translation;
            }
            throw new QualityThresholdNotMet(text);
        } catch (error) {
            if (error instanceof NotAvailable) {
                await this.request(text);
                return await this.premium(text, minimumQuality);
            }
            throw error;
        }
    }
```

### DIFF
--- a/exercises/concept/translation-service/service.spec.js
+++ b/exercises/concept/translation-service/service.spec.js
@@ -185,7 +185,14 @@ describe('Premium service', () => {
     await expect(actual).rejects.toThrow(Error);
   });
 
-  test('it ensures the quality of the translation', async () => {
+  test('it recognizes sufficient quality', async () => {
+    const actual = service.premium('‘arlogh Qoylu’pu’?', 40);
+    const expected = 'What time is it?';
+
+    await expect(actual).resolves.toBe(expected);
+  });
+
+  test('it recognizes insufficient quality', async () => {
     const actual = service.premium('majQa’', 100);
     const expected = QualityThresholdNotMet;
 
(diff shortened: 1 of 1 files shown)


junedev (approved):

junedev:
Thanks for your contribution and sorry for the long waiting time for feedback.

Instead of completely removing the one part of the test, I moved it into another test as no other case properly covers the case that there is non-zero but sufficient quality.

Two comments regarding your issue description that are not related to the issue itself:
* Please keep in mind that translation service is a learning exercise to learn/practice (or demonstrate your knowledge of) promises. Using async/await defeats that point. Learning exercises are meant to be solved with the concepts they teach. This is different from practices exercises which should be solved the best way you possibly can.
* The example at the bottom is a "dangerous" solution. If fetch would not stop to return `NotAvailable` for some reason, that solution would end up in an infinite loop.

Issue #1600, opened 2022-01-01
Windowing System: Improvement Ideas


junedev:
After seeing a couple of solutions to "Windowing System" during mentoring, I think the exercise could need some changes to better achieve it's learning objectives.

* For `Position` and `Size`, the constructor and the method have the same implementation. I saw a solution that used the method in the constructor which is not a very typical example.
* Since non of the properties are marked as private with the underscore, it's harder to argue that people should use the method instead of assigning values directly.
* The calculation for the window resize and move is rather involved which does not add value for the actual learning objectives but it seems students spend quite some time finding creative solutions there, e.g.
  ![image](https://user-images.githubusercontent.com/12543047/147859416-62818b06-71c5-4558-ba1e-d08cbda753d8.png)

Some ideas how to improve the exercise:
* Maybe we should mark some properties as private with the underscore to discourage the direct assignment. That might also make it easier to pick up unwanted code with the analyzer later. (Not sure how to best test in this case though. Either the test disrespects the underscore which is not a good example or the test needs to fall back on some user-implemented serialization function. But if that function is wrong all the tests will fail which is also annoying as everything else the user did could be correct.)
* The methods on `Position` and `Size` could change the values by some amount instead of just setting a value. That would make them distinct from the constructor. (Probably only needed if we stick with "public properties".)
* Simplify the window resize/move method e.g. by saying some lower bound is always respected

SleeplessByte:
(Maybe related: https://javascript.info/private-protected-properties-methods)

Since this is a concept exercise, we can add some tests to remove some of this behaviour, if you'd like. Technically we _could_ also start pushing the `#private` syntax (even though I hate it) given its current stage in the TC39 process.

junedev:
I read a lot of negative posts about the real private properties lately so I don't feel like pushing them either. Also I prefer to stick to stage 4 as much as possible in the concepts.

junedev:
I would love to fix the exercise but I am not sure how to best approach this. I don't think it would be good to just make the tests enforce the use of the methods instead of direct assignment if there is no actual difference. It makes it very hard to argue the case in mentoring. I would prefer if it would be clear from the type of the problem that the method needs to be used, that some property needs to be "private", etc.

I am happy to make bigger changes to the tasks and story if someone has an idea how to construct the exercise in a way that a desired solution comes more naturally.

Here the minimum I would like to get out of the exercise:
[...]

PR #1699, opened 2022-02-25
Fix misplaced test in Windowing System

massartval:
The test 'resize respects limits due to position and screen size' is misplaced. It should be with the other tests for the 'resize' method, not with the tests for the 'move' method.

### DIFF
--- a/exercises/concept/windowing-system/windowing-system.spec.js
+++ b/exercises/concept/windowing-system/windowing-system.spec.js
@@ -92,11 +92,11 @@ describe('resize', () => {
     expect(programWindow.size.width).toBe(1);
     expect(programWindow.size.height).toBe(1);
   });
-  
-    test('resize respects limits due to position and screen size', () => {
+
+  test('resize respects limits due to position and screen size', () => {
     const programWindow = new ProgramWindow();
-    const newPosition = new Position(710, 525);
-    programWindow.move(newPosition);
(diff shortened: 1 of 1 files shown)


junedev (changes requested):
I am sorry you had to wait for feedback on this, I was quite busy recently.

The original test case tests the interplay of move and resize. It is in the "move" describe group because the student can then get `resize` passing before moving on to the next task.

You removed the `move` call but the new test code is not the desired way to interact with the object here. Students should use the methods. (Sidenote: we are discussing here https://github.com/exercism/javascript/issues/1600 how to improve the exercise to make this more apparent/ required). 

I am happy to accept a change to the test name if you found the current one confusing but the test position and code should stay as is.

exercises/concept/windowing-system/windowing-system.spec.js
```diff
@@ -92,9 +92,20 @@ describe('resize', () => {
...
+    const newSize = new Size(1000, 1000);
+    programWindow.resize(newSize);
+
+    expect(programWindow.size.width).toBe(90);
+    expect(programWindow.size.height).toBe(75);
+  });
 });
 
-describe('move', () => {
+xdescribe('move', () => {
```
junedev, line 108: For concept exercises, all tests are unskipped.

junedev:
I am closing this PR as it seems abandoned.

PR #1709, opened 2022-03-21
Add 'if' to match other test cases

drudgesentinel:
Added the word 'if' to make instructions more clear (also now matches phrasing in other tests)


junedev (approved):

junedev:
Good catch, thanks for fixing this!

PR #1719, opened 2022-04-04
Differentiate between top and bottom

paleite:


junedev (approved):
Good catch, thanks for the fix!

junedev:
\format

junedev:
(I needed to close and re-open because of a CI bug.)

PR #1730, opened 2022-04-15
Update exercism.io into exercism.org in Template Strings concept

jiegillet:
I found a slightly outdated reference while going through the concept exercises.

### DIFF
--- a/concepts/template-strings/about.md
+++ b/concepts/template-strings/about.md
@@ -22,8 +22,8 @@ All types of expressions can be used with template strings.
 ```javascript
 const track = 'JavaScript';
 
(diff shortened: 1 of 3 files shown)


junedev (approved):

junedev:
Thanks! I think Erik also wanted to do some scripting at some point to fix all the old links.

jiegillet:
Those weren't real links, they probably wouldn't have be flagged.
Before doing the PR I did a global search for "exercism.io" and these were the only hits (actually one more, but [that one is still active](https://tracks.exercism.io/javascript/main/unimplemented)), so it should be fine :)

PR #1733, opened 2022-04-18
fix inner brackets in conditionals concept

saugatdhimal:


junedev:
@saugatdhimal I am sorry, I have to close this PR. Some else picked up the issue around the same time and posted in the issue so I merged that contribution. Please always post in the issue that you will be working on it before starting the work to avoid this in the future.

PR #1734, opened 2022-04-18
add missing inner brackets in conditionals concept

Samhhali:
add inner brackets in conditionals concept that were removed by the auto-formatting.
add prettier-ignore to prevent prettier from removing them


kerolloz (approved):

junedev (approved):
Thanks for taking care of this.

PR #1766, opened 2022-05-10
Correct 'milage' to 'mileage'

goldendemise:
The correct spelling of the word is 'mileage' rather than 'milage'

### DIFF
--- a/concepts/classes/about.md
+++ b/concepts/classes/about.md
@@ -228,7 +228,7 @@ class Car {
     return this._mileage;
   }
 
-  set milage(value) {
+  set mileage(value) {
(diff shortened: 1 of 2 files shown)


tejasbubane:
Hi @goldendemise thanks for catching this! :+1: 
There are few more instances in [about.md](https://github.com/exercism/javascript/blob/4f9b1406ea5137b41a9f935e1b5890c42f50870c/concepts/classes/about.md) - it would be great if you can fix those too.

goldendemise:
I've added the spelling fixes to `about.md`

tejasbubane (approved):
Thanks for catching and fixing this! :+1:

PR #1777, opened 2022-05-18
Remove duplicate test case in Simple Linked List

pointlessrapunzel:
The same test case 'can convert to an array' is duplicated on lines 104 and 112. This commit removes the unnecessary duplicate.


tejasbubane (approved):
@pointlessrapunzel Thanks for catching this! :+1:

PR #1799, opened 2022-06-08
Fixed Typo

farrukh-saif:
Replaced "for .. of" with "for...of"


junedev (commented):
Good catch. Please take note of the first point in the automated reply. The mistake needs to be fixed in 2 other places as well. Would be great if you can update the PR accordingly.

farrukh-saif:
Done!

PR #1802, opened 2022-06-11
Update instructions.md

dsletten:
Multiplication (*) incorrectly represented via underscore character.


SleeplessByte (changes requested):
Unfortunately this won't work, because the auto formatter will reformat it.

The way to solve this is to update this file against the [common file](https://raw.githubusercontent.com/exercism/problem-specifications/main/exercises/all-your-base/description.md), which escapes the `*` characters. You can do this manually or follow the sync description.md instructions in our CONTRIBUTING.md file.

Would you be so kind to apply the further change? I'll run the CI for you so you can see why this PR doesn't work as is.

dsletten:
Hi,
I don't understand what the problem is. The text shows up properly in other tracks.

junedev:
@dsletten As you can see in the CI checks, your proposed version does not pass the formatting check that is done for this tracks repository. Not all tracks do automated formatting checks, this is why your version might not cause issues in other tracks.

The formatter we use thinks you want to create italic text with the `*` and wants that you correct it to use underscore instead. Running auto-formatting will do this change and this is probably how the issue in the current version appeared in the first place.

SleeplessByte described above how you can avoid this issue so that your code can pass the formatting check and there will not be any issue in case someone runs auto-formatting on the code base. Would be great if you could follow the suggestion so we can get this merged.

Cool-Katt:
Closing, as it's already resolved in newer versions of the docs.

Issue #1717, opened 2022-04-01
Elyses Enchantments


LumpyHippo:
A recent change broke this exercise. What used to be task 7 is somehow combined with task 5. The results page only shows 7 tasks even though there should be 8. Task 8's condition are now listed as task 7. The original Task 7's condition are inside of Task 5 and because it is looking for a completely different function it fails causing this exercise to be incomplete and preventing progress past the Comparison exercise.

SleeplessByte:
Hi there!

I can confirm there are now 8 tasks and only 7 in the results pitch. 
If _all tests pass_, it should not prevent your from continuing, but we will fix the issue with only 7 tasks.

LumpyHippo:
I'm a beginner at Javascript so I don't know how to properly work around the issue.  Task 5 looks like it should only have Test 16, 17 and 18. The code I have for Task 5 passes those three tests.  However it fails Test 19, 20 and 21 which look like they should be part of the missing Task 7.

I made a solution for (the missing) Task 7 which is

> export function removeItemAtBottom(cards) {
>   cards.shift
>   return cards;
> }

but I still get the error 

> expect(removeItemAtBottom(stack)).toStrictEqual(expected);

as if it doesn't see the function.

SleeplessByte:
Great that you were able to solve the exercise :). 

We'll fix the "visual" issue with the tasks.

alonsogarciapablo:
+1. I can confirm this is still an issue. I was able to complete the exercise though.

PR #1831, opened 2022-07-08
fix order of tests

jecaps:
In the [exercise](https://exercism.org/tracks/javascript/exercises/elyses-enchantments/edit) on exercism.org implementing the function `insertItemAtBottom()` passes the tests of Task 7. While these are the correct tests they should be swapped with those from Task 6.

The tests listed under Task 6 are actually the tests for removing an item and should be part of Task 7. (see screenshot)

image


jecaps:
This might be related to issue #1717

junedev (approved):

PR #1843, opened 2022-07-17
Fix typo in bird-watcher/.docs/instructions.md

lujanfernaud:
Add missing "to".

> To practice, use a for loop **to** solve each of the tasks below.


tejasbubane (approved):
Thanks for catching this :+1:

PR #1858, opened 2022-08-06
Clarify instruction for Task 3

Tribhuwan-Joshi:
Added some more instruction for the task `Compose transformation functions`.

### DIFF
--- a/exercises/concept/coordinate-transformation/.docs/instructions.md
+++ b/exercises/concept/coordinate-transformation/.docs/instructions.md
@@ -28,7 +28,7 @@ const result = doubleScale(6, -3);
 
 ## 3. Compose transformation functions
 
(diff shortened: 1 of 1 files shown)


junedev (changes requested):
Thanks for the suggestion but I don't feel the additional sentence adds value here. Compare how the existing sentence reads without the math notation vs. your suggestion:

> where the result of the first function is used as the input to the second function

> pass the result of first function into the second function and should return the result

Feel free to take a pass at improving the existing text if you like but adding another sentence that says the same does not improve the situation imo.

Tribhuwan-Joshi:
okay I understand , I will close this pull request

junedev:
@Tribhuwan-Joshi Thank you for your understanding!

PR #1861, opened 2022-08-11
fix: programWindow instructions

varjmes:
Hi there!

I believe that this number needs to be changed so it makes more sense for the idea of not being able to move out of bounds.


SleeplessByte:
@junedev can you check? I actually think both may be incorrect XD.

varjmes:
I think I spotted another error so I tacked it on top.

SleeplessByte:
Thank you James! I'll double check but the two changes seem reasonable. It's a bit of a weird duck ;)

PR #1865, opened 2022-08-14
Update links in array introduction

raydancer:
I made some changes to the links in this introduction.  The basic links at the end weren't working for some reason, so I changed them to be more like the links in the Strings introduction.  I kept the MDN attributions at the end because the text was pulled directly from MDN reference site.

### DIFF
--- a/concepts/arrays/introduction.md
+++ b/concepts/arrays/introduction.md
@@ -14,7 +14,7 @@ numbers[2];
 // => 3
 ```
 
-To retrieve the number of elements that are in an array, use the `length` property:
+To retrieve the number of elements that are in an array, use the [`length`][mdn-length] property:
 
 ```javascript
 const numbers = [1, 'two', 3, 'four'];
@@ -33,11 +33,11 @@ numbers;
 
 ## Methods
 
-Some of the [methods][array_methods] that are available on every Array object can be used to add or remove from the array.
+Some of the [methods][mdn-array-methods] that are available on every Array object can be used to add or remove from the array.
 
 ### push
 
-> The `push()` method adds one or more elements to the end of an array and returns the new length of the array.[^1]
+> The [`push()`][mdn-push] method adds one or more elements to the end of an array and returns the new length of the array.[^1]
 
 ```javascript
 const numbers = [1, 'two', 3, 'four'];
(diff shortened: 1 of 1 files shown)


junedev:
@raydancer Thanks for raising this but I have trouble reproducing the problem. All the places that you changed are quotes and the source for the quote is provided in the footnote. I tested the footnote and the link in the footnote on the website https://exercism.org/tracks/javascript/concepts/arrays and it all seems to work fine. :thinking: 

Could you clarify this:
> The basic links at the end weren't working for some reason

What happened when you went to the page linked there?

raydancer:
You're correct, those links work when you go to the introduction page; however, the links don't work in the sidebar of the exercise editor: https://exercism.org/tracks/javascript/exercises/elyses-enchantments/edit

Is there something that breaks basic quotes when bringing them to the exercise sidebar?  If so, this might affect other introduction pages.

Also, I added inline code text markdown to unshift and splice to match the other sections.

SleeplessByte:
@raydancer I think you mean that they are not linked here, right:

image

raydancer:
Right.  The URL's are there, but they appear as text, not anchors.  I see that the link markdown was correct, but I don't understand why they didn't appear as links.

SleeplessByte:
I think this may be something @ErikSchierboom would want to look at. I think what you're saying is: in one place of exercise, the links _are_ auto-linked, but in another place they are not. The solution would be to make them autolink on both places, or neither places.

ErikSchierboom:
> I think this may be something @ErikSchierboom would want to look at. I think what you're saying is: in one place of exercise, the links are auto-linked, but in another place they are not. The solution would be to make them autolink on both places, or neither places.

@raydancer Could you create an issue at [exercism/exercism](https://github.com/exercism/exercism/issues) for this?

raydancer:
> @raydancer Could you create an issue at [exercism/exercism](https://github.com/exercism/exercism/issues) for this?

Done:
https://github.com/exercism/exercism/issues/6522

junedev:
I feel the better solution here is to wait for the Exercism-wide fix instead of merging this workaround. The links can still be copy and pasted and mainly exist only to provide the sources for the quotes. I am closing the PR.

PR #1872, opened 2022-08-20
Fix typecheck error on function getFirstOddCard

NickHerrig:
An correct solution does not follow type checking as stack.find() could return undefined. 

image

### DIFF
--- a/exercises/concept/elyses-analytic-enchantments/.meta/exemplar.js
+++ b/exercises/concept/elyses-analytic-enchantments/.meta/exemplar.js
@@ -51,7 +51,7 @@ export function doesStackIncludeOddCard(stack) {
  *
  * @param {number[]} stack
  *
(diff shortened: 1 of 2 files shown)


SleeplessByte:
> An correct solution does not follow type checking as stack.find() could return undefined.

Correct as in satisfies the tests, or correct as in satisfies the description and tests, or correct as in objectively correct?

Since this is an exercise early in the syllabus and the descriptions state that you will always able to find something, the implementation of `| undefined` is much better. Note that when this (and the other exercises) were written, TypeScript was less strict and would infer `number` instead of `number | undefined` when a JSDoc comment was present. However, you may find other cases of similar issues. If they are in the `Elyses` exercise(s), we'd love another PR.

SleeplessByte (approved):

PR #1896, opened 2022-09-10
Fix test description in coordinate-transformation

lujanfernaud:
In the previous test description we state that the "correct" order is `g(f(x))`. Since the order should be the opposite in the description being fixed, it should be `f(g(x))`

```
  test('should compose in the correct order: g(f(x))', () => {
    const composed = composeTransform(scaler, translator);
    expect(composed(0, 0)).toEqual([-6, 10]);
  });


  test('should compose in the opposite order: f(g(x))', () => {
    const composed = composeTransform(translator, scaler);
    expect(composed(0, 0)).toEqual([-18, 20]);
  });
```


SleeplessByte (approved):

SleeplessByte:
Thanks! You are correct :)

PR #1909, opened 2022-09-26
fix pronoun inconsistencies

BTowersCoding:
I livestreamed the Freelancer Rates exercise today and was overjoyed by the recent update to use more inclusive language! However, a couple of pronouns remained which are corrected here.


SleeplessByte (approved):
Thanks Bobby!

PR #1924, opened 2022-10-01
Feature/tp/prerequisites

TomPradat:
This partially fixes the issue #983 

Would that be possible to get an hacktoberfest label on this PR please :pray:

### DIFF
--- a/config.json
+++ b/config.json
@@ -658,7 +658,7 @@
         "name": "Scrabble Score",
         "uuid": "11771d47-1109-4579-a62b-e0b8e9583485",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["strings", "array-transformations"],
         "difficulty": 5,
         "topics": ["conditionals", "loops", "maps", "strings"]
       },
--- a/config.json
+++ b/config.json
@@ -691,7 +691,7 @@
         "name": "Bank Account",
         "uuid": "8beedf70-28b4-4cfb-ab53-309ee6f6aa78",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["classes", "conditionals", "numbers"],
         "difficulty": 3,
         "topics": ["classes", "conditionals"]
       },
--- a/config.json
+++ b/config.json
@@ -667,7 +667,7 @@
         "name": "Allergies",
         "uuid": "9d33d21c-e695-427f-9f58-dd9498d61318",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["classes", "math", "numbers", "for-loops", "arrays"],
         "difficulty": 6,
         "topics": ["arrays", "bitwise_operations", "conditionals", "loops"]
       },
--- a/config.json
+++ b/config.json
@@ -676,7 +676,13 @@
         "name": "Word Count",
         "uuid": "0073ff9a-cd6a-43cf-b8bf-4d5d8117b81b",
         "practices": [],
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -676,7 +676,13 @@
...
         "uuid": "0073ff9a-cd6a-43cf-b8bf-4d5d8117b81b",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "strings",
+          "for-loops",
+          "regular-expressions",
+          "arrays",
+          "objects"
+        ],
```
TomPradat, line 685: Here is my solution : https://exercism.org/tracks/javascript/exercises/word-count/solutions/TomPradat

I chose `for-loops` instead of `array-transformations` given the difficulty of the exercise. I omitted `conditionals` as the concept is a prerequisite of `objects` and `for-loops`

TomPradat: By the way, I think the difficulty of this exercise is underrated, regex are not that easy to deal with, I'd give a 4 or 5. If you agree I can make a new PR ? Or just a separated commit on this PR

config.json
```diff
@@ -658,7 +658,7 @@
         "name": "Scrabble Score",
         "uuid": "11771d47-1109-4579-a62b-e0b8e9583485",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["strings", "array-transformations"],
```
TomPradat, line 661: Here is my solution : https://exercism.org/tracks/javascript/exercises/scrabble-score/solutions/TomPradat

I only put 2 prerequisites because `array-transformations` already "includes" many other concetps

config.json
SleeplessByte: Make a new PR but yeah let's bump it.

config.json
```diff
@@ -667,7 +667,7 @@
         "name": "Allergies",
         "uuid": "9d33d21c-e695-427f-9f58-dd9498d61318",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["classes", "math", "numbers", "for-loops", "arrays"],
```
TomPradat, line 670: Here is my solution : https://exercism.org/tracks/javascript/exercises/allergies/solutions/TomPradat

I used array transformations a lot, but looking at the community solutions, I found [this one](https://exercism.org/tracks/javascript/exercises/allergies/solutions/enricofelinto) that I like better. That's why I didn't put `array-transformations` as a prerequisite, you can easily replace the `filter` method with a for loop

`math` is not in the list of concepts but is present a lot in other exercises' prerequisites.

config.json
```diff
@@ -685,7 +691,7 @@
         "name": "Bank Account",
         "uuid": "8beedf70-28b4-4cfb-ab53-309ee6f6aa78",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": ["classes", "conditionals", "numbers"],
```
TomPradat, line 694: Here is my solution : https://exercism.org/tracks/javascript/exercises/bank-account/solutions/TomPradat

This is pretty straightforward IMO

config.json
SleeplessByte: Myeah I wonder what happens with `math`. I guess we'll find out after merging this :P

SleeplessByte (approved):

config.json
TomPradat: I think I read somewhere that the exercise is only unlocked after all concepts have been completed. I couldn't find where I saw that though

PR #1927, opened 2022-10-02
Create representations.md

SleeplessByte:


junedev (commented):
According to this doc https://exercism.org/docs/building/tracks/docs the filename should be different. Not sure whether that is up-to-date, you might want to double check the latest file name in the F# track repo.

junedev (approved):

exercises/shared/.docs/representations.md
```diff
@@ -0,0 +1,191 @@
...
+- All _identifiers_ are renamed, using a "key lookup"
+
+## BEFORE YOU SUBMIT
+
+Before you submit your representation feedback, please check the following things:
+
+1. You don't duplicate analyzer feedback
+2. You check the "examples" tab in the submit dialog and see if the feedback makes sense for _all_ tabs.
+3. You check that you have not referred to whitespace or comments
+4. You check that you don't refer to function names, or variable names as they appear in the solution, but rather use the mapping provided (or leave names out).
```
junedev, line 18: I thought the the mapping is not provided because the UI only shows a solution and never the representation.

SleeplessByte:
@junedev thanks! I created this file by Clicking here: 

image


@iHiD / @ErikSchierboom I will merge this, but if it doesn't work, you know what to do ;). If it DOES work you _also_ know what to do ;)

junedev:
I approved since this is definitely better than having nothing but I think it is a bit long and the details about the representer versio are too much to show  to every super mentor all the time. I liked the F# version that had the details in a separate track doc that was then linked.

SleeplessByte:
Yeah i just wanted to have something there, and later make this "better". Haven't quite found a good way making this concise but still correct.

exercises/shared/.docs/representations.md
SleeplessByte: Correct. But it's supposed to be provided. It's in one of the todos

PR #1934, opened 2022-10-03
Update custom-signs specs to match required word rather than exact string

### LINKED ISSUE #1931: Custom Signs: Misleading instructions for task 3
Task 3 of the concept exercise custom-signs states:

> If the age is 50 or older, the sign will include the word _mature_, otherwise the sign will include the word _young_.

This made me believe that the tests would accept any message for the sign, as long as it included the word _mature_ if the age is 50 or older, or the word _young_ otherwise. However, that seems to not be the case, and the tests expect only very specific strings:

https://github.com/exercism/javascript/blob/2be1e8e8585f14213b44374d98162b18aa1eded3/exercises/concept/custom-signs/custom-signs.spec.js#L21-L31

I think either the description should be changed to state the phrase accepted is the one on the example provided or alternatively the tests should be changed to accept any phrase, in which case they would only for the presence of _young_ or _mature_ in those phrases.

tejasbubane:
Closes #1931


junedev (changes requested):
I think the issue would be better solved by fixing the instructions. With the new tests, a solution that just returns one word would pass the test but such a solution would defeat the point of the exercise (practicing template strings together with the ternary operator).

Instructions can include something like "The concrete output format is shown in the example below." to clarify.

tejasbubane:
@junedev Updated the instructions instead.

junedev (approved):

PR #1970, opened 2022-10-28
FIxed typo in testing which wasn't letting updatePlayer test pass

IAmRC1:


SleeplessByte:
There is no typo in this test. The test is correct. The [_proof_](https://github.com/exercism/javascript/blob/main/exercises/concept/high-score-board/.meta/exemplar.js) is used in our CI to make sure all tests are correct.

IAmRC1:
Screenshot 2022-10-30 at 12-56-14 Getting Started Exercism's Docs

Just compare `actual` and `expected` in the screenshot

SleeplessByte:
Instead of _insisting_ it must be the tests that are incorrect, please provide the diff in the [_proof_](https://github.com/exercism/javascript/blob/main/exercises/concept/high-score-board/.meta/exemplar.js?rgh-link-date=2022-10-29T12%3A39%3A22Z) that is currently incorrect according to you, that would **require** the test to be changed.

The tests pass with the current exemplar solution

PR #1982, opened 2022-11-08
Bob approaches

bobahop:
Initial commit for file structure and code. No performance article for this one, so this is pretty much there. Will let it sit a while and then go back over it.


bobahop:
@junedev @SleeplessByte  Have used the output from https://prettier.io/playground/ and it is not accepted here. I'm focused now on getting the answer-array content.md to pass,  Any suggestions for that would be appreciated. Hopefully succeeding at that will guide for the rest.

SleeplessByte:
Hi @bobahop,

It's probably due to ever so slightly different settings or version. Easiest on this repo is to add a comment with the following in it (this comment will also trigger the formatting):

/format

Thanks so far!

bobahop:
C'mon,

/format

please

bobahop:
@ErikSchierboom JS Bob approaches ready for review

exercises/practice/bob/.approaches/if-statements/content.md
```diff
@@ -0,0 +1,65 @@
...
+  return 'Whatever.';
+}
+```
+
+In this approach you have a series of `if` statements using the private methods to evaluate the conditions.
+As soon as the right condition is found, the correct response is returned.
+
+Note that there are no `else if` or `else` statements.
+If an `if` statement can return, then an `else if` or `else` is not needed.
+Execution will either return or will continue to the next statement anyway.
```
ErikSchierboom, line 25: Maybe put this in a note block?

````suggestion
```exercism/note
Note that there are no `else if` or `else` statements.
If an `if` statement can return, then an `else if` or `else` is not needed.
Execution will either return or will continue to the next statement anyway.
```
````

**Review by ErikSchierboom — APPROVED** [review]
@SleeplessByte We're approaching the approaches (pun intended) PRs with an optimistic merging strategy. As such, we're merging fairly quickly. If you have any comments, please leave them as review comments such that they can be addressed in a follow-up PR.

File: exercises/practice/bob/.approaches/if-statements/content.md (hunk 1 above)
Comment (bobahop) [reply exercises/practice/bob/.approaches/if-statements/content.md:25]: We didn't in C#, but it's a good idea.

**Comment by bobahop** [comment]
@ErikSchierboom suggestion accepted. All checks passed.

File: exercises/practice/bob/.approaches/answer-array/content.md [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+If the string has no characters left, it returns the response for saying nothing.
+
+```exercism/caution
+Note that a `null` or `undefined` `String` would be different from a `String` of all whitespace.
+A `null` or `undefined` `String` would raise a `TypeError` if `trimEnd` were applied to it.
+```
+
+The first half of the shout condition
+
+```
```
SleeplessByte, line 35:
````suggestion
```javascript
````

File: exercises/practice/bob/.approaches/if-statements/content.md [hunk 3]
```diff
@@ -0,0 +1,67 @@
...
+  if (speech == '') return 'Fine. Be that way!';
+
+  const isQuestion = speech.endsWith('?');
+  const isShout = /[A-Z]{1}/.test(speech) && speech == speech.toUpperCase();
+
+  if (isShout)
+    return isQuestion
+      ? "Calm down, I know what I'm doing!"
+      : 'Whoa, chill out!';
+  if (isQuestion) return 'Sure.';
```
SleeplessByte, line 15: Not using braces `{` for if statements is very non-standard in JavaScript. I see them a lot in the approaches and whilst mostly a stylistic choice, I don't think it's smart to have so many examples with a defacto not-accepted style.

bobahop: Thanks, the javascript is how the formatter formatted it, I believe.

SleeplessByte: `prettier` would never format it like this if brackets were present.

bobahop: So does running format in a comment not actually format the code correctly? If not, sorry, that was what I was depending on.

SleeplessByte: It does!

In generally prettier will not add or remove brackets. It _will_ however format the code inside the brackets, if they are present :)

PR #1985, opened 2022-11-09
Leap approaches

bobahop:
/format


bobahop:
Please

/format

please

bobahop:
Although they are on separate branches, it seems the Leap files somehow got included with the Grains PR, so it may be best to just review the Grains PR for both approaches. If both approach's files can be successfully merged, then this PR can just be closed, I think.

bobahop:
This PR can be closed without merging since the Grains branch came from the Leap branch and merging Grains took care of the Leap files.

PR #2024, opened 2022-12-20
Update windowing-system.spec.js

anakinsonone:
Line 89, 90. While solving the task 4, I passed test 9 but couldn't pass test 10. So I logged the inputs and then found that for test 10, I was receiving input (80, -10) but the program was checking for input (0, -10). So I think this might be an issue.

### DIFF
--- a/exercises/concept/windowing-system/windowing-system.spec.js
+++ b/exercises/concept/windowing-system/windowing-system.spec.js
@@ -86,10 +86,10 @@ describe('resize', () => {
 
   test('does not resize below 1', () => {
     const programWindow = new ProgramWindow();
-    const newSize = new Size(0, -10);
+    const newSize = new Size(80, -10);
     programWindow.resize(newSize);
 
-    expect(programWindow.size.width).toBe(1);
+    expect(programWindow.size.width).toBe(80);
     expect(programWindow.size.height).toBe(1);
   });
 });


SleeplessByte:
Hi @anakinsonone . There is no issue with the tests. 

Most likely you're checking for some _absolute_ value instead of a _delta_.

anakinsonone:
Hi @SleeplessByte , I don't get it. I am sure my solution is correct. I pass the test where the given input is (300, 200). My solution correctly resizes the window to (300, 200). But for the next test the inputs that I receive are (80, -10) but the editor checks for the input (0, -10). If these were the values that my code was receiving, it would provide the correct answer as (1, 1) but it does not.

anakinsonone:
image

Here, you can see for yourself. I just console logged the "newSize" parameter. As you can see the output on the right, it states (80, -10). But in the test case, you can see, it shows "newSize" to be (0, -10)

SleeplessByte:
You may want to check your implementation of `class Size`. If you are seeing `Size { width = 80, height = -10 }`, it's likely that the _width_ parameter is still using the _default_ value of 80 instead of overwriting it with what is being passed into the constructor.

Showing tests are as expected

I cannot reproduce this with my own solution; 

Menu to revert exercise to start

You may want to try this again by resetting the exercise, after making sure you have no pending updates on https://exercism.org/tracks/javascript/exercises/windowing-system .

Tests should _not_ need to be changed, as shown in the screenshot above - we even have an example implementation that passed the tests in the CI, so we are 100% confident that there is no problem with the tests.

anakinsonone:
Yup, got it. Sorry for the troubles @SleeplessByte . Got it.

PR #2036, opened 2023-01-05
Update instructions.md

muneer-ahmed-khan:
remove the the "s" character in Check Valid Phone Number example to have same output number as input number.


SleeplessByte (approved):
Thanks

PR #2042, opened 2023-01-13
fix: test case for getFirstOddCard

shubhamakpaliwal:


SleeplessByte:
2 is not an odd number. The tests are correct.

PR #2049, opened 2023-01-20
Ensure recalculation when a new function is passed in

johnsyweb:
Context
-------

Coming from other programming languages, the [scope](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures#scoping_with_let_and_const) of the cached values may be surprising in JavaScript. When looking at community solutions, I was certainly surprised that other folks were not testing whether `f` had changed between invocations of the closure returned from `memoizeTransform()`.

Change
------

Add a new test case asserting that different values of `f` have different caches.


SleeplessByte (approved):

johnsyweb:
Thank you 🙇

PR #2062, opened 2023-01-31
Correct a bank-account test to close account

itsjustaaron:
The test that checks "close already closed account throws error"  does not currently try to close the account twice. Adding in `account.close()` before the `expect()` should fix this.

### DIFF
--- a/exercises/practice/bank-account/bank-account.spec.js
+++ b/exercises/practice/bank-account/bank-account.spec.js
@@ -66,6 +66,7 @@ describe('Bank Account', () => {
 
   xtest('close already closed account throws error', () => {
     const account = new BankAccount();
+    account.close();
     expect(() => {
       account.close();
     }).toThrow(ValueError);


SleeplessByte:
As you know from the forum post, we're discussing how to resolve this. 

I did want to re-open this to _show_ you that we currently have CI running on each exercise, on each test. If a test is changed, almost always the "proof" also needs to be updated.

The proof lives here: https://github.com/exercism/javascript/blob/main/exercises/practice/bank-account/.meta/proof.ci.js

If we're changing the test to only throw on the "second" close, then we must update the solution to be "open by default", which may or may not be what we want to do, depending on the outcome of the [forum post](https://forum.exercism.org/t/js-bank-account-test-error/3264/2).

SleeplessByte:
`problem-specifications` [was updated](https://github.com/exercism/problem-specifications/pull/2192) so we know how to resolve it now. This PR is therefor outdated. 

Thank you for the contribution nonetheless!

Issue #1770, opened 2022-05-12
Elyses enchantments: Tests requires a solution with mutations


luckysoni:
The spec has tests that require mutation of the array being passed to the function being tested. This restricts users to submit solutions that use immutability.

Consider the following test:
```
test('adding multiple cards to the stack to the bottom', () => {
    const stack = [1];

    insertItemAtBottom(stack, 5);
    insertItemAtBottom(stack, 9);

    const expected = [9, 5, 1];
    expect(stack).toStrictEqual(expected);
  });
```

The assertion is being made on the original `stack` which will require it to be mutated.

I am happy to fix the tests if folks agree that this Is suboptimal.

SleeplessByte:
This is intentional, because we want to teach people how to _mutate_ the array.

PR #2064, opened 2023-02-04
Update enchantments.spec.js to fit the documentation provided

Dyrits:
The `enchantment.js` file is documented, specifying that the function should return an array of numbers. The tests don't use the returned value, but consider that the initial array has been modified. Considering that the other tests are using the returned value, why not these ones ?


SleeplessByte:
Hi Dylan!

Short form answer is here: https://github.com/exercism/javascript/issues/1770
Long form answer and discussion is here: http://forum.exercism.org/t/discouraged-code-practice-found-in-elyses-enchantments/242/8

Feel free to respond on the forum or open a new topic there to discuss further if you would like so :)

PR #2109, opened 2023-03-18
Update introduction.md

gelbelle:
Clarified when square bracket notation is required when using a key to access your object value.
I found this incredibly difficult when I was first learning and it caused me many hours of googling as to why my dot notation wasn't working.

### DIFF
--- a/concepts/objects/introduction.md
+++ b/concepts/objects/introduction.md
@@ -75,7 +75,8 @@ obj[key];
 ```
 
 Using the dot notation as a shorthand has the same restriction as omitting the quotation marks.
-It only works if the key follows the identifier naming rules. In other words, if you're key is not a single word using only letters, you must use square bracket notation.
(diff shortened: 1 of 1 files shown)


concepts/objects/introduction.md
```diff
@@ -75,7 +75,7 @@ obj[key];
 ```
 
 Using the dot notation as a shorthand has the same restriction as omitting the quotation marks.
-It only works if the key follows the identifier naming rules.
+It only works if the key follows the identifier naming rules. In other words, if you're key is not a single word using only letters, you must use square bracket notation.
```
SleeplessByte, line 78:
```suggestion
It only works if the key follows the identifier naming rules.
In other words, if you're key is not a single word using only letters, you must use square bracket notation.
```

Moving this to the next line so it adheres to our markdown styleguide.

SleeplessByte (approved):
Thank you, this is a good change.

PR #2159, opened 2023-04-29
Add prerequisites

safwansamsudeen:
As part of #984, I've added the prerequisites for the House exercises and removed the `topics` key as per Erik's comment in #960. Note that many entries with `prerequisites` still have `topics` - this should be addressed, maybe?

I just completed this exercise myself. This is [the top solution](https://exercism.org/tracks/javascript/exercises/house/solutions/sporks) and it uses arrays, something needed to implement this exercise without hardcoding values. You also need to learn strings, along with string manipulation (I'm not sure if we should add to the `practices` key in this PR: when I get the clarification, I'll add `template-strings` to it). Finally, all the top solutions use `for` loops - perhaps this could be done using `map` or `while`, but the most straightforward solution will use `for` loops.

I apologize for the `sync exercise` commit and it's reversion - it happened due to a conflict with another issue I opened. New-ish to git!

### DIFF
--- a/.github/workflows/action-format.yml
+++ b/.github/workflows/action-format.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            github.issues.createComment({
+            github.rest.issues.createComment({
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
@@ -34,7 +34,7 @@ jobs:
         id: fork_status
         run: |
           IS_FORK="$(jq '.head.repo.fork' "/tmp/pr.json")"
-          echo "::set-output name=fork::$IS_FORK"
+          echo "fork=$IS_FORK" >> "$GITHUB_OUTPUT"
 
       - name: 'Setup SSH deploy key'
         if: steps.fork_status.outputs.fork == 'false'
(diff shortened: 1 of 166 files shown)


safwansamsudeen:
@SleeplessByte thanks for reopening this. I realize that you're currently not accepting CCs, but @TomPradat specifically said he'll look into this when he has time.

TomPradat (approved):
I myself have done this exercise using recursion, but given the difficulty of the exercise and given the place of the "recursion" concept in the concepts tree, I believe it's a good idea not to put it in the prerequisites 👍 

IMO there is no need to add `template strings` here. And I totally agree with the other prerequisites you have added. For loop is the most straightforward I believe too.

No worries about the sync exercise commit, these commits will all be squashed anyways.

Thank you for working on this !

PS: Sorry for the time it took me to review this PR, I had a lot to deal with

safwansamsudeen:
@TomPradat thanks for the review. Let me know when you're free again so that I can submit PRs for the rest of the exercises.

Yeah, we don't need `template-strings` in prerequisites - I meant that we add it to `practices`, as Erik has said.

SleeplessByte (approved):

TomPradat:
> @TomPradat thanks for the review. Let me know when you're free again so that I can submit PRs for the rest of the exercises. What about merging this?

You can submit PRs one by one. Don't expect me to be quick these days at reviewing but I'll do it gladly when I have some time :slightly_smiling_face:.

> Yeah, we don't need `template-strings` in prerequisites - I meant that we add it to `practices`, as Erik has said.

:+1:

PR #2160, opened 2023-04-30
Add basic files for Pizza Order exercise

safwansamsudeen:
This is the PR in response to #945. I've added `hints.md`, `instructions.md`, and `introduction.md`.

For `.docs/about.md`, I couldn't find any file like that in other JS exercises - could you explain further what I'm supposed to write in that file? Or do you maybe mean `concepts/recursion/about.md`? I also couldn't find `languages/javascript/exercises/concept/README.md`. In this repo, I'm guessing we should fill out `concepts/recursion/introduction.md` - but what should I put in that file vs. `.docs/introduction.md`? I looked around in other concept exercises but couldn't understand still.

Let me know what I should change or add here.

### DIFF
--- a/exercises/concept/pizza-order/.docs/instructions.md
+++ b/exercises/concept/pizza-order/.docs/instructions.md
@@ -1,15 +1,19 @@
 # Instructions
+
 You run a pizza shop, and offer three types of pizzas:
+
 - Margherita: $7
 - Caprese: $9
 - Formaggio: $10
 
-If customers want, they can add an unlimited number of extra options: either "ExtraSauce" for $1 or "ExtraToppings" for $2. 
+If customers want, they can add an unlimited number of extra options: either "ExtraSauce" for $1 or "ExtraToppings" for $2.
 
 Your task is to write code that assists the customer in figuring out the cost to them.
 
 ## Calculate the price of a pizza
+
 Provided the pizza's name as the first argument, and an unlimited number of added options, calculate the price of the pizza in dollars.
+
 ```js
 pizzaPrice('Margherita');
 // => 7
@@ -28,6 +32,7 @@ pizzaPrice(
 ```
 
 ## Calculate the total price of an order
+
 Your function will take in a list of `PizzaOrder`s and should return the total price of the order in dollars.
 Each `PizzaOrder` has a `pizza` property - the pizza's name, and an `extras` property - the list of extra options.
 
@@ -37,4 +42,5 @@ const caprese = new PizzaOrder('Caprese', 'ExtraToppings');
 orderPrice([margherita, caprese]);
 // => 18
 ```
+
 You'll realize that you can't write this using recursion, as one test with a tremendous amount of orders will raise a `Maximum call stack size exceeded`. No worries, this is intentional - try implementing this function using an imperative loop!
--- a/exercises/concept/pizza-order/.docs/instructions.md
+++ b/exercises/concept/pizza-order/.docs/instructions.md
@@ -87,4 +87,3 @@ There are only two viable solutions to a call stack error caused by a synchronou
 - rewrite the recursive function to an imperative loop, which will execute the body of the loop, without having to enter a function, thus without increasing the stack.
 ~~~~
 <!-- prettier-ignore-end -->
-
--- a/exercises/concept/pizza-order/.docs/instructions.md
+++ b/exercises/concept/pizza-order/.docs/instructions.md
@@ -33,7 +33,7 @@ pizzaPrice(
 
 ## Calculate the total price of an order
 
-Your function will take in a list of `PizzaOrder`s and should return the total price of the order in dollars.
+Your function is called with a list of `PizzaOrder`s and should return the total price of the order in dollars.
 Each `PizzaOrder` has a `pizza` property - the pizza's name, and an `extras` property - the list of extra options.
 
 ```js
@@ -43,4 +43,5 @@ orderPrice([margherita, caprese]);
 // => 18
 ```
 
-You'll realize that you can't write this using recursion, as one test with a tremendous amount of orders will raise a `Maximum call stack size exceeded`. No worries, this is intentional - try implementing this function using an imperative loop!
+You'll realize that you can't write this using recursion, as one test with a tremendous amount of orders will raise a `Maximum call stack size exceeded`.
+No worries, this is intentional - try implementing this function using an imperative loop!
--- a/exercises/concept/pizza-order/.docs/instructions.md
+++ b/exercises/concept/pizza-order/.docs/instructions.md
@@ -1,1 +1,40 @@
 # Instructions
+You run a pizza shop, and offer three types of pizzas:
+- Margherita: $7
+- Caprese: $9
+- Formaggio: $10
+
+If customers want, they can add an unlimited number of extra options: either "ExtraSauce" for $1 or "ExtraToppings" for $2. 
+
+Your task is to write code that assists the customer in figuring out the cost to them.
+
+## Calculate the price of a pizza
+Provided the pizza's name as the first argument, and an unlimited number of added options, calculate the price of the pizza in dollars.
+```js
+pizzaPrice('Margherita');
+// => 7
+
+pizzaPrice('Caprese', 'ExtraSauce', 'ExtraToppings');
+// => 12
+
+pizzaPrice(
+  'Caprese',
+  'ExtraToppings',
+  'ExtraToppings',
+  'ExtraToppings',
+  'ExtraToppings'
+);
+// => 17
+```
+
+## Calculate the total price of an order
+Your function will take in a list of `PizzaOrder`s and should return the total price of the order in dollars.
+Each `PizzaOrder` has a `pizza` property - the pizza's name, and an `extras` property - the list of extra options.
+
+```js
+const margherita = new PizzaOrder('Margherita');
+const caprese = new PizzaOrder('Caprese', 'ExtraToppings');
+orderPrice([margherita, caprese]);
+// => 18
+```
+You'll realize that you can't write this using recursion, as one test with a tremendous amount of orders will raise a `Maximum call stack size exceeded`. No worries, this is intentional - try implementing this function using an imperative loop!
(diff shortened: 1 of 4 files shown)


safwansamsudeen:
@SleeplessByte any update on this?

exercises/concept/pizza-order/.docs/introduction.md
```diff
@@ -1 +1,67 @@
 # Introduction
+
+The ability for something to be defined in terms of itself is called recursion. Recursive functions are functions that call themselves.
```
SleeplessByte, line 3:
```suggestion
The ability for something to be defined in terms of itself is called recursion.
Recursive functions are functions that call themselves.
```

exercises/concept/pizza-order/.docs/introduction.md
```diff
@@ -1 +1,67 @@
...
+
+Suppose that you have a function called `recurse`. This function is recursive if it calls itself inside its body, like this:
```
SleeplessByte, line 5:
```suggestion
Suppose that you have a function called `recurse`.
This function is recursive if it calls itself inside its body, like this:
```

exercises/concept/pizza-order/.docs/introduction.md
```diff
@@ -1 +1,67 @@
...
+
+```js
+function recurse() {
+  // ...
+  recurse();
+  // ...
+}
+```
+
+A recursive function should always have a condition to stop calling itself and return a value, known as a _base case_. Otherwise, it will call itself indefinitely. In the absence of a base case, or when a function calls itself too many times (the number varies per the compiler), a `Maximum call stack size exceeded` error will be raised.
```
SleeplessByte, line 15:
```suggestion
A recursive function usually has a condition to stop calling itself and return a value, known as a _base case_.
If a base case is missing, in most cases, because it will call itself indefinitely, it would be able to run forever. 
In reality, in most of those situations, you'll end up with a "StackSize error": an error raised by the runtime because the _stack_ of function calls has grown beyond a predefined limit because each recursive call adds to this _stack_ until it returns (and it doesn't).
The message of this error is `Maximum call stack size exceeded`.
```

exercises/concept/pizza-order/.docs/introduction.md
```diff
@@ -1 +1,67 @@
...
+  if (baseCondition) {
+    // stop calling itself
+    //...
+  } else {
+    recurse();
+  }
+}
+```
+
+Recursive functions often can be used instead of `for` loops for more succinct code. For example, take a countdown.
```
SleeplessByte, line 28:
```suggestion
Recursive functions often can be used instead of `for` loops for more succinct code.
For example, take a countdown.
```

exercises/concept/pizza-order/.docs/instructions.md
```diff
@@ -1 +1,46 @@
...
+  'ExtraToppings',
+  'ExtraToppings',
+  'ExtraToppings'
+);
+// => 17
+```
+
+## Calculate the total price of an order
+
+Your function will take in a list of `PizzaOrder`s and should return the total price of the order in dollars.
```
SleeplessByte, line 36:
```suggestion
Your function is called with a list of `PizzaOrder`s and should return the total price of the order in dollars.
```

exercises/concept/pizza-order/.docs/instructions.md
```diff
@@ -1 +1,46 @@
...
+Each `PizzaOrder` has a `pizza` property - the pizza's name, and an `extras` property - the list of extra options.
+
+```js
+const margherita = new PizzaOrder('Margherita');
+const caprese = new PizzaOrder('Caprese', 'ExtraToppings');
+orderPrice([margherita, caprese]);
+// => 18
+```
+
+You'll realize that you can't write this using recursion, as one test with a tremendous amount of orders will raise a `Maximum call stack size exceeded`. No worries, this is intentional - try implementing this function using an imperative loop!
```
SleeplessByte, line 46:
```suggestion
You'll realize that you can't write this using recursion, as one test with a tremendous amount of orders will raise a `Maximum call stack size exceeded`.
No worries, this is intentional - try implementing this function using an imperative loop!
```

SleeplessByte, line 46: I don't understand this:

1. The exemplar solution uses `recursion`.
2. The point of this exercise is to use recursion.
3. Why is this suggesting to use an imperative loop?

exercises/concept/pizza-order/.docs/introduction.md
SleeplessByte: Counter examples are recursive functions that are killed from the outside, and tail-call optimized recursive functions (not possible in browsers and node at the moment). Async recursive functions also are options and technically don't require a base case.

SleeplessByte (commented):
Few remarks. 

Syntax/formatting changes are due to the Exercism Markdown style imposed on these files which requires each sentence to be on a separate line.

exercises/concept/pizza-order/.docs/introduction.md
safwansamsudeen: I'll take some time to improve on this, then.

exercises/concept/pizza-order/.docs/instructions.md
safwansamsudeen: So this was my confusion about this PR, as I've said in Issue #945. You say: 
> It must be solved using some imperative loop (for loop, .foreach, or best: .reduce)

I first asked a question on the issue, and then after I reread the issue, it seemed that you wanted to teach the student about these errors. I thus proceeded likewise.

> The exemplar solution uses recursion.

No, it uses `.reduce`.
```js
export function orderPrice(pizzaOrders) {
  return pizzaOrders.reduce(
    (result, order) => result + pizzaPrice(order.pizza, ...order.extras),
    0
  );
}
```

> The point of this exercise is to use recursion.

Exactly, which is why I find this to be somewhat convoluted. Should we remove the large test case, maybe? If we do that, it then doesn't teach the errors themselves (about which I agree that it's important to teach).

exercises/concept/pizza-order/.docs/introduction.md
safwansamsudeen: Or on retrospect, I'll just commit this for now. I'll add the tail-call optimization later.

safwansamsudeen:
What's the reason for the most recent commit failing the checks?

exercises/concept/pizza-order/.docs/instructions.md
SleeplessByte: Oh right, yes I was looking at the wrong thing.

Eh. Ok, I'll change the wording a bit. Your intention (and understanding) was correct. We really do want people to hit this "issue" because it is _very_ common when trying to use recursion where you cannot, like in this case.

Thanks!

SleeplessByte:
> What's the reason for the most recent commit failing the checks?

You can run `yarn format` to fix it, but the problem is that there is a formatting issue. Run `./bin/check-formatting.sh` to discover them, or use: `yarn prettier --check "**/*.{js,jsx,ts,tsx,css,sass,scss,html,json,md,yml}"`

exercises/concept/pizza-order/.docs/instructions.md
```diff
@@ -1 +1,47 @@
...
+
+```js
+const margherita = new PizzaOrder('Margherita');
+const caprese = new PizzaOrder('Caprese', 'ExtraToppings');
+orderPrice([margherita, caprese]);
+// => 18
+```
+
+You'll realize that you can't write this using recursion, as one test with a tremendous amount of orders will raise a `Maximum call stack size exceeded`.
+No worries, this is intentional - try implementing this function using an imperative loop!
```
SleeplessByte, line 47:
`````suggestion
You'll realize that you can't write this using recursion, as one test with a tremendous amount of orders will raise a `Maximum call stack size exceeded`.
No worries, this is intentional - try implementing this function using an imperative loop!
You have many options, such as, but not limited to using `reduce` or a `for` loop.

<!-- prettier-ignore-start -->
~~~~exercism/advanced
When the JavaScript interpreter is running the JavaScript code, it will keep track of which functions it has entered (started to call) on a data structure called "a stack".
When the function returns (ends), it is removed from the stack.

However, this stack has a limited size.
The most common mistake made is a recursive function that never ends.
Each call is placed on the stack, but before it returns, another call is placed on the stack.

```javascript
function kaboom() {
  kaboom()
}

kaboom()
// => RangeError: Maximum call stack size exceeded
```

The stacktrace of this error shows the same line over and over, which makes sense, because the function calls itself.
Whilst it has no real practical application in most cases, you can find out how tall that stack can get.

```javascript
let calls = 0;
function kaboom() {
  calls +=1 ;
  kaboom()
}

kaboom()
// => RangeError: Maximum call stack size exceeded

console.log(calls)
// => a number, generally higher than 10.000
```

There are only two viable solutions to a call stack error caused by a synchronous recursive function:
- ensure the functions return before the stack limit is reached, usually by adding or fixing a base case.
- rewrite the recursive function to an imperative loop, which will execute the body of the loop, without having to enter a function, thus without increasing the stack.
~~~~
<!-- prettier-ignore-end -->

`````

SleeplessByte (approved):
Thank you so much @safwansamsudeen and sorry for the long turnaround time. This is a great addition to this exercise!

safwansamsudeen:
Thanks for the quick review after you arrived!

Is there anything else we can do for this exercise? It's still shows as WIP in the UI, and maybe #945 should be closed.

SleeplessByte:
@safwansamsudeen just this: https://github.com/exercism/javascript/pull/2191

safwansamsudeen:
@SleeplessByte, for some reason, the hint of the second task is also being shown for the first:
image
The hints are correct in the [doc](https://github.com/exercism/javascript/blob/main/exercises/concept/pizza-order/.docs/hints.md), and the heading seems to be correct, so why's this happening?

SleeplessByte:
They are not the same ;)

In the second you take an array and shift one out. This keeps a single array and a single item. In the first you copy the entire array except the first item.

Which one is more resource heavy?

safwansamsudeen:
Ah, `.slice` copies the array, I see. Thanks!

Check the [hint duplication here](https://exercism.org/tracks/javascript/exercises/pizza-order/edit).

AxelRigal:
Hello first of all thanks for your exercise ! 

I'm doing it and its great one !

Little think if you could update add a return there and everywhere there is recursivity.
image

Thank you very much again.
Sorry to put this comment here but idk where to notify it.
👋

safwansamsudeen:
@AxelRigal thanks for your comment!

This change makes sense to me - what do you think, @SleeplessByte?

I'm glad to hear that you enjoyed the exercise. Generally, we now use the [forum](https://forum.exercism.org/) to post any suggestions.

PR #2164, opened 2023-05-03
Bump node LTS to v18 and add v20 to CI

tejasbubane:
Deprecate versions < 18

### DIFF
--- a/.github/workflows/ci.js.yml
+++ b/.github/workflows/ci.js.yml
@@ -29,7 +29,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: ['18', '19', '20']
+        node-version: ['18', '20']
 
     steps:
       - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.js.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: ['16', '17', '18']
+        node-version: ['18', '19', '20']
```
SleeplessByte, line 32: Since 19 is unstable, I don't think we should test for it at all (same for 17, 21, etc.)

.github/workflows/pr.ci.js.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: ['16', '17', '18']
+        node-version: ['18', '19', '20']
```
SleeplessByte, line 36: Since 19 is unstable, I don't think we should test for it at all (same for 17, 21, etc.)

SleeplessByte (approved):

PR #2189, opened 2023-06-09
Fix coordinate-transformation test case answers

hsk-kr:
- since the test cases reuse closure functions, the answers need to be accumulated.


hsk-kr:
It's really late to reply. Yes, it has turned out that I misunderstood the assignment. Thank you for your quick response.

PR #2218, opened 2023-07-19
Replaced the magic number 22 with a constant named billableDaysPerMonth

mazzog:
As per the discussion pertaining to the topic in the forums at the following link:

[Thread](https://forum.exercism.org/t/about-the-exemplar-solution-in-the-freelancer-rates-exercise-in-the-js-track/6540)


SleeplessByte (approved):
Thank you

PR #2228, opened 2023-07-30
Make tests robust against missing implementations of testees

MatthijsBlom:
Previously, when `translate2d` or `scale2d` were still lacking an implementation, the tests would crash instead of fail properly. This patch removes the need for the testees to be implemented, so that all missing implementations result in proper test failures only.

This issue was [discussed on the forum](https://forum.exercism.org/t/bug-on-coordinate-transformation-fixed/6678).


SleeplessByte (approved):

MatthijsBlom:
Oops, should have passed `[no important files changed]` 🤦

SleeplessByte:
I forgot too...

PR #2287, opened 2023-09-28
Grains approach fix

bobahop:
fixed to `throw new Error` instead of just `throw`, which apparently does not throw an error.


tejasbubane (approved):
Thanks :+1:

PR #2297, opened 2023-10-17
Add node 21 to CI

tejasbubane:
https://nodejs.org/en/blog/announcements/v21-release-announce

### DIFF
--- a/.github/workflows/ci.js.yml
+++ b/.github/workflows/ci.js.yml
@@ -29,7 +29,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: ['18', '20']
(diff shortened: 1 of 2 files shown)


tejasbubane:
@SleeplessByte Yes. I think we should keep the LTS (18) and latest (21) versions in CI. I guess many users installing nodejs first time might install the latest version.

SleeplessByte:
The latest version when uneven often just doesn't work for various reasons. I don't think we will be _able_ to solve the issues that they run into.

Can we mark the 21 one as "allowed to fail"?

I do think we should add 20!

PR #2317, opened 2023-11-12
fix: change story website to non-existant destination

vaeng:
See: https://forum.exercism.org/t/ad-link-in-lucky-numbers-exercise/8204/2

### DIFF
--- a/exercises/concept/lucky-numbers/.docs/instructions.md
+++ b/exercises/concept/lucky-numbers/.docs/instructions.md
@@ -1,7 +1,7 @@
 # Instructions
 
 Your friend Kojo is a big fan of numbers.
-He has a small website for playing around with numbers: www.kojo-presents-fun-with-numbers.com.
+He has a small website called 'fun-with-numbers'.
 Kojo is not that good at programming so he asked you for help.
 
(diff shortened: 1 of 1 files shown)


neenjaw (approved):

neenjaw:
@vaeng I think you'll need @iHiD or @ErikSchierboom to approve this one.  Not sure how available the core js maintainers are at the moment

exercises/concept/lucky-numbers/.docs/instructions.md
```diff
@@ -1,7 +1,7 @@
 # Instructions
 
 Your friend Kojo is a big fan of numbers.
-He has a small website for playing around with numbers: www.fun-with-numbers.com.
+He has a small website for playing around with numbers: www.kojo-presents-fun-with-numbers.com.
```
kotp, line 4: Why would we have a link that goes nowhere?  It is not guaranteed to go nowhere and may need to be changed later.

kotp (changes requested):
We could use the established imaginary (yet real) provided example standard, such as `https://fun-with-numbers.example.com` instead.

vaeng:
Maybe it is best to just drop the url. it does not add to the exercise at all.

ErikSchierboom (approved):

PR #2321, opened 2023-11-17
Format track config with `configlet fmt`

BNAndras:
The configlet CI is currently failing because `configlet fmt` is showing the track config to be unformatted. This is a recently added check for configlet.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -10,3 +10,4 @@ LICENSE
 # These are formatted via configlet and will not match prettier
 exercises/**/.meta/config.json
 exercises/**/.approaches/config.json
+config.json
--- a/.prettierignore
+++ b/.prettierignore
@@ -11,3 +11,7 @@ LICENSE
 exercises/**/.meta/config.json
 exercises/**/.approaches/config.json
 config.json
+
+# Originates from https://github.com/exercism/problem-specifications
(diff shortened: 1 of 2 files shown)


BNAndras:
Right, if I run `./bin/format.sh`, we go back to the original formatting that `configlet fmt` doesn't like.  `configlet fmt` should format the config.json files so maybe prettier shouldn't format those JSON files. That'll let us make both CI checks happy.

SleeplessByte:
If we want to format it using configlet, we should probably:

- ignore the file using [`.prettierignore`](https://prettier.io/docs/en/ignore.html#ignoring-files-prettierignore)
- update `format.sh` to call both `prettier` and `configlet`.

I would be happy to see that change!

BNAndras:
I’ll take a look. configlet create now formats the track config as well so that this issue is likely to come up again.

SleeplessByte:
Understood! Ping me when done or if you cannot figure it out and I'll fix it ;)

BNAndras:
Last commit was because raindrops has an exercism/note section in the instructions.md. Prettier wanted to revert the tildes to backticks.

BNAndras:
@SleeplessByte, the CI is passing.

SleeplessByte (approved):

PR #2345, opened 2023-12-10
Fix 'name cannot be modified' test

filbo:
Without the IIFE, `expect()` fails to detect the failure; see https://forum.exercism.org/t/8300

### DIFF
--- a/exercises/practice/robot-name/robot-name.spec.js
+++ b/exercises/practice/robot-name/robot-name.spec.js
@@ -73,7 +73,7 @@ describe('Robot', () => {
     const modifyInternal = () => {
       robot.name += 'a modification';
     };
-    expect(modifyInternal).toThrow();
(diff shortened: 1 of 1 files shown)


exercises/practice/robot-name/robot-name.spec.js
```diff
@@ -73,7 +73,7 @@ describe('Robot', () => {
     const modifyInternal = () => {
       robot.name += 'a modification';
     };
-    expect(modifyInternal).toThrow();
+    expect(() => modifyInternal()).toThrow();
```
SleeplessByte, line 76: This code is _exactly_ the same but if it solves something then... sure!

SleeplessByte (approved):

SleeplessByte:
This is blocked by #2321

SleeplessByte:
Could you rebase this on main?

PR #2355, opened 2024-01-08
Add initial functions document

SleeplessByte:
Some minimal information that we want[ed] to share about Functions in JavaScript


ErikSchierboom (approved):
Neat! What a bewildering variety of syntax :)

PR #2360, opened 2024-01-15
Run `configlet sync --docs` to sync things and update Leap

IsaacG:
+cc @ErikSchierboom  @junedev @tejasbubane


ErikSchierboom:
For some reason there are many conflicts.

PR #2374, opened 2024-01-26
Stop making copies of LICENSE and .gitignore files for each exercise.

IsaacG:
The base repo LICENSE should suffice.

git will honor the root .gitignore if a child directory does not contain one.

### DIFF
--- a/scripts/sync
+++ b/scripts/sync
@@ -38,7 +38,7 @@ function copyConfigForAssignment(assignment) {
     .ShellString(JSON.stringify(mergedPackageJson, undefined, 2) + '\n')
     .to(assignmentPackageFilename);
 
-  ['.eslintrc', '.npmrc', 'babel.config.js', 'LICENSE', '.gitignore'].forEach(
+  ['.eslintrc', '.npmrc', 'babel.config.js'].forEach(
     (file) => {
       shell.cp(file, destination);
     }


SleeplessByte:
Both need to be copied because the exercises are downloaded individually. There is no "root" file when setting up a track via the Exercism CLI.

IsaacG:
Is there a reason this track needs to download the license with every exercise? Should this happen on all tracks? Most tracks do not do this.

SleeplessByte:
We have seen JavaScript students directly upload their solutions to their github or gitlab. Stripping the license is probably not something we want ;)

IsaacG:
How does that differ from other tracks?

SleeplessByte:
I don't know why what other tracks have done should impact how JavaScript is set up. For example for the longest time JavaScript was the only track that formatted the config.json to be sorted on difficulty and then lexicographically. We also include the `.gitignore` because people don't know how git works, but technically it doesn't need to be there, again because it helps reducing support requests of people uploading MBs/GBs to repositories.

As long as there is no top-down guidance on these things, I don't think tracks need to be uniform here.

IsaacG:
If it makes sense that each `exercise download` should include a license to protect Exercism's intellectual property, then that should always be true. If there is no need for each download to include a license to protect their IP, then there is no need for one. 

Some aspects of track tooling make sense to change from track to track based on language needs and tooling. Other aspects, like whether or not a license file should always be present, are not related to the language or tooling in any way. Things not related to the language or tooling should be uniform where possible and handled at an org level. 

Is there any reason the license thing should be different in this track vs others? If not, it should be consistent and handled at the org level. 

+cc @iHiD and @ErikSchierboom for licensing input.

ErikSchierboom:
A forum post would be great.

kotp:
The forum post is http://forum.exercism.org/t/downloading-the-license-file-with-every-exercism-download-command/9958

PR #2379, opened 2024-01-30
[New Exercise] Parallel letter frequency

Cool-Katt:
Adding a new exercise ` parallel-letter-frequency`

I also want to write up something about a web worker solution (maybe an approach or an article) but i might need some help with that, maybe in a separate PR or in a follow-up commit to this one. Opinions on that are appreciated.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2541,13 +2541,7 @@
         "name": "Eliud's Eggs",
         "uuid": "2a3ecf62-fb5d-4dac-8369-72e7976a8f57",
         "practices": [],
-        "prerequisites": [
-          "strings",
-          "regular-expressions",
-          "rest-and-spread",
-          "arrow-functions",
-          "basics"
-        ],
+        "prerequisites": [],
         "difficulty": 2
       },
       {
--- a/config.json
+++ b/config.json
@@ -2543,6 +2543,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 2
+      },
+      {
+        "slug": "parallel-letter-frequency",
+        "name": "Parallel Letter Frequency",
+        "uuid": "63126f78-ba0d-4271-978d-49e9312f0db2",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 7
       }
     ]
   },
--- a/config.json
+++ b/config.json
@@ -2541,7 +2541,13 @@
         "name": "Eliud's Eggs",
         "uuid": "2a3ecf62-fb5d-4dac-8369-72e7976a8f57",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "strings",
+          "regular-expressions",
+          "rest-and-spread",
+          "arrow-functions",
+          "basics"
+        ],
         "difficulty": 2
       },
       {
@@ -2549,7 +2555,13 @@
         "name": "Parallel Letter Frequency",
         "uuid": "63126f78-ba0d-4271-978d-49e9312f0db2",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "strings",
+          "regular-expressions",
+          "rest-and-spread",
+          "arrow-functions",
+          "basics"
+        ],
         "difficulty": 7
       }
     ]
--- a/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
+++ b/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
@@ -1,0 +1,1 @@
+As a stretch goal, consider if your implementation
--- a/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
+++ b/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
@@ -17,9 +17,11 @@ Even though Javascript by default is single-threaded, there is a way to execute 
 through the [Web Worker API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API).
 
 As described by MDN:
+
 > Web Workers makes it possible to run a script operation in a background thread separate from the main execution thread of a web application.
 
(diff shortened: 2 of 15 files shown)


exercises/practice/parallel-letter-frequency/.meta/proof.ci.js
```diff
@@ -0,0 +1,19 @@
...
+const processSingleText = (text, result) => {
+  let res = [...text].reduce((acc, cur) => {
+    acc[cur] = (acc[cur] || 0) + 1;
+    return acc;
+  }, result);
+
+  return new Promise((resolve) => {
+    resolve(res);
+  });
+};
```
ErikSchierboom, line 19: I'm not a JavaScript maintainer, but to me this looks like the processing isn't actually done in parallel but calculated sequentially and then wrapped in a promise. I think if you move the processing code within the callback in the `Promise` constructor, then it would do parallel processing.

exercises/practice/pop-count/package.json
```diff
@@ -3,6 +3,7 @@
   "description": "Exercism practice exercise on pop-count",
   "author": "Katrina Owen",
   "contributors": [
+    "Cool-Katt (https://github.com/Cool-Katt)",
```
ErikSchierboom, line 6: This looks like an unrelated change

config.json
```diff
@@ -2541,8 +2541,28 @@
...
         "uuid": "2a3ecf62-fb5d-4dac-8369-72e7976a8f57",
         "practices": [],
-        "prerequisites": [],
+        "prerequisites": [
+          "strings",
+          "regular-expressions",
+          "rest-and-spread",
+          "arrow-functions",
+          "basics"
+        ],
```
ErikSchierboom, line 2550: These seem to be unrelated changes.

Cool-Katt: Last PR about `pop-count` got merged before I could add prerequisites and I didn't really want to open a separate PR for this only so I lumped these in here. I could remove them from this PR if you'd like?

exercises/practice/parallel-letter-frequency/.meta/proof.ci.js
SleeplessByte: There is no way to process things in parallel in JavaScript because JavaScript is single threaded. You _can_ however process things concurrently. Not that the code will achieve it, because all this does is spawn micro tasks on the stack and the evaluate them in order, it would be "close enough".

Moving the processing to the `Promise` constructor doesn't change anything. It just defers the processing to inside the microtask.

Cool-Katt: Yeah, I'm aware of the single thread limitation, I wrote about it a bit in the Discord and ultimately decided to cheat a little with `Promise.all()`. 
According to MDN however, you can use [Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) for multi thread processing, as each worker is spawned in a separate thread. Something like this maybe (I haven't tested this code, yet)
```js
// worker.js
onmessage = function (event) {
  let res = [...event.data].reduce((acc, cur) => {
    acc[cur] = (acc[cur] || 0) + 1;
    return acc;
  }, {});
  postMessage(res);
};
```
```js
// main.js
function workerDemo(text) { 
    const worker = new Worker('worker.js');
    const result = {}
    worker.postMessage(text);
    worker.onmessage = function (event) {
        result = event.data;
    };
    return result;
};
```

In theory, if you split the initial array in two, or however many, and then spawn a separate worker for each part, they should execute each in parallel. But my problems with `Web Workers` are that:
1. I don't know how to test them properly
2. I don't want to write tests that make it impossible to pass if you don't use `Web Workers`
3. I think that the learning curve is too big so it'll be a bit of a hurdle for most people without some further reading (speaking from my experience from mentoring, most requests i get are for simpler exercises so...)

I could (probably would) write up some explanation, about `Web Workers` but yeah, your input is greatly appreciated.

ErikSchierboom: @SleeplessByte does it still make sense to implement this exercise then?

config.json
ErikSchierboom: They should be separate PR'S yes

Cool-Katt: got it

exercises/practice/parallel-letter-frequency/parallel-letter-frequency.js
```diff
@@ -0,0 +1,8 @@
+//
+// This is only a SKELETON file for the 'Poker' exercise. It's been provided as a
```
ErikSchierboom, line 2:
```suggestion
// This is only a SKELETON file for the 'Parallel Letter Frequency' exercise. It's been provided as a
```

exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
```diff
@@ -0,0 +1 @@
+As a stretch goal, consider if your implementation
```
ErikSchierboom, line 1:
```suggestion
# Instructions append 

As a stretch goal, consider if your implementation
```

I also think that you're missing some words at the end.

ErikSchierboom (approved):
Nice!

exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
Cool-Katt: yep, should be all good now, but do tell if you think it's not suitable

exercises/practice/parallel-letter-frequency/.meta/proof.ci.js
SleeplessByte: @ErikSchierboom with the additions from @Cool-Katt I like it. It's a good primer that JS _is_ single threaded by default. I think it's fine like this.

PR #2398, opened 2024-02-09
[New Exercise] Ledger

Cool-Katt:
Do tell if you think that the example file needs more work to make it even more bad. I tried to keep it reasonable.


Cool-Katt:
And again, bump here.
I would be happy to get this merged so let me know if I need to do anything else.

exercises/practice/ledger/.meta/proof.ci.js
```diff
@@ -0,0 +1,67 @@
...
+      day: '2-digit',
+      month: '2-digit',
+      year: 'numeric',
+    })} | `;
+    // Write entry description to table
+    let truncatedDescription =
+      entry.description.length > 25
+        ? `${entry.description.substring(0, 22)}...`
+        : entry.description.padEnd(25, ' ');
+    table += `${truncatedDescription} | `;
```
SleeplessByte, line 56: I think we would add a method to `LedgerEntry` `truncatedDescription(length)`, or `displayDescription(length)`.

exercises/practice/ledger/.meta/proof.ci.js
```diff
@@ -0,0 +1,67 @@
...
+    // Write entry change to table
+    let changeStr = (entry.change / 100).toLocaleString(
+      locale,
+      formattingOptions,
+    );
```
SleeplessByte, line 61: I think we would add a method to `LedgerEntry` `displayChange(locale, formattingOptions)`

exercises/practice/ledger/.meta/proof.ci.js
```diff
@@ -0,0 +1,67 @@
...
+  );
+  // Writing each entry to table
+  entries.forEach((entry) => {
+    table += '\n';
+    // Write entry date to table
+    table += `${entry.date.toLocaleDateString(locale, {
+      day: '2-digit',
+      month: '2-digit',
+      year: 'numeric',
+    })} | `;
```
SleeplessByte, line 50: I think we would add a method to `LedgerEntry` `displayDate(locale)`

exercises/practice/ledger/.meta/proof.ci.js
```diff
@@ -0,0 +1,67 @@
...
+    // Write entry change to table
+    let changeStr = (entry.change / 100).toLocaleString(
+      locale,
+      formattingOptions,
+    );
+    let trailingSpace = changeStr.includes(')') ? '' : ' ';
+    table += `${changeStr}${trailingSpace}`.padStart(13, ' ');
+  });
+
+  return table;
```
SleeplessByte, line 66: After my comments, I actually think what I would have done if I kept with the objects is something like:

```javascript
class LedgerEntry {
  constructor(date, description, change) {
    this.date = new Date(date);
    this.description = description;
    this.change = change;
  }
}

class FormattedLedgerEntry {
  constructor(entry, locale, dateFormat, currencyFormat) {
    this.entry = entry
    this.locale = locale;
    this.dateFormat = dateFormat
    this.currencyFormat = currencyFormat;
  }

  date(format = this.dateFormat) {
    return entry.date.toLocaleDateString(this.locale, format)
  }

  description(length = 25) {
    if (entry.description.length > length ) {
      return `${entry.description.substring(0, length - 3)}...`
    }

    return entry.description.padEnd(length, ' ');
  }
  
  change(offset = 13, format = this.currencyFormat) {
    const formatted = (entry.change / 100).toLocaleString(locale, format);
    
    const trailingSpace = formatted.includes(')') ? '' : ' ';
    `${formatted }${trailingSpace}`.padStart(offset, ' ');
  }

  toTableRow() {
    return [
      this.date(),
      this.description(),
      this.change()
    ].join(' | ')
  }
}

// Then:
const dateFormat = ...
const currencyFormat = ...

const rows = entries.map((entry) => {
  const formattedEntry = new FormattedLedgerEntry(entry, locale, dateFormat, currencyFormat))
  return entry.toTableRow()
})

return [
  headerRow(locale),
  ...rows
].join("\n")
```

Or something like that. 

Probably makes it much easier to maintain.

SleeplessByte (commented):
I love to hate it <3

exercises/practice/ledger/.meta/proof.ci.js
Cool-Katt: I think I like your solution way more than mine 😁, so with your permission I'd borrow some of it, tho maybe when this exercise goes live.

I also tried to keep it functional, but I don't think it makes sense, particularly since I'm still using a `class`.

SleeplessByte: Ya go for it! Use whatever you like. I was mostly just thinking of what I would've done

Cool-Katt:
@SleeplessByte 
I know it doesn't make a difference for the actual exercise, but I updated the `proof.ci.js` based on what you suggested.
I'm not sure it's cleaner but it looks more uniform at least.

SleeplessByte:
Oh yeah it's just a proof but I do like it better like this 😁

SleeplessByte (approved):

PR #2418, opened 2024-03-19
Allow synchronous and concurrent and parallel solutions to Parallel Letter Frequency

mk-mxp:
This change allows students to solve the exercise in a multitude of concurrent, parallel and synchronous ways. I did not change the example code to proof that one can still solve the exercise synchronously.

See https://forum.exercism.org/t/issues-with-parallel-letter-frequency-in-javascript/10290

I also added some more clarifications about concurrent and parallel execution in JavaScript. And a word of warning to the `Worker` part. I may re-write that part for Node.js, if required.

As a proof for async execution using concurrent promises I used this code. It uses the same external accumulator as the synchronous solution, but returns unresolved promises and resolves later:

```javascript
export const parallelLetterFrequency = async (texts) => {
    const frequencies = {};
    return Promise.all(texts.map((text) => letterFrequency(frequencies, text))).then(() => frequencies);
};

const letterFrequency = (frequencies, text) => {
    return new Promise((resolve) => {
        setTimeout(() => {
            for (const [letter] of text.toLowerCase().match(/\p{Letter}/gu) || [])
                frequencies[letter] = (frequencies[letter] || 0) + 1;

            resolve(frequencies);
        }, 1);
    });
};
```

I also have a `Worker` implementation doing the things required for real parallelism, which passes the changed tests:

```javascript
import { Worker } from "node:worker_threads";

const workerCode = `
const { parentPort } = require('node:worker_threads');

function countLetters(text) {
    return [...text].reduce(handleLetter, {});
}

function handleLetter(accumulator, letter) {
    if (/\\P{L}/iu.test(letter))
        return accumulator;
    letter = letter.toLowerCase();
    accumulator[letter] = (accumulator[letter] || 0) + 1;
    return accumulator;
}

parentPort?.on('message', function (message) {
    parentPort?.postMessage(countLetters(message));
});
`;

export const parallelLetterFrequency = async (texts) => {
    return Promise
        .all(texts.map((line) => countForOneLine(line)))
        .then((results) => results.reduce(cumulate, {}))
        .catch((reason) => console.log(reason))
        ;
};

function countForOneLine(text) {
    return new Promise((resolve, reject) => {
        const worker = new Worker(workerCode, {eval: true});
        worker.on('error', (error) => reject(error));
        worker.on('exit', (reason) => reject('Exit. ' + reason));
        worker.on('message', (message) => {
            worker.unref();
            resolve(message);
        });
        worker.on('online', () => {
            worker.postMessage(text);
        });
    });
}

function cumulate(accumulator, result) {
    for (const letter in result) {
        accumulator[letter] = (accumulator[letter] || 0) + result[letter];
    }

    return accumulator;
}
```

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,6 @@
 /.github/labels.yml
 /.github/workflows/sync-labels.yml
+/.github/workflows/no-important-files-changed.yml
 exercises/**/README.md
 !/README.md
 
--- a/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
+++ b/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
@@ -1,8 +1,23 @@
 # Instructions append
 
 Due to the single-threaded nature of Javascript, code that appears to execute in parallel,
-such as `async functions` or `Promises`, actually execute concurrently.
-Such solutions will pass all the tests, even though they do not meet the requrement for parallel execution.
+such as `async functions` or `Promise`, actually execute concurrently at most.
+In fact, all code is executed synchronously even in `Promise` constructors, until control is passed back to the Event-Loop (the outermost function call ends).
+Such solutions will pass all the tests, even though they do not meet the requirement for parallel or concurrent execution.
+
+## Achieving concurrent execution for the exercise
+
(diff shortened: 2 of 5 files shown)


SleeplessByte:
Thank you for the work regardless!

PR #2538, opened 2024-09-12
[factory-sensors] Add test for temperature of 0 degrees

themetar:
This exercise should reject solutions with `!temperature`.

1) In the story `null` is a special signal that the sensor is broken. That means in-universe the sensor will needlessly be replaced if it happens to encounter temperatures of 0 degrees.

2) In our world, the assignment requires that students make distinction between nulls and numbers. 0 is a valid number. There is no reason to have it _intentionally_ be handled differently than -1 or 1 and lumped with `null`.

This is supported by the existing exemplar solution; it uses `if (temperature === null)`

### DIFF
--- a/exercises/concept/factory-sensors/.meta/config.json
+++ b/exercises/concept/factory-sensors/.meta/config.json
@@ -4,7 +4,8 @@
   ],
   "contributors": [
     "SleeplessByte",
-    "junedev"
+    "junedev",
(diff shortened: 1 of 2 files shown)


SleeplessByte:
Yah this makes a lot of sense.

SleeplessByte (approved):

PR #2577, opened 2025-01-04
Updating tests for armstrong-numbers

jagdish-15:
# Pull Request

This PR adds two new test cases to the Armstrong-Numbers exercise to align it with the `problem-specifications` repository. However, both numbers in these new tests are extremely large and will cause the `test-runner` to fail. Please let me know how you'd like me to proceed with this!

### DIFF
--- a/exercises/practice/armstrong-numbers/.docs/instructions.append.md
+++ b/exercises/practice/armstrong-numbers/.docs/instructions.append.md
@@ -1,2 +1,4 @@
-Note: Some of the tests might pass a `BigInt` as input.
+~~~~exercism/note
+Some of the tests might pass a `BigInt` as input.
 Ensure that your implementation can handle such cases.
+~~~~
--- a/exercises/practice/armstrong-numbers/.docs/instructions.append.md
+++ b/exercises/practice/armstrong-numbers/.docs/instructions.append.md
@@ -1,4 +1,6 @@
+<!-- prettier-ignore-start -->
 ~~~~exercism/note
 Some of the tests might pass a `BigInt` as input.
 Ensure that your implementation can handle such cases.
 ~~~~
+<!-- prettier-ignore-end -->
--- a/exercises/practice/armstrong-numbers/.docs/instructions.append.md
+++ b/exercises/practice/armstrong-numbers/.docs/instructions.append.md
@@ -1,0 +1,2 @@
+Note: Some of the tests might pass a `BigInt` as input.
+Ensure that your implementation can handle such cases.
--- a/exercises/practice/armstrong-numbers/.meta/tests.toml
+++ b/exercises/practice/armstrong-numbers/.meta/tests.toml
@@ -1,30 +1,43 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
 
 [c1ed103c-258d-45b2-be73-d8c6d9580c7b]
 description = "Zero is an Armstrong number"
 
 [579e8f03-9659-4b85-a1a2-d64350f6b17a]
-description = "Single digit numbers are Armstrong numbers"
+description = "Single-digit numbers are Armstrong numbers"
 
 [2d6db9dc-5bf8-4976-a90b-b2c2b9feba60]
-description = "There are no 2 digit Armstrong numbers"
+description = "There are no two-digit Armstrong numbers"
 
 [509c087f-e327-4113-a7d2-26a4e9d18283]
-description = "Three digit number that is an Armstrong number"
+description = "Three-digit number that is an Armstrong number"
 
 [7154547d-c2ce-468d-b214-4cb953b870cf]
-description = "Three digit number that is not an Armstrong number"
+description = "Three-digit number that is not an Armstrong number"
 
 [6bac5b7b-42e9-4ecb-a8b0-4832229aa103]
-description = "Four digit number that is an Armstrong number"
+description = "Four-digit number that is an Armstrong number"
 
 [eed4b331-af80-45b5-a80b-19c9ea444b2e]
-description = "Four digit number that is not an Armstrong number"
+description = "Four-digit number that is not an Armstrong number"
 
 [f971ced7-8d68-4758-aea1-d4194900b864]
-description = "Seven digit number that is an Armstrong number"
+description = "Seven-digit number that is an Armstrong number"
 
 [7ee45d52-5d35-4fbd-b6f1-5c8cd8a67f18]
-description = "Seven digit number that is not an Armstrong number"
+description = "Seven-digit number that is not an Armstrong number"
+
+[5ee2fdf8-334e-4a46-bb8d-e5c19c02c148]
+description = "Armstrong number containing seven zeroes"
+
+[12ffbf10-307a-434e-b4ad-c925680e1dd4]
+description = "The largest and last Armstrong number"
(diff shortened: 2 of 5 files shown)


exercises/practice/armstrong-numbers/armstrong-numbers.spec.js
```diff
@@ -37,4 +37,12 @@ describe('Armstrong Numbers', () => {
   xtest('Seven digit number that is not an Armstrong number', () => {
     expect(isArmstrongNumber(9926314)).toEqual(false);
   });
+
+  xtest('Armstrong number containing seven zeroes', () => {
```
SleeplessByte, line 41: What's the runtime for these two for the proof.ci solution on your machine?

SleeplessByte: FYI the tests are failing because that's not a valid integer. You probably need to switch to [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) for these tests.

jagdish-15: Yep, I’m aware of that, and that’s why comments are included in the test cases’ `canonical-data.json` file in the `problem-specifications` repository, indicating that these tests can be skipped if appropriate.  

```
    {
      "uuid": "5ee2fdf8-334e-4a46-bb8d-e5c19c02c148",
      "description": "Armstrong number containing seven zeroes",
      "comments": [
        "The numeric size of this input number is 108 bits. Consider skipping this test if appropriate."
      ],
      "scenarios": ["big-integer"],
      "property": "isArmstrongNumber",
      "input": {
        "number": 186709961001538790100634132976990
      },
      "expected": true
    },
    {
      "uuid": "12ffbf10-307a-434e-b4ad-c925680e1dd4",
      "description": "The largest and last Armstrong number",
      "comments": [
        "The numeric size of this input number is 127 bits. Consider skipping this test if appropriate."
      ],
      "scenarios": ["big-integer"],
      "property": "isArmstrongNumber",
      "input": {
        "number": 115132219018763992565095597973971522401
      },
      "expected": true
    }
```

I was wondering if we should include these tests or not because doing so might cause all community solutions to become outdated and would require new students to use `BigInt`. This is the concern I was referring to in my first message—apologies for not being clearer earlier! Let me know your thoughts.

Cool-Katt: Well, updating the tests causes all solutions to be rerun again so they will become outdated anyway. You are correct that the current solutions that don't use BigInt will be invalid.

I feel that if there's an easy way to implement a solution that would pass all the tests, including the new ones, then they should be included, so if you can figure it out, feel free to commit.

jagdish-15: I've got it working using the BigInt approach, but we can't directly pass those large numbers as they are in the test file. Instead, we need to pass them either as strings or as BigInt values. I've demonstrated both approaches in the second-to-last and last tests, respectively. Please review them and let me know which approach we should proceed with.

Cool-Katt (approved):
Looks good to me, and it passes the tests!

jagdish-15:
@Cool-Katt, the last two tests are not consistent with each other yet. Let me know which approach you'd prefer me to implement for both.

Cool-Katt:
I'm thinking that the one that explicitly passes a `BigInt` as input is better, since we want the tests to serve as a sort of hint for people that get stuck. @jagdish-15

jagdish-15:
Sure @Cool-Katt, I'll make the changes ASAP!

exercises/practice/armstrong-numbers/.meta/tests.toml
```diff
@@ -1,30 +1,43 @@
...
-description = "Four digit number that is not an Armstrong number"
+description = "Four-digit number that is not an Armstrong number"
 
 [f971ced7-8d68-4758-aea1-d4194900b864]
-description = "Seven digit number that is an Armstrong number"
+description = "Seven-digit number that is an Armstrong number"
 
 [7ee45d52-5d35-4fbd-b6f1-5c8cd8a67f18]
-description = "Seven digit number that is not an Armstrong number"
+description = "Seven-digit number that is not an Armstrong number"
```
SleeplessByte, line 37: Other test descriptions needs to be updated to match this.

exercises/practice/armstrong-numbers/armstrong-numbers.spec.js
SleeplessByte: We barely have any `BigInt` exercises, so let's just upgrade everything.

I prefer it if you pass the numbers into the tests with the `n` suffix:

```javascript
const largeNumber = 186709961001538790100634132976990n
expect(isArmstrongNumber(largeNumber)).toEqual(true);
```

exercises/practice/armstrong-numbers/armstrong-numbers.spec.js
```diff
@@ -37,4 +37,15 @@ describe('Armstrong Numbers', () => {
...
+    const largeNumber = BigInt(`186709961001538790100634132976990`);
+    expect(isArmstrongNumber(largeNumber)).toEqual(true);
+  });
+
+  xtest('The largest and last Armstrong number', () => {
+    expect(
+      isArmstrongNumber(`115132219018763992565095597973971522401`),
```
SleeplessByte, line 48:
```suggestion
      isArmstrongNumber(115132219018763992565095597973971522401n),
```

exercises/practice/armstrong-numbers/armstrong-numbers.spec.js
SleeplessByte: I would also recommend to add an `appends` file to note that some of the tests will pass a `BigInt`. Your proof will keep working because:

```javascript
const bigInput = BigInt(42);
// 42n

const bigInput = BigInt(42n);
// 42n
```

SleeplessByte (changes requested):

SleeplessByte:
Certainly 

This exercise is synced with the problem-specifications, so when a change is introduced there, we receive it via pull request.

This will replace the `instructions.md` as expected.

We now want to add a note to that file that the solution needs to be able to handle `BigNum` numbers. We cannot add it to the instructions because it would eventually be overwritten.

However:
 
> `.docs/instructions.append.md`: additional introduction text to append after the existing instructions (optional)

Edit or create this file for this exercise and it will remain in place when the main instructions are updated.

jagdish-15:
Sure, I'll make the changes as soon as possible! Additionally, I had a question unrelated to the PR, so I’ve sent an email to the address listed on your GitHub profile since GitHub doesn’t support direct messaging. I’d really appreciate it if you could take a look. Thank you!

SleeplessByte:
> Sure, I'll make the changes as soon as possible! Additionally, I had a question unrelated to the PR, so I’ve sent an email to the address listed on your GitHub profile since GitHub doesn’t support direct messaging. I’d really appreciate it if you could take a look. Thank you!

Maybe you want to send me another one including your GitHub username? I cannot find anything from ~3 days ago that could be from you (but I also get quite a bit of email...)

SleeplessByte:
> Sure, I'll make the changes as soon as possible! Additionally, I had a question unrelated to the PR, so I’ve sent an email to the address listed on your GitHub profile since GitHub doesn’t support direct messaging. I’d really appreciate it if you could take a look. Thank you!

I found it!

exercises/practice/armstrong-numbers/.docs/instructions.append.md
```diff
@@ -0,0 +1,2 @@
+Note: Some of the tests might pass a `BigInt` as input.
```
Cool-Katt, line 1: I would prefer it if this was an exercism/note block.
See [this](https://exercism.org/docs/building/markdown/markdown#h-special-blocks-sometimes-called-admonitions)

jagdish-15: I've added the `exercism/note` block, but when I ran the format file, it attempted to replace it with backticks. This seems to be causing the current issue with the CL.

Cool-Katt: should be fixed now.

SleeplessByte: The backticks also should still work, I think!

jagdish-15: > should be fixed now.

It was fixed, Thank you so much!

jagdish-15:
Hi @SleeplessByte and @Cool-Katt,

Please let me know if there's anything else you'd like me to address or adjust in this PR. I'm happy to make any necessary updates!

SleeplessByte (approved):

PR #2622, opened 2025-04-30
Update instructions.md

sparky9043:
Change wording to clarify that there are 22 billable days per month and the days given for each project are all billable days


iHiD:
(This has come out of a conversation with me)

Cool-Katt (approved):

Cool-Katt:
I was going to mention that this is the second instruction clarification for this exercise in a span of two weeks, and suggest we put the new instructions in `problem-spec` but I don't think this exercise exists there (or if it does, it's with a different name?)
@iHiD

PR #2647, opened 2025-06-05
Syncing test.toml and updating the test code for largest-series-product

jagdish-15:
# Pull Request

### DIFF
--- a/exercises/practice/largest-series-product/.meta/proof.ci.js
+++ b/exercises/practice/largest-series-product/.meta/proof.ci.js
@@ -3,7 +3,7 @@ export const largestProduct = (digits, seriesLength) => {
     return 1;
   }
   if (seriesLength > digits.length) {
-    throw new Error('span must be smaller than string length');
+    throw new Error('span must not exceed string length');
   }
   if (seriesLength < 0) {
     throw new Error('span must not be negative');


jagdish-15:
@SleeplessByte @Cool-Katt 

Do you think it's worth changing the error comments, considering many existing solutions on the track might fail if we go ahead with this? 

Lemme know so I can proceed with the proof solution!

SleeplessByte:
> @SleeplessByte @Cool-Katt 
> 
> Do you think it's worth changing the error comments, considering many existing solutions on the track might fail if we go ahead with this? 
> 
> Lemme know so I can proceed with the proof solution!

What does "changing the error comments" mean?

jagdish-15:
By error comment I was referring to the message inside `new Error()`. This sync only changes that message for two tests, so I’m wondering if it’s worth updating them, especially since all the old solutions fail now due to these new messages differing from the original ones.

SleeplessByte:
Yes, being in sync with spec removes bike shedding. Check out this message: https://github.com/exercism/javascript/pull/2647#issuecomment-2945718121

If we merge with that, no solutions will be retested.

exercises/practice/largest-series-product/.meta/tests.toml
```diff
@@ -38,6 +38,11 @@ description = "reports zero if all spans include zero"
 
 [5d81aaf7-4f67-4125-bf33-11493cc7eab7]
 description = "rejects span longer than string length"
+include = false
```
SleeplessByte, line 41: What does this mean?

jagdish-15: This means we are not gonna include this test in the test file. This is often used when tests are reimplemented

Cool-Katt:
I feel that tagging this with [no important files changed] may be okay, to not rerun all the current solutions. 

Other option would be to simply test it the function throws anything, which I don't think is okay, or to test if it throws an instance of some error class (that would have to be written because it currently doesn't exist)

jagdish-15:
Tagging it makes more sense to me for now!

Cool-Katt (approved):

PR #2648, opened 2025-06-06
Syncing test.toml and updating the test code for reverse-string

jagdish-15:
# Pull Request

### DIFF
--- a/exercises/practice/reverse-string/reverse-string.spec.js
+++ b/exercises/practice/reverse-string/reverse-string.spec.js
@@ -37,4 +37,22 @@ describe('ReverseString', () => {
     const actual = reverseString('drawer');
     expect(actual).toEqual(expected);
   });
+
+  xtest('wide characters', () => {
+    const expected = '猫子';
+    const actual = reverseString('子猫');
+    expect(actual).toEqual(expected);
+  });
+
+  xtest('grapheme cluster with pre-combined form', () => {
+    const expected = 'dnatsnehctsrüW';
+    const actual = reverseString('Würstchenstand');
+    expect(actual).toEqual(expected);
+  });
+
+  xtest('grapheme clusters', () => {
+    const expected = 'มรกแรปโนยขีเผู้';
+    const actual = reverseString('ผู้เขียนโปรแกรม');
+    expect(actual).toEqual(expected);
+  });
 });
--- a/exercises/practice/reverse-string/.meta/config.json
+++ b/exercises/practice/reverse-string/.meta/config.json
@@ -5,6 +5,7 @@
   "contributors": [
     "ankorGH",
     "d-vail",
+    "jagdish-15",
     "ovidiu141",
     "SleeplessByte"
   ],
--- a/exercises/practice/reverse-string/.meta/tests.toml
+++ b/exercises/practice/reverse-string/.meta/tests.toml
@@ -1,6 +1,13 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
 
 [c3b7d806-dced-49ee-8543-933fd1719b1c]
 description = "an empty string"
@@ -19,3 +26,12 @@ description = "a palindrome"
 
 [b9e7dec1-c6df-40bd-9fa3-cd7ded010c4c]
 description = "an even-sized word"
+
+[1bed0f8a-13b0-4bd3-9d59-3d0593326fa2]
+description = "wide characters"
+
+[93d7e1b8-f60f-4f3c-9559-4056e10d2ead]
+description = "grapheme cluster with pre-combined form"
+
+[1028b2c1-6763-4459-8540-2da47ca512d9]
+description = "grapheme clusters"


jagdish-15:
This change introduces grapheme cluster & wide character handling to the exercise, which would require updating the proof solution accordingly. We had previously decided to skip such changes in the C track, see [this PR](https://github.com/exercism/c/pull/1037) for context. Could you please confirm if we’re moving forward with this change here?

Cool-Katt:
Well ,there is a way to work with graphemes in JS, but most of the default behaviour doesn't account for it so if we add these tests I can guarantee you that almost all the current solution will fail. 
Do we expect students to be able to deal with such weirdness at the level that this exercise is? If we do end up adding the tests it'll probably be good to also amend the instructions with some sort of explanation. @SleeplessByte opinions?

----

I had to deal with this in `micro-blog` and i wrote up an entire Dig Deeper section about it, so maybe check that if you want to know more of the technical angle.

SleeplessByte:
I haven't looked at the current proof. Is this updating from `.split('')` to `[...x]` / `Array.from(x)`?

If yes, we should do this. People shouldn't split into characters like the first one anyway. If no, and it is harder, we can decide between not including or making them optional. We have a special flag and syntax to allow for that.

jagdish-15:
I'll look into it this Monday or Tuesday!

jagdish-15:
I got it running after checking out the `Intl.Segmenter` object from the Dig Deeper section by @Cool-Katt! 

> Do we expect students to be able to deal with such weirdness at the level that this exercise is? If we do end up adding the tests it'll probably be good to also amend the instructions with some sort of explanation.

Now that's a topic of discussion. What do you think @SleeplessByte @Cool-Katt

jagdish-15:
Since this is a practice exercise, I think it’s totally fine to include these tests. If we decide to go ahead, we’ll just need to add an `instructions.append.md` file and update the difficulty from **Easy** to **Medium**.

Alternatively, we can always choose to skip the tests.  
Let me know what you think!

SleeplessByte:
Can you try if `let characters = [...string]` works ?

Cool-Katt:
Well, I checked the current solutions and I don't really know how many total submissions there are, but exactly two of them use `Intl.Segmenter`, ~and one of the two is kinda "cheety" by using `.reverse()` directly~, so I can tell you that this change will cause some confusion. 

This discussion may need to happen on the forum to get more voices on the matter, so feel free to open a thread and link this PR there @jagdish-15. Presonaly, I'm tentatively for a merge and bumping the difficulty.

Cool-Katt:
@SleeplessByte If you mean something like this:
```js
const expected = 'dnatsnehctsrüW';
const actual = 'Würstchenstand';
console.log([...actual].reverse().join('').localeCompare(expected))
// => 1 (not the same)
```
This definitely doesn't work, in the test with `'Würstchenstand'`, it puts the umlaut (the two dots on top of the `'u'`) on the `'r'` instead, which is also frustrating because it visually is hard to tell the difference.

SleeplessByte:
Ok perfect. 

I dont think we should make this exercise more difficult. I rather have an exercise that goes further exploring this concept.

Cool-Katt:
We do (kind of)! `micro-blog` is dealing with this weirdness.

I'm okay with not complicating things more for this exercise, as it is a more beginner friendly exercise.

SleeplessByte:
> We do (kind of)! `micro-blog` is dealing with this weirdness.
> 
> I'm okay with not complicating things more for this exercise, as it is a more beginner friendly exercise.

Ah yeah, I remember that one!

@jagdish-15 can you remove the tests that require the segmenter?

jagdish-15:
> Can you try if `let characters = [...string]` works ?

I did try it, but it didn't work, as @Cool-Katt showed

> @jagdish-15 can you remove the tests that require the segmenter?

Should I remove it altogether or skip them using `test.skip(...)`?

SleeplessByte:
> Should I remove it altogether or skip them using `test.skip(...)`?

Oh right. We have that option. I guess we can add a track insert (instruction addendum):

````markdown
~~~exercism/advanced
If you solve this using the CLI, there are test cases that require you to deal with complex characters.
You can optionally enable these tests by removing `.skip` from the test.
~~~
````

@Cool-Katt what do you think?

**Comment by jagdish-15** [comment]
> I guess we can add a track insert (instruction addendum):

I vaguely remember seeing something similar already implemented in another exercise. In that case, comments were added above the skipped tests to inform CLI users that the tests are skipped by default (since most solutions would time out), but they can enable them manually if they want to experiment.

So I think staying consistent with that approach would be a good idea! But we'll see what @Cool-Katt says

**Comment by Cool-Katt** [comment]
Yep, I agree with `skip` and an addendum.

**Comment by jagdish-15** [comment]
I was discussing this exercise, [pythagorean-triplet test file](https://github.com/exercism/javascript/blob/main/exercises/practice/pythagorean-triplet/pythagorean-triplet.spec.js#L64). It has skipped tests, but there is no addendum referring to that. Should we append instructions there as well, to address the skipped tests?

**Comment by Cool-Katt** [comment]
Yep.
People rarely ever read the test file separately so I feel that the note that DJ suggested is good to include.

**Comment by jagdish-15** [comment]
I've added the part suggested by @SleeplessByte and raised another PR (#2671) to do something similar for the `reverse-string` exercise. @Cool-Katt and @SleeplessByte take a look and lemme know if it works!

File: exercises/practice/reverse-string/reverse-string.spec.js [hunk 1]
```diff
@@ -37,4 +37,22 @@ describe('ReverseString', () => {
     const actual = reverseString('drawer');
     expect(actual).toEqual(expected);
   });
+
+  xtest('wide characters', () => {
+    const expected = '猫子';
+    const actual = reverseString('子猫');
+    expect(actual).toEqual(expected);
+  });
```
SleeplessByte, line 45: This one passes with the previous solution? If yes, this is good, if no it also needs to be skipped :)

jagdish-15: Yes it does indeed pass with the previous solution (I remember it did pass with  `[...string]`) but I'll still check it once again to be sure

jagdish-15: Yes, the old solution passes the test as well, so there’s no need to skip it.
By the way, should we stick with the segmenter solution or revert to the original?
I think keeping the segmenter version is a better choice, especially since it also passes the skipped tests.

Cool-Katt: The proof solution is more for the CI, to prove that the tests can be passed, so the segmenter version can stay.

devcyjung:
@SleeplessByte This PR seems to be causing error in online editor

I suspect changing test.skip to xtest would fix it

Error message:

We received the following error when we ran your code:

Expected to see 0 skipped tests and 0 skipped test suites. None of the tests in this exercise are optional. The skipped tests will not show up, but were found during the last run.

jagdish-15:
Yes, we do receive the error message, though it's unclear why. We use a similar test-skipping pattern in the [pythagorean-triplet test file](https://github.com/exercism/javascript/blob/main/exercises/practice/pythagorean-triplet/pythagorean-triplet.spec.js#L64), and that one doesn’t seem to cause any problems.

> I suspect changing `test.skip` to `xtest` would fix it.

That’s probably not the issue. We already use `xtest` for the regular cases as part of the test runner’s setup — only the first test is defined using `test`, and the rest rely on `xtest` by design.


@Cool-Katt @SleeplessByte, got any clue?

devcyjung:
the way I see it is test.skip is for computationally heavy test cases that would take too long to complete in online environment, otherwise xtest is used

If you look at the scripts there are specific commands such as

sed -i -e 's/xtest/test/g' "${test_file}"

which replace xtest with test before running the tests

Also this is where the error is coming from

https://github.com/exercism/js-test-runner-clientside/blob/94e033dd93e9260b64dabe4ccef315d20a028b7a/src/output.ts#L40

You should either flag it as optional test or use the xtest @jagdish-15

SleeplessByte:
It should use `test.skip`. 

It's possible that we just got the new online runner. I just haven't enabled the online user to be able to handle this. If that's the case (I'll check) and because it takes a few days to make changes to the online editor, we'll comment out the test for now.

If that's not the issue the. It should still use `test.skip` and something else is going on.

jagdish-15:
The [`pythagorean-triplet`](https://github.com/exercism/javascript/blob/main/exercises/practice/pythagorean-triplet/pythagorean-triplet.spec.js) tests are working fine, even though they also use `test.skip()`, so the issue seems to be specific to our current case.

Still, should I go ahead and push the changes after fully commenting out the code?

SleeplessByte:
> The [`pythagorean-triplet`](https://github.com/exercism/javascript/blob/main/exercises/practice/pythagorean-triplet/pythagorean-triplet.spec.js) tests are working fine, even though they also use `test.skip()`, so the issue seems to be specific to our current case.
> 
> Still, should I go ahead and push the changes after fully commenting out the code?

Nah, I am investigating now as you mentioned `test.skip` is fine.

SleeplessByte:
@jagdish-15 the issue is `flag.tests.includes-optional` is not set to true in the config, which is what @devcyjung also mentioned in the last sentence. Can you PR that so I can merge it?

jagdish-15:
@SleeplessByte 
I've raised the PR #2682 (please discard the earlier one, sorry for that)

PR #2651, opened 2025-06-06
Syncing test.toml and updating the test code and the proof solutions for sublist

jagdish-15:
# Pull Request

This PR syncs `tests.toml` with the `problem-specifications` repository and verifies the presence of new tests in the test files and updates the proof solution.

### DIFF
--- a/exercises/practice/sublist/.meta/config.json
+++ b/exercises/practice/sublist/.meta/config.json
@@ -4,6 +4,8 @@
   ],
   "contributors": [
     "ankorGH",
+    "atk",
+    "jagdish-15",
     "rchavarria",
     "SleeplessByte",
     "tejasbubane",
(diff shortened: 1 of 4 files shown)


jagdish-15:
I've added @atk as a contributor in the `config.json` file, since I included their solution as the proof solution for this exercise.

The earlier proof solution was failing the newly added test.  
@Cool-Katt, your solution would also fail this test, as it relies on converting the input to a string. In the new test case, the string representation happens to match as a substring, but the actual list does not, which causes the failure.  

You can refer to the new test case for more details on the edge case it catches.

exercises/practice/sublist/.meta/config.json
```diff
@@ -4,6 +4,8 @@
   ],
   "contributors": [
     "ankorGH",
+    "atk",
```
Cool-Katt, line 7: EDIT: I read the comment you left *after* writing my own. Sry

PR #2653, opened 2025-06-06
Syncing test.toml and updating the test code and the proof solutions for word-count

jagdish-15:
# Pull Request

### DIFF
--- a/exercises/practice/word-count/word-count.spec.js
+++ b/exercises/practice/word-count/word-count.spec.js
@@ -80,8 +80,11 @@ describe('countWords', () => {
       laugh: 1,
       then: 1,
       cry: 1,
+      "you're": 1,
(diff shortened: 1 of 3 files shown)


exercises/practice/word-count/word-count.spec.js
```diff
@@ -132,4 +135,12 @@ describe('countWords', () => {
     };
     expect(countWords(",\n,one,\n ,two \n 'three'")).toEqual(expectedCounts);
   });
+
+  xtest('alternating word separators not detected as a word', () => {
```
Cool-Katt: Wrong test name here

jagdish-15: Sorry, will check it out shortly!

PR #2675, opened 2025-06-13
Update tests flatten array

jagdish-15:
# Pull Request


SleeplessByte (approved):
Great.

I feel that we don't need to re-run the tests on this one for the few people who passed the big null list and not the smaller ones, which would be a weird solution to begin with.

PR #2677, opened 2025-06-14
Update tests forth

jagdish-15:
# Pull Request

### DIFF
--- a/exercises/practice/forth/.meta/proof.ci.js
+++ b/exercises/practice/forth/.meta/proof.ci.js
@@ -67,6 +67,9 @@ export class Forth {
 
   performCommand(command) {
     if (command.arity > this.stack.length) {
+      if (this.stack.length == 1) {
+        throw new Error('Only one value on the stack');
+      }
       throw new Error('Stack empty');
     }
 


jagdish-15:
Additionally, a few test cases in `problem-specifications` expect the error message `"only one value in stack"` for certain scenarios. Currently, the implementation uses the same `"stack empty"` message for both empty stack and single-value stack cases.

Updating this would require changes to the relevant tests and the proof solution implementation. Since it's outside the scope of this PR (which focuses on syncing new tests), I haven't included it here.

If this change makes sense, I'm happy to raise a separate PR to refactor the tests and update the proof accordingly, just let me know.

SleeplessByte:
> Updating this would require changes to the relevant tests and the proof solution implementation. Since it's outside the scope of this PR (which focuses on syncing new tests), I haven't included it here.
> 

It does seem that the sync makes it relevant as it touches for example

```toml
description = "division -> errors if there is only one value on the stack"
```

I wonder why we chose to implement this as stack empty, but it does _seem_ like it should be stack one.

jagdish-15:
Sure, Imma make those changes in the test file and update the proof solution!

Cool-Katt (approved):

Cool-Katt:
@jagdish-15 That'll be a separate PR or you'll commit to this one? 
I'll hold the merge for now, I suppose.

jagdish-15:
@SleeplessByte said that this touches the examples with `one value on stack`, so I suppose I should commit in this one itself

SleeplessByte:
Yeah easier to do it here!

jagdish-15:
@SleeplessByte @Cool-Katt 
Mind taking a look?

jagdish-15:
I think this should be merged with `[no important files changed]`! That way, the old solutions do not fail, and the changes stay relevant only to future students.

Cool-Katt:
Well, an old solution will fail regardless, if the students decide to retest, since they'll see that the exercise was updated

PR #2679, opened 2025-06-15
Simplify objects concept

SleeplessByte:
[Raised in 2023 by @safwansamsudeen](https://forum.exercism.org/t/task-no-6-in-high-score-board-tests-using-functions-as-object-props-but-has-this-been-covered/2659), functions as first-class citizens and higher order functions have not been introduced at this point in the syllabus yet.


SleeplessByte:
**Merge with `[no important files changed]`!**

PR #2686, opened 2025-06-19
Remove nested describes to unbreak task-based testing

IsaacG:
Fix issue where task based outputs do not work due to nested `describe()` calls. See [Discord discussion](https://discord.com/channels/854117591135027261/1385322924428230816).

PR #2690, opened 2025-06-22
Update tests kindergarten garden

jagdish-15:
# Pull Request

### DIFF
--- a/exercises/practice/kindergarten-garden/kindergarten-garden.spec.js
+++ b/exercises/practice/kindergarten-garden/kindergarten-garden.spec.js
@@ -1,239 +1,241 @@
 import { describe, expect, test, xtest } from '@jest/globals';
 import { Garden } from './kindergarten-garden';
 
-describe('partial Garden', () => {
-  test('garden with single student', () => {
-    expect(new Garden('RC\nGG').plants('Alice')).toEqual([
-      'radishes',
-      'clover',
-      'grass',
-      'grass',
-    ]);
-  });
-
-  xtest('different garden with single student', () => {
-    expect(new Garden('VC\nRC').plants('Alice')).toEqual([
-      'violets',
-      'clover',
-      'radishes',
-      'clover',
-    ]);
-  });
-
-  xtest('garden with two students', () => {
-    expect(new Garden('VVCG\nVVRC').plants('Bob')).toEqual([
-      'clover',
-      'grass',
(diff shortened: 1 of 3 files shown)


exercises/practice/kindergarten-garden/kindergarten-garden.spec.js
```diff
@@ -1,239 +1,241 @@
...
-  test('garden with single student', () => {
-    expect(new Garden('RC\nGG').plants('Alice')).toEqual([
-      'radishes',
-      'clover',
-      'grass',
-      'grass',
-    ]);
-  });
+describe('Kindergarten Garden', () => {
+  describe('partial Garden', () => {
```
Cool-Katt: Nested `describe` blocks aren't supported currently.

Cool-Katt (changes requested):

exercises/practice/kindergarten-garden/kindergarten-garden.spec.js
jagdish-15: But exercises like `high-scores`, `forth`, and several others already use nested `describe` blocks. Am I missing something here?

Is the concern about the depth of nesting, like these having only two levels, while this has three? Just trying to understand if that’s the distinction you’re referring to.

Cool-Katt: It's an old format that we no longer support. 
Recently this broke `freelance-rates` (see #2686) so I'd like to avoid such cases in the future.

It's not that it's illegal to write nested describe blocks, jest allows it, but the current iteration of our test runner is unhappy with this kind of thing, so we specifically request one describe block per group of tests.

jagdish-15: Sure, that makes sense!  
I’ll make sure to avoid using nested `describe` blocks if I come across them in the future. And make changes in this one!

Cool-Katt:
@jagdish-15 did you commit the tests file (or make changes to it)? Not seeing anything added to it.

jagdish-15:
Ah, the changes might not be visible since the tests were already in sync, it was just the `toml` file that was out of date. I’ve also removed myself from the contributors list, as the change was small!

jagdish-15:
@Cool-Kat @SleeplessByte 
I think this can be merged!

PR #2698, opened 2025-06-25
Add practice and prerequisites to "dnd-character" exercise

quintuple-mallard:
This PR adds "randomness" as a practice and "randomness" and "classes" as prerequisites to the D&D Character exercise on the JavaScript track.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2229,9 +2229,7 @@
         "slug": "dnd-character",
         "name": "D&D Character",
         "uuid": "b373e13c-f179-4b36-b6e8-2a0f41540344",
-        "practices": [
-          "randomness"
-        ],
+        "practices": [],
         "prerequisites": [
           "randomness",
           "classes"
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -2229,13 +2229,14 @@
         "slug": "dnd-character",
         "name": "D&D Character",
         "uuid": "b373e13c-f179-4b36-b6e8-2a0f41540344",
-        "practices": [],
-        "prerequisites": [],
-        "difficulty": 2,
-        "topics": [
-          "classes",
+        "practices": [
           "randomness"
```
SleeplessByte: We actually do not "support" `practices` right now, but `prequisites` is great. Can you remove the `practices` value?

quintuple-mallard: Should I fully remove it, or just set it to `[]`?

edit: looked at `config.json`, looks like it should be `[]`. Has been fixed.

config.json
```diff
@@ -2230,12 +2230,11 @@
...
-        "difficulty": 2,
-        "topics": [
-          "classes",
-          "randomness"
-        ]
+        "prerequisites": [
+          "randomness",
+          "classes"
+        ],
+        "difficulty": 2
```
SleeplessByte: Probably want to keep the `topics`?

SleeplessByte (approved):

PR #2713, opened 2025-07-03
Clarify URL format in Regular Chatbot exercise's instructions.md

emgordon154:
Currently, the instructions say to get the URLs, but it doesn't specify what "URLs" are. (https://exercism.org/tracks/javascript/exercises/regular-chatbot is a valid URL, but this exercise does not test for such complicated URLs, and the reference solution in exemplar.js would not reject it.) It would be frustrating for a user to exert their effort trying to write a regex for URLs the exercise isn't testing for (especially if the tests were to reject a valid URL, though I haven't checked if they currently could).


Cool-Katt (approved):

Cool-Katt:
Sorry this took a while.

PR #2722, opened 2025-07-17
Improve test names and prevent some solutions

SleeplessByte:
Changes:
- Whilst everyone is using `typeof` correctly, the newer tests prevent you from coercing incorrectly. 
- It also tests numeric strings correctly (to prevent incorrect regex solutions), adds negative numbers, and numbers that fit in `BigInt` but not `Number` (to force `isFinite`). 
- I also added tests to prevent people from comparing `object.prototype` (to force `instanceof`), by adding branded check to `ElectronicDevice`.
- Added some fake arrays to prevent `instanceof Array`.
- Finally, increased the tests for the final few tasks, particularly checking `type: null` so people cannot do `!object.type`.

### DIFF
--- a/exercises/concept/recycling-robot/.docs/hints.md
+++ b/exercises/concept/recycling-robot/.docs/hints.md
@@ -40,14 +40,14 @@
 - `typeof` returns a string.
 - You can check the length of an array to find out how many elements it contains.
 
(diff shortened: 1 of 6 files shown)


SleeplessByte:
@quintuple-mallard just pinging you in case you're interested!

PR #2730, opened 2025-07-24
Recycling Robot: incorrect return value in example

glennj:


SleeplessByte:
Woops. Thanks @glennj !

PR #2745, opened 2025-08-17
Update tests bank-account

jagdish-15:


exercises/practice/bank-account/bank-account.spec.js
```diff
@@ -107,15 +125,46 @@ describe('Bank Account', () => {
...
+        } catch (e) {
+          throw new Error(`Exception should not be thrown: ${e.message}`);
+        }
+      })
+    );
+  
+    return Promise.all(tasks);
+  }  
+
+  xtest('Changing balance directly throws error', () => {
```
jagdish-15, line 169: This test is an extra one! It isn’t listed in `test.toml` or in `canonical-data.json` on the `problem-specifications` repo.  
It comes from the original test file.  

Should we keep it, or would it be better to remove it?

Cool-Katt: my 2 cents: The test should stay, as it's a valid use case and it's not superseded by any of the other tests

PR #2750, opened 2025-08-25
Update tests word-search

jagdish-15:


jagdish-15:
This one was a bit unusual—the last test (before adding the 4 at the end of the file) seemed off. I’ve fixed it to align with the `canonical-data.json`, added the remaining four tests, removed all the `describe` sections to maintain consistency with the canonical data, and updated the proof solution accordingly.  

Here’s the odd test I was referring to:

```
    xtest('should fail to locate a word that is not in the puzzle', () => {
      const grid = [
        'jefblpepre',
        'camdcimgtc',
        'oivokprjsm',
        'pbwasqroua',
        'rixilelhrs',
        'wolcqlirpc',
        'screeaumgr',
        'alxhpburyi',
        'jalaycalmp',
        'clojurermt',
      ];

      const expectedResults = {
        fail: undefined,
      };
      const wordSearch = new WordSearch(grid);

      expect(wordSearch.find(['fail'])).toEqual(expectedResults);
    });
```

In the `expectedResult`, this was checking if the word `"fail"` exists in the puzzle rather than verifying whether all the other words are located, and it fails only for the word that isn’t present in the grid! Or am I missing something?

jagdish-15:
@SleeplessByte 

What do you think?

SleeplessByte:
Eh, we'll have to check the problem-spec PR to figure out what was intended.

jagdish-15:
> Eh, we'll have to check the problem-spec PR to figure out what was intended.

I've changed the test mentioned above to match what was intended in the `problem-specificatio` repo. I just wanted to confirm it's what was expected.

PR #2769, opened 2025-10-01
Enhance template string lesson with newline escape sequence demonstration

FFFF-0000h:


SleeplessByte:
Forum post here: https://forum.exercism.org/t/clarity-on-template-strings/19454/2

concepts/template-strings/introduction.md
```diff
@@ -32,7 +32,20 @@ When you are needing to have strings formatted on multiple lines:
 `This is an example of using template
 strings to accomplish multiple
 lines`;
+
+/* => This is an example of using template
+     strings to accomplish multiple
+     lines
+*/
 ```
+If you want to represent a newline inside a single-line string (not using backticks), you can use the newline escape sequence `\n`:
```
SleeplessByte:
```suggestion
If you want to represent a newline inside a regular string instead of using a template string (ie. not using backticks), you can use the newline escape sequence `\n`:
```

FFFF-0000h:
> Because word-wrap can be on and you can compose strings over multiple lines using `+`, I rather say regular as opposed to template, instead of single-line. WDYT?

That makes sense.

SleeplessByte:
(Because you are doing this from `main`, we cannot run our format command. Next time it helps if you PR from a branch!)

FFFF-0000h:
> (Because you are doing this from `main`, we cannot run our format command. Next time it helps if you PR from a branch!)

Haha. I knew there would be an issue when I deleted "patch-2" earlier. I don't mind creating a new PR, t for Thanks.

FFFF-0000h:
Wait, do you think it's best practice I create a new PR. I have searched and it seems Github doesn't allow changing the base in a PR, correct me if I'm wrong.
If not, we can go ahead and close this PR and deal with the newly created PR @ https://github.com/exercism/javascript/pull/2776

Cool-Katt:
@FFFF-0000h Since these are small PRs and they're basically identical in content, I'll let it slide this time, but usually we would like to avoid opening PRs left and right. In cases like this, if it's only the formatting CI complaining, usually one of the maintainers would have to run a format in a separate PR after merging and that's why it's extra work.

@SleeplessByte We should probably amend the `CONRIBUTING.md` or the general `README.md` with instructions about branch naming, but I'll open a forum post to talk.

FFFF-0000h:
> @FFFF-0000h Since these are small PRs and they're basically identical in content, I'll let it slide this time, but usually we would like to avoid opening PRs left and right. In cases like this, if it's only the formatting CI complaining, usually one of the maintainers would have to run a format in a separate PR after merging and that's why it's extra work.
> 
> @SleeplessByte We should probably amend the `CONRIBUTING.md` or the general `README.md` with instructions about branch naming, but I'll open a forum post to talk.

Thanks @Cool-Katt, noted.

PR #2788, opened 2025-10-19
Train Driver: Fix incorrect JSDoc type

rivea0:
Changed `@param {...numbers} ids` to `@param {...number} ids` as `numbers` is not a valid JSDoc type.

See https://forum.exercism.org/t/incorrect-function-stub-and-jsdoc-type-in-train-driver-exercise/19608.


SleeplessByte (approved):
Thanks!

PR #2789, opened 2025-10-28
Fix typo in instructions for 'High Score Board' exercise

mytiador:
There are only 5 functions to implement, not 6.


SleeplessByte:
Forum post here: https://forum.exercism.org/t/typo-in-instructions-for-high-score-board-exercise/19770

PR #2806, opened 2026-01-06
sync `satellite` tests

BNAndras:
Related to https://github.com/exercism/javascript/issues/2797. Mark as tiny please.

### DIFF
--- a/exercises/practice/satellite/.meta/tests.toml
+++ b/exercises/practice/satellite/.meta/tests.toml
@@ -1,6 +1,13 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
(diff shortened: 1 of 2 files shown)


BNAndras:
@exercism/javascript 
After merging this PR, https://github.com/exercism/javascript/actions/runs/20740558348/job/59546344880 reported a precheck failure indicating `amusement-park`'s package.json doesn't match the package.json at the root of the repo. Running scripts/sync.mjs, I get 164 package.json files that need to be synced because the root package.json has `"core-js": "~3.45.1",` where the individual exercise package.json files had `"core-js": "~3.42.0",`.

This discrepancy came from a dependabot PR two months ago that bumped the root package.json core-js version. Should I PR an update to the individual package.json files?

BNAndras:
> The "Sync all exercises" action has started running.

Uh oh the robots are taking over.

SleeplessByte:
> This discrepancy came from a dependabot PR two months ago that bumped the root package.json core-js version. Should I PR an update to the individual package.json files?

@BNAndras yes if you can PR the sync, that'd be great. (just one pr that does all).

We sometimes miss it. When core-js is updated it also needs to be changed in the config file (I think it's the one in babel). Can you update that too before you sync?

https://github.com/exercism/javascript/blob/main/babel.config.js#L2

PR #2812, opened 2026-01-24
Use correct exercise name in the skeleton comment for Camicia

ryanplusplus:


BNAndras:
Thanks for catching that!

PR #2824, opened 2026-03-06
add `split-second-stopwatch`

BNAndras:


BNAndras:
This exercise might be more interesting when https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Duration becomes more widespread. The duration as a string feels a bit meh.

Cool-Katt (approved):

